PR #17, opened 2018-02-27
Definition of type families for expressions and sigma statements

aslesarenko:
Questions:
- do we need structure types e.g. `struct { a: Int; b: Boolean } ?
- do we need subtyping?
- do we assume all items in a collection have the same type?

### DIFF
--- a/docs/wpaper/sigma.tex
+++ b/docs/wpaper/sigma.tex
@@ -7,7 +7,7 @@
 \documentclass[11pt]{article}
 \usepackage{fullpage}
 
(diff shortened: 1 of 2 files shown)


kushti:
@aslesarenko I don't think structural types are needed. All the items in a collection should have the same type. On subtyping,  the only case I can think about is to make BooleanLike type to get then SBoolean and SigmaBoolean (but is this needed?)

aslesarenko:
@kushti structures can be used to represent context and it's properties and they can be added later.

PR #19, opened 2018-03-15
Sigma lang type checking

ergomorphic:
Type checker `SigmaTyper` is implemented. For more examples see `SigmaTyperTests`.
Basic examples:
- predefined objects with properties
```
let s = INPUTS.size
```
- let definitions
```
let s = INPUTS.size
let h = HEIGHT + 1
s < h
```
- lambda expressions
```
fun (out: Box) = {
     out.amount >= minToRaise && out.propositionBytes == projectPubKey.propBytes
}
```
- function definitions
```
fun plus(x: Int, y: Int) = x + y
```
- conditional expressions
```
if (condition) {
  let A = 10; A
} else
  if ( x == y) 2 else 3
```
- arrays
```
let items = Array(Array(x + 1))
let first = items(0)
```

### DIFF
--- a/src/main/scala/sigmastate/lang/SigmaTyper.scala
+++ b/src/main/scala/sigmastate/lang/SigmaTyper.scala
@@ -4,7 +4,7 @@ import org.bitbucket.inkytonik.kiama.attribution.Attribution
 import sigmastate._
 import sigmastate.Values._
 import sigmastate.lang.Terms._
-import sigmastate.utxo.Inputs
+import sigmastate.utxo.{Inputs, ByIndex}
 
 /**
   * Analyses for typed lambda calculus expressions.  A simple free variable
@@ -24,16 +24,10 @@ class SigmaTyper(globalEnv: Map[String, Any], tree : SigmaTree) extends Attribut
       case e : SValue =>
 //        checkType(e, tipe) ++
             check(e) {
-              case Apply(e1, e2) =>
-                check(tipe(e1)) {
-                  case _: SFunc => noMessages
-                  case _ =>
-                    message(e1, "application of non-function")
-                }
               case e @ Ident(x, _) =>
                 message(e, s"unknown name '$x'", tipe(e) == NoType)
               case e: SValue =>
-                message(e, s"Expression doesn't have type ${e}: NoType: context ${tree.parent(e)}", tipe(e) == NoType)
+                message(e, s"Expression ${e} doesn't have type: context ${tree.parent(e)}", tipe(e) == NoType)
             }
     }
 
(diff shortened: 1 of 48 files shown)


src/test/scala/sigmastate/lang/BinderTest.scala
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  property("function definitions") {
+    bind(env, "{let f = fun (a: Int) = a + 1; f}") shouldBe
+        Block(Let("f", SFunc(IndexedSeq(SInt), SInt), Lambda(IndexedSeq("a" -> SInt), SInt, Plus(IntIdent("a"), 1))), Ident("f"))
+    bind(env, "{fun f(a: Int) = a + x; f}") shouldBe
+        Block(Let("f", SFunc(IndexedSeq(SInt), SInt), Lambda(IndexedSeq("a" -> SInt), SInt, Plus(IntIdent("a"), 10))), Ident("f"))
+  }
+
+  property("unary operations") {
```
kushti: as discussed, negation should be removed

src/test/scala/sigmastate/lang/SigmaParserTest.scala
```diff
@@ -184,18 +209,54 @@ class SigmaParserTest extends PropSpec with PropertyChecks with Matchers {
...
+      """.stripMargin) shouldBe
+        Block(Let("f", Lambda(IndexedSeq("x" -> SInt), Plus(IntIdent("x"), 1))), Ident("f"))
+    parse(
+      """{fun f(x: Int) = x + 1
+       |f}
+      """.stripMargin) shouldBe
+        Block(Let("f", Lambda(IndexedSeq("x" -> SInt), Plus(IntIdent("x"), 1))), Ident("f"))
   }
 
   property("unary operations") {
```
kushti: negation should be removed

src/main/scala/sigmastate/lang/SigmaTyper.scala
```diff
@@ -0,0 +1,272 @@
...
+      case op @ NEQ(e1, e2) => binOpTipe(op, e1, e2)(tipe(e1), SBoolean)
+
+      case op @ AND(xs) =>
+        val xsT = checkTyped(op, tipe(xs), SCollection(SBoolean))
+        xsT.elemType
+      case op @ OR(xs) =>
+        val xsT = checkTyped(op, tipe(xs), SCollection(SBoolean))
+        xsT.elemType
+
+      case op @ Not(e) =>
```
kushti: negation should be removed

src/test/scala/sigmastate/lang/BinderTest.scala
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  property("predefined functions") {
+    bind(env, "all(Array(c1, c2))") shouldBe
+        AND(ConcreteCollection(Vector(TrueLeaf, FalseLeaf)))
+  }
+
+    property("let constructs") {
+      bind(env, "{let X = 10; X > 2}") shouldBe
+        Block(Let("X", SInt, IntConstant(10)), GT(IntIdent("X"), 2))
```
kushti: previously, byte-length id was used to denote user-provided values. the reason for that is efficient serialization (we need to fight for each byte in the blockchain) of identities. I think we should convert string identifiers to bytes

PR #73, opened 2018-04-26
Examples and tests updates

catena2w:


kushti:
@catena2w tests are failing

PR #104, opened 2018-05-07
Serialization

catena2w:


src/main/scala/sigmastate/utxo/transformers.scala
```diff
@@ -101,6 +105,8 @@ case class Where[IV <: SType](input: Value[SCollection[IV]],
                                id: Byte,
                                condition: Value[SBoolean.type])
   extends Transformer[SCollection[IV], SCollection[IV]] {
+  override val opCode: OpCode = OpCodes.Undefined
```
kushti: serialization is needed for this class

src/main/scala/sigmastate/utxo/transformers.scala
```diff
@@ -83,6 +85,8 @@ case class Append[IV <: SType](input: Value[SCollection[IV]], col2: Value[SColle
 
 case class Slice[IV <: SType](input: Value[SCollection[IV]], from: Value[SInt.type], until: Value[SInt.type])
   extends Transformer[SCollection[IV], SCollection[IV]]{
+  override val opCode: OpCode = OpCodes.Undefined
```
kushti: serialization is needed for this class

PR #107, opened 2018-05-08
Reworking cost() function, Deserialize operation, cost estimation in Interpreter.reduceToCrypto()

kushti:


src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -109,7 +122,7 @@ trait Interpreter {
 
     case t: Transformer[_, _] if t.transformationReady => t.function()
 
-    case _ => null // this means the node cannot be evaluated
+    case v: SValue  => specificTransformations(context, v)
```
ergomorphic: My current understanding is that specificTransformations should be done before evaluateNode is called.
So here we will just waste CPU cycles.
It is better to call `super.evaluateNode(...)`, because rewriting in Scala usually implemented using "stackable traits" pattern

src/main/scala/sigmastate/utxo/ErgoContext.scala
```diff
@@ -52,34 +52,40 @@ object ErgoContext {
...
 }
 
 /** When interpreted evaluates to a collection of BoxConstant built from Context.boxesToSpend */
 case object Inputs extends LazyCollection[SBox.type] {
   override val opCode: OpCode = OpCodes.InputsCode
 
-  val cost = 1
+  override def cost[C <: Context[C]](context: C) =
+    context.asInstanceOf[ErgoContext].boxesToSpend.size * Cost.BoxConstant
+
```
ergomorphic: @kushti 
If we assume that Context is immutable during script execution and all boxes in the context already have attached registers then Inputs.cost should include costs of all registers.
In other words we need to measure the cost of Inputs and Outputs checking the size of the all data in them.

src/main/scala/sigmastate/utxo/transformers.scala
```diff
@@ -270,67 +299,88 @@ case class ExtractScriptBytes(input: Value[SBox.type]) extends Extract[SByteArra
...
   override def function(box: EvaluatedValue[SBox.type]): Value[V] =
     box.value.get(registerId).orElse(default).get.asInstanceOf[Value[V]]
 }
 
+trait Deserialize[V <: SType] extends NotReadyValue[V] with Rewritable
 
 
-case class Deserialize[V <: SType](input: Value[SByteArray.type])(implicit val tpe: V)
-  extends Transformer[SByteArray.type, V] with NotReadyValue[V] with Rewritable {
+case class DeserializeContext[V <: SType](id: Byte)(implicit val tpe: V)
```
ergomorphic: @kushti implicit parameters in case classes require explicit implementation of `equals` and `hashCode`

src/main/scala/sigmastate/Values.scala
```diff
@@ -301,7 +294,9 @@ object Values {
   case class ConcreteCollection[V <: SType](value: IndexedSeq[Value[V]])(implicit val tItem: V)
     extends EvaluatedValue[SCollection[V]] with Rewritable {
     override val opCode: OpCode = ConcreteCollectionCode
-    val cost: Int = value.size
+
+    def cost[C <: Context[C]](context: C): Long = value.size
```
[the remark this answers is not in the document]
kushti: fixed

src/main/scala/sigmastate/utxo/ErgoContext.scala
kushti: @ergomorphic ok, I've made a new pricing function for a box based on its size

src/main/scala/sigmastate/interpreter/Interpreter.scala
kushti: @ergomorphic done, specificTransformations eliminated

kushti:
@ergomorphic I've addressed all the comments

PR #122, opened 2018-05-16
Fix: Extract collections from registers

### LINKED ISSUE #88: Extract collection from register
I can't find an example of collection extraction from a register, and expected way to do this like `ExtractRegisterAs[SCollection[Boolean]](Self, R4)(SCollection[Boolean])` does not work

### LINKED ISSUE #98: Replace SByteArray to SCollection[Byte]
Byte type is needed for this
relates to #87, #88

ergomorphic:
Resolves #88
Resolves #98

### DIFF
--- a/src/main/scala/sigmastate/trees.scala
+++ b/src/main/scala/sigmastate/trees.scala
@@ -13,6 +13,7 @@ import sigmastate.serialization.OpCodes
 import sigmastate.serialization.OpCodes._
 import sigmastate.utxo.CostTable.Cost
 import sigmastate.utxo.{CostTable, Transformer}
+import sigmastate.utils.Helpers._
 
 import scala.annotation.tailrec
 import scala.collection.mutable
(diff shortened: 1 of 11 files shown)


src/main/scala/sigmastate/trees.scala
```diff
@@ -104,7 +109,7 @@ case class AND(input: Value[SCollection[SBoolean.type]])
 
   //todo: reduce such boilerplate around AND/OR, folders, map etc
   override def transformationReady: Boolean =
-    input.evaluated && input.items.forall(_.evaluated)
+    input.evaluated && input.fold(_.items.forall(_.evaluated), c => true)
```
kushti: why are you reducing in transformationReady() ?

ergomorphic: this is not reducing. This probably confusing function name. It actualy does pattern matching ConcreteColletion and CollectionConstant cases. Check the function.
I can rename it in the next PR.

PR #127, opened 2018-05-21
Add optional default value to `ByIndex`

### LINKED ISSUE #99: ByIndex problems
- [x] index should not be Scala Int. `Value[SInt.type]` expected
- [ ] default value may me useful

relates to #87, #88, #98

greenhat:
Replaced by #167
Closes #99
Todo:
- [ ] check/confirm that default value is evaluated;

### DIFF
--- a/src/main/scala/sigmastate/lang/SigmaBinder.scala
+++ b/src/main/scala/sigmastate/lang/SigmaBinder.scala
@@ -73,7 +73,11 @@ class SigmaBinder(env: Map[String, Any]) {
 
     // Rule: col(i) --> ByIndex(col, i)
     case Apply(Typed(obj, tCol: SCollection[_]), Seq(IntConstant(i))) =>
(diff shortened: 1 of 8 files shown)


greenhat:
@kushti @catena2w I've with passing default value as a second parameter (see [previous thoughts](https://github.com/ScorexFoundation/sigmastate-interpreter/issues/99#issuecomment-388035036)) 

Here is a couple of usage examples:
`Array(1, 2)(3, 0) == 0`
`OUTPUTS.map(fun (box: Box) = box.value)(3, 0)== 0`

ergomorphic:
@greenhat @catena2w I don't like the idea of having such syntax `Array(1, 2)(3, 0) == 0`.
This makes syntax unreadable.

greenhat:
@ergomorphic Agreed. Let's go with `getOrElse(index, default)`

greenhat:
Continued in #167

PR #132, opened 2018-05-22
New generalized constant representation and DataSerializer

ergomorphic:


src/test/scala/sigmastate/serialization/DataSerializerSpecification.scala
```diff
@@ -0,0 +1,33 @@
...
+    res shouldBe obj
+  }
+
+  property("Data serialization round trip") {
+    forAll { x: Byte => roundtrip[SByte.type](x, SByte) }
+    forAll { x: Boolean => roundtrip[SBoolean.type](x, SBoolean) }
+    forAll { x: Long => roundtrip[SInt.type](x, SInt) }
+    forAll { x: BigInteger => roundtrip[SBigInt.type](x, SBigInt) }
+    forAll { x: EcPointType => roundtrip[SGroupElement.type](x, SGroupElement) }
+    forAll { x: ErgoBox => roundtrip[SBox.type](x, SBox) }
```
greenhat: Neat! I guess this makes `ErgoBoxSerializerSpec` obsolete, right?

ergomorphic: Yes, ones this will be finished. Here is also an example using ByteBuffer and ByteArrayBuilder as more efficient and compact way to do serialization.

PR #144, opened 2018-05-31
Eliminate implicit parameters in case classes to avoid implementing `equals()`/ `hashCode()`

### LINKED ISSUE #111: Implement equals() / hashCode() for case classes with implicit parameters
Equals and hashCode are needed to be done manually for case classes with implicit parameters, such as DeserializeContext, DeserializeRegister, ExtractRegisterAs etc

greenhat:
Closes #111

### DIFF
--- a/src/main/scala/sigmastate/Values.scala
+++ b/src/main/scala/sigmastate/Values.scala
@@ -402,7 +402,7 @@ object Values {
     lazy val value = None
   }
 
-  case class ConcreteCollection[V <: SType](items: IndexedSeq[Value[V]])(implicit val elementType: V)
+  case class ConcreteCollection[V <: SType](items: IndexedSeq[Value[V]], elementType: V)
     extends EvaluatedCollection[V] with Rewritable {
     override val opCode: OpCode = ConcreteCollectionCode
 
@@ -420,13 +420,17 @@ object Values {
     def deconstruct = immutable.Seq[Any](elementType) ++ items
 
     def reconstruct(cs: immutable.Seq[Any]) = cs match {
-      case Seq(t: SType, vs@_*) => ConcreteCollection[SType](vs.asInstanceOf[Seq[Value[V]]].toIndexedSeq)(t)
+      case Seq(t: SType, vs@_*) => new ConcreteCollection[SType](vs.asInstanceOf[Seq[Value[V]]].toIndexedSeq, t)
       case _ =>
         illegalArgs("ConcreteCollection", "(IndexedSeq, SType)", cs)
     }
   }
   object ConcreteCollection {
-    def apply[V <: SType](items: Value[V]*)(implicit tV: V) = new ConcreteCollection(items.toIndexedSeq)
+    def apply[V <: SType](items: Value[V]*)(implicit tV: V) =
+      new ConcreteCollection(items.toIndexedSeq, tV)
+
+    def apply[V <: SType](items: => Seq[Value[V]])(implicit tV: V) =
+      new ConcreteCollection(items.toIndexedSeq, tV)
   }
 
   trait LazyCollection[V <: SType] extends NotReadyValue[SCollection[V]]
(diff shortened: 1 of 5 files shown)


src/main/scala/sigmastate/lang/syntax/Identifiers.scala
```diff
@@ -6,8 +6,10 @@ import sigmastate.lang.syntax.Basic._
 
 //noinspection ForwardReference
 object Identifiers {
-  case class NamedFunction(f: Char => Boolean)
-      (implicit name: sourcecode.Name) extends (Char => Boolean){
+  case class NamedFunction(f: Char => Boolean,
+                           name: sourcecode.Name = sourcecode.Name.generate)
+    extends (Char => Boolean){
+
```
greenhat: @ergomorphic I'm not sure the default value is much better here.

greenhat:
@ergomorphic I found only two case classes in our codebase with implicit parameters.

src/main/scala/sigmastate/lang/syntax/Identifiers.scala
ergomorphic: Don’t think we need to change this at all. This issue relate to IR nodes only. Those that are subject to rewriting and thus need structural equality.

src/main/scala/sigmastate/Values.scala
```diff
@@ -424,6 +424,16 @@ object Values {
       case _ =>
         illegalArgs("ConcreteCollection", "(IndexedSeq, SType)", cs)
     }
+
+    override def equals(obj: scala.Any): Boolean = obj match {
+      case cc: ConcreteCollection[_] =>
```
ergomorphic: Please check if we can simple move elementType to regular args. We already have ConcreteCollection.apply with implicit arg.

greenhat: Oops, I missed it. Thanks!

greenhat:
@ergomorphic Fixed.

src/main/scala/sigmastate/Values.scala
```diff
@@ -420,13 +420,17 @@ object Values {
...
       case _ =>
         illegalArgs("ConcreteCollection", "(IndexedSeq, SType)", cs)
     }
   }
   object ConcreteCollection {
-    def apply[V <: SType](items: Value[V]*)(implicit tV: V) = new ConcreteCollection(items.toIndexedSeq)
+    def apply[V <: SType](items: Value[V]*)(implicit tV: V) =
+      new ConcreteCollection(items.toIndexedSeq, tV)
+
+    def apply[V <: SType](items: => Seq[Value[V]])(implicit tV: V) =
```
ergomorphic: Why it is by-name?

src/main/scala/sigmastate/Values.scala
```diff
@@ -420,13 +420,17 @@ object Values {
     def deconstruct = immutable.Seq[Any](elementType) ++ items
 
     def reconstruct(cs: immutable.Seq[Any]) = cs match {
-      case Seq(t: SType, vs@_*) => ConcreteCollection[SType](vs.asInstanceOf[Seq[Value[V]]].toIndexedSeq)(t)
+      case Seq(t: SType, vs@_*) => new ConcreteCollection[SType](vs.asInstanceOf[Seq[Value[V]]].toIndexedSeq, t)
```
ergomorphic: What about removing Rewritable, it is not necessary any more

src/main/scala/sigmastate/Values.scala
greenhat: To trick the compiler and not have "same type after erasure" error with first `apply`. Using `IndexedSeq` triggers the same error with compiler generated "default" `apply` for case class. Kinda ugly, but did not find anything better.

src/main/scala/sigmastate/Values.scala
greenhat: Done.

PR #157, opened 2018-06-04
Implement VLQ encoding in binary Writer/Reader

### LINKED ISSUE #155: Implement VLQ encoding in binary Writer/Reader
`BinaryWriter` has methods `putShort`, `putInt`, `putLong` to serialize corresponding numeric values into  binary stream.
`BinaryReader` has opposite operations.

We need to save bytes in a stream by using [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity)

- [ ] find and analyze implementation alternatives to satisfy the following 
- [ ] implement in BinaryReader/Writer and cover with tests
- [ ] update serialization tests for sigma IR

Requirements:
 - unsigned encoding is natural choice to store Array length 
 - Encoding for signed and unsigned integers should be different, because unsigned encoding is more efficient [here is example implementation for unsigned](https://github.com/git/git/blob/7fb6aefd2aaffe66e614f7f7b83e5b7ab16d4806/varint.c#L4)
 - For signed integers we should also combine VLQ with [zigzag encoding](https://developers.google.com/protocol-buffers/docs/encoding#types)
- most probably implementation should be taken from existing cross platform library to facilitate alternative implementations on non JVM platforms

greenhat:
Closes #155 
Todo:
- [ ] split API into signed/unsigned types;
- [ ] VLQ and ZigZag format test with predefined expected values for reader/writer;
- [ ] round trip prop test for reader/writer;
- [ ] implement VLQ(see implementation in [protobuf](https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java#L1374), scodec);
- [ ] implement ZigZag encoding for signed values (see protobuf's [imp](https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java#L949);
- [ ] switch to unsigned API calls where appropriate;
- [ ] update serialization tests for sigma IR;


src/test/scala/sigmastate/utils/ByteReaderWriterImpSpecification.scala
```diff
@@ -1,8 +1,53 @@
...
+        Arbitrary.arbLong.arbitrary,
+        arrayGen[Byte]))
+  } yield anyValSeq
+
+  private implicit val arbSeqPrimVal: Arbitrary[Seq[Any]] = Arbitrary(seqPrimValGen)
 
-class ByteReaderWriterImpSpecification extends PropSpec with PropertyChecks with Matchers {
+  property("round trip serialization/deserialization of arbitrary value list") {
+    forAll { values: Seq[Any] =>
+      val writer = values.fold(Serializer.startWriter()) { case (w: ByteArrayWriter, anyVal: Any) =>
```
ergomorphic: @greenhat There are some issues, mostly related to Scala style
1) AnyVal is a type in Scala, so identifier anyVal suggests it has AnyVal type (not Any)
2) Over functionalization is extreme style not encouraged in general. 
Especially is the context of mutable API. 
Writer/Reader is mutable API, in which sequence of operation is primary and everything else is secondary.
Functional operations works best, with immutable data-centric APIs.
I suggest to use imperative `for (anyVal <- values) {}` which make sequence of operations clear.

src/test/scala/sigmastate/utils/ByteReaderWriterImpSpecification.scala
```diff
@@ -1,8 +1,53 @@
...
+          case v: Short => w.putShort(v)
+          case v: Int => w.putInt(v)
+          case v: Long => w.putLong(v)
+          case v: Array[Byte] => w.putInt(v.length).putBytes(v)
+          case ref@_ => fail(s"writer: unsupported value type: ${ref.getClass}");
+        }
+      }.asInstanceOf[ByteArrayWriter]
 
+      val reader = Serializer.startReader(writer.toBytes, 0)
+      values.map {
```
ergomorphic: @greenhat The same here, `map` is not really to organize loops.

src/test/scala/sigmastate/utils/ByteReaderWriterImpSpecification.scala
greenhat: @ergomorphic Agreed.
I got carried away with this `fold`. Although I did see the mutable API I just wanted to encode the notion of using the returned writer which does not make much sense.

src/test/scala/sigmastate/utils/ByteReaderWriterImpSpecification.scala
greenhat: I've used `map` to collect `Assertion`s for a better error tracing and not to organize loops. However, it does not add a value here since assertion failing exceptions do just fine. Thanks for catching this.

src/test/scala/sigmastate/utils/ByteReaderWriterImpSpecification.scala
greenhat: @ergomorphic BTW, would `foreach` be a better option than `for` here? I found it more concise with less indenting.
```
      for(any <- values) {
        any match {
          case ...
        }
      }
```
vs 
```
      values.foreach {
        case ...
      }
```

src/main/scala/sigmastate/utils/ByteWriter.scala
```diff
@@ -12,8 +12,39 @@ trait ByteWriter {
   def put(x: Byte): ByteWriter
   def putBoolean(x: Boolean): ByteWriter
   def putShort(x: Short): ByteWriter
+
```
greenhat: @ergomorphic I'd appreciate your feedback for the writer/reader API changes below. I've made them with this one commit. Basically, use VLQ with ZigZag in `putInt`, `putLong` and add `putUInt`, `putULong` with pure VLQ for positive values encoding. See scaladoc in `ByteArrayWriter` methods for details.

src/main/scala/sigmastate/utils/ByteReader.scala
```diff
@@ -27,8 +55,77 @@ class ByteBufferReader(buf: ByteBuffer) extends ByteReader {
...
+    *       encoded with [[ByteArrayWriter.putInt]].
+    * @see [[https://en.wikipedia.org/wiki/Variable-length_quantity]]
+    * @return signed Int
+    */
+  @inline override def getInt(): Int =
+  // should only be changed simultaneously with `putInt`
+    getSInt()
+
+  /**
+    * Decode signed Int previously encoded with [[ByteArrayWriter.putSInt]] using VLQ with ZigZag.
```
ergomorphic: Maybe putInt?

src/main/scala/sigmastate/utils/ByteReader.scala
```diff
@@ -27,8 +55,77 @@ class ByteBufferReader(buf: ByteBuffer) extends ByteReader {
...
+    *
+    * @note Uses ZigZag encoding. Should be used to decode '''only''' a value that was previously
+    *       encoded with [[ByteArrayWriter.putSInt]].
+    * @see [[https://en.wikipedia.org/wiki/Variable-length_quantity]]
+    * @return signed Int
+    */
+  @inline def getSInt(): Int = ByteBufferReader.decodeZigZagInt(getULong().toInt)
```
ergomorphic: Maybe getInt? SInt is a type, so the reading may be confusing. Also Int, UInt is typical pair.

greenhat: @ergomorphic Agreed. Not much sense in having both `*SInt` and `*Int` dangling around when  `*Int` behavior is settled (VLQ + ZigZag vs. VLQ).

greenhat: Fixed. And `*Long` as well`

PR #162, opened 2018-06-07
Flatten multiple AND, OR and optimize serialization

### LINKED ISSUE #154: Optimize represenation and serialization of multiple AND/OR
script `true && true && true` should be decoded do `AND(TrueLeaf, TrueLeaf, TrueLeaf)`, now it is decoded to `AND(AND(TrueLeaf, TrueLeaf), TrueLeaf)`.
Test created in https://github.com/ScorexFoundation/sigmastate-interpreter/pull/146

greenhat:
Close #154 

Todo:
- [x] flatten multiple AND in `SigmaSpecializer`;
- [x] flatten multiple OR in `SigmaSpecializer`;
- [x] optimize serialization (see [comment](https://github.com/ScorexFoundation/sigmastate-interpreter/issues/154#issuecomment-394110136))
- [x] prop check tests with arbitrary expression trees;
- [x] test cases with `SigmaBoolean`;
- [x] `getUInt` should return Long;
- [x] test flattening `AND(true, true, AND(ProveDlog(...), ProveDlog(...)))`
- [x] add `putUShort` and `getUShort: Int` with VLQ encoding and use it for collection size instead of `putShort` in serializers;


src/test/scala/sigmastate/lang/SigmaSpecializerTest.scala
```diff
@@ -77,4 +77,15 @@ class SigmaSpecializerTest extends PropSpec with PropertyChecks with Matchers wi
...
+    spec("true && (true && (true && true)) && true") shouldBe
+      AND(TrueLeaf, TrueLeaf, TrueLeaf, TrueLeaf, TrueLeaf)
+  }
+
+  property("OR flattening") {
+    spec("true || true || true") shouldBe OR(TrueLeaf, TrueLeaf, TrueLeaf)
+    spec("true || (true || true) || true") shouldBe
+      OR(TrueLeaf, TrueLeaf, TrueLeaf, TrueLeaf)
+  }
 }
```
ergomorphic: @greenhat please add more comprehensive test cases
true && true
true && false
true && (true && 10 == 10)
(10 == 10 && 10 == 10 && 10 == 10)
etc. all different possible combinations

greenhat: Thanks! Will do. If only I could figure out a way to employ prop checks here. No worthy ideas so far but I'll keep digging.

src/main/scala/sigmastate/serialization/transformers/LogicalTransformerSerializer.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  override def parseBody(bytes: Array[Byte], pos: Position): (Transformer[I, O], Consumed) = {
+    val r = Serializer.startReader(bytes, pos)
+    val size = r.getShort()
+    val values =  (1 to size).map(_ => r.getValue().asInstanceOf[Value[SBoolean.type]])
+    cons(values) -> r.consumed
+  }
+
+  override def serializeBody(obj: Transformer[I, O]): Array[Byte] = {
+    obj.input.asInstanceOf[Value[SCollection[SBoolean.type]]] match {
```
ergomorphic: @greenhat why we need this type cast? Seams like the compiler should know the type of `obj.input`. 
If it is unavoidable it is better to line-comment the reason.

src/main/scala/sigmastate/serialization/transformers/LogicalTransformerSerializer.scala
```diff
@@ -0,0 +1,38 @@
...
+        val size = ccSize.toShort
+        val w = Serializer.startWriter()
+          .putShort(size)
+        for (item <- items) {
+          w.putValue(item)
+        }
+        w.toBytes
+    }
+  }
+}
```
ergomorphic: @greenhat I think we can optimize further serialization for the case when we know all items are `BooleanConstant`
in which case we can use single bit for an item.
The same for the case of `CollectionConstant`.

src/test/scala/sigmastate/lang/SigmaSpecializerTest.scala
```diff
@@ -77,4 +93,49 @@ class SigmaSpecializerTest extends PropSpec with PropertyChecks with Matchers wi
...
+    spec("true && (true && (true && true)) && true") shouldBe
+      AND(TrueLeaf, TrueLeaf, TrueLeaf, TrueLeaf, TrueLeaf)
+  }
+
+  property("AND flattening, SigmaBooleans untouched") {
+    val sigmaBooleans1 = AND(Seq(TrueLeaf, CAND(Seq(proveDlogGen.sample.get, proveDHTGen.sample.get))))
+    spec(Map(), sigmaBooleans1) shouldBe sigmaBooleans1
+    val sigmaBooleans2 = AND(Seq(TrueLeaf, COR(Seq(proveDlogGen.sample.get, proveDHTGen.sample.get))))
+    spec(Map(), sigmaBooleans2) shouldBe sigmaBooleans2
+  }
```
ergomorphic: @greenhat How ProveDlog and DiffieHellmanTuple sigma booleans interact with booleans during flattening? Do they flatten together?

src/test/scala/sigmastate/serialization/AndSerializerSpecification.scala
```diff
@@ -1,19 +1,19 @@
...
     ("object", "bytes"),
-    (AND(boolConst(true), boolConst(true)),  Array[Byte](AndCode, ConcreteCollectionCode, 0, 2, SBoolean.typeCode, TrueCode, TrueCode)),
-    (AND(boolConst(true), boolConst(false)), Array[Byte](AndCode, ConcreteCollectionCode, 0, 2, SBoolean.typeCode, TrueCode, FalseCode)),
-    (AND(boolConst(false),boolConst(true)),  Array[Byte](AndCode, ConcreteCollectionCode, 0, 2, SBoolean.typeCode, FalseCode, TrueCode)),
-    (AND(boolConst(false),boolConst(false)), Array[Byte](AndCode, ConcreteCollectionCode, 0, 2, SBoolean.typeCode, FalseCode, FalseCode))
+    (AND(boolConst(true), boolConst(true)),  Array[Byte](AndCode, 0, 2, TrueCode, TrueCode)),
+    (AND(boolConst(true), boolConst(false)), Array[Byte](AndCode, 0, 2, TrueCode, FalseCode)),
+    (AND(boolConst(false),boolConst(true)),  Array[Byte](AndCode, 0, 2, FalseCode, TrueCode)),
+    (AND(boolConst(false),boolConst(false)), Array[Byte](AndCode, 0, 2, FalseCode, FalseCode))
   )
```
ergomorphic: @greenhat I expected these cases serialize into 4 bytes as result of using VLQ encoding of collection length.

src/test/scala/sigmastate/lang/SigmaSpecializerTest.scala
greenhat: Sigma booleans are ignored during flattening. No. 

I just checked that we don't touch them during flattening.

src/main/scala/sigmastate/serialization/transformers/LogicalTransformerSerializer.scala
greenhat: To be addressed in #166.

src/test/scala/sigmastate/serialization/AndSerializerSpecification.scala
greenhat: @ergomorphic Do you want to introduce `put/getUShort`? `putUInt` will do the job but will look kinda sloppy given we limit collection size to `Short.MaxValue`.

src/main/scala/sigmastate/serialization/transformers/LogicalTransformerSerializer.scala
greenhat: Fixed. Switched to `matchCase`.

src/test/scala/sigmastate/lang/SigmaSpecializerTest.scala
ergomorphic: I mean if we have `AND(true, true, AND(ProveDlog(...), ProveDlog(...)))` what will happen?
I think we should flatten.
CAND, COR should be preserved however.
In current version of IR, SigmaBoolean extends Value[SBoolean.type], so they can be mixed among arguments of AND and OR

src/test/scala/sigmastate/serialization/AndSerializerSpecification.scala
ergomorphic: yes, put/getUShort is required, because we have limit of 65536 items in a collection.
Of case in the API it we should use Int `def getUShort: Int`

src/test/scala/sigmastate/lang/SigmaSpecializerTest.scala
greenhat: Oh, I thought they belong to `CAND/COR` ops. My bad. I'm sure the will be flattened. Let me add the case.

src/test/scala/sigmastate/serialization/AndSerializerSpecification.scala
greenhat: Agreed. That also means that we should have `def getUInt:  Long` I suppose.

ergomorphic: yeah, that is right

PR #163, opened 2018-06-07
Fix comparison of collections

ergomorphic:


src/main/scala/sigmastate/Values.scala
```diff
@@ -447,5 +455,9 @@ object Values {
       case const: CollectionConstant[T]@unchecked => whenConstant(const)
       case _ => sys.error(s"Unexpected node $coll")
     }
+    def toConcreteCollection: ConcreteCollection[T] = {
+      val cc = matchCase(cc => cc, _.toConcreteCollection)
```
catena2w: Why not just `matchCase(cc => cc, _.toConcreteCollection)`?

ergomorphic: I usually use such pattern to be able to setup a breakpoint and see the result of a function

ergomorphic: Unfortunately the debugger has no way to see the result of a function

Issue #84, opened 2018-04-28
Implement Finite State Machine with a gradual reveal


kushti:

kushti:
see comments to empty FsmExampleSpecification class

PR #165, opened 2018-06-11
Fsm Example

kushti:
Fix for #84 : the test for FSM execution has been implemented


src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -233,11 +233,31 @@ trait Interpreter {
...
+    */
+
+    //TODO: simplify logic below
+    case i: IsMember
+      if i.tree.evaluated && i.key.isEvaluated && i.proof.isEvaluated =>
+
+      val keyBytes = i.key match {
+        case constant: Constant[SCollection[SByte.type]] => constant.value
+        case _ => i.key.items.map(_.asInstanceOf[EvaluatedValue[SByte.type]]).map(_.value.asInstanceOf[Byte]).toArray
+      }
```
ergomorphic: @kushti it is possible to use matchCase extension method
`val keyBytes = i.key.matchCase(cc => cc.value, c => c.value)`

src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -233,11 +233,31 @@ trait Interpreter {
...
+    /*
+        case IsMember(tree: AvlTreeConstant, ByteArrayConstant(key), ByteArrayConstant(proof)) =>
+          val bv = tree.createVerifier(SerializedAdProof @@ proof)
+          val res = bv.performOneOperation(Lookup(ADKey @@ key))
+          BooleanConstant.fromBoolean(res.isSuccess  && res.get.isDefined)
+    */
+
+    //TODO: simplify logic below
+    case i: IsMember
+      if i.tree.evaluated && i.key.isEvaluated && i.proof.isEvaluated =>
```
ergomorphic: @kushti I found that `evaluated` field is always true for ConcreteCollection, even if not all items are EvaluatedValue.
use `case IsMember(tree: EvaluatedValue[AvlTree], key: EvaluatedValue[SByteArray], proof: EvaluatedValue[SByteArray])`

kushti:
@ergomorphic all the objections have been addressed, please review again!

src/main/scala/sigmastate/Values.scala
```diff
@@ -136,6 +136,9 @@ object Values {
       case Constant(value: Int, SInt) => Some(value)
       case _ => None
     }
+
+    def zero = IntConstant(0)
```
ergomorphic: @kushti it is better to name constants with uppercase because 1) then they can be used in pattern matching 2) they will look like objects which are semantically close

src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -233,11 +233,14 @@ trait Interpreter {
...
-      
-    case IsMember(tree: AvlTreeConstant, ByteArrayConstant(key), ByteArrayConstant(proof)) =>
-      val bv = tree.createVerifier(SerializedAdProof @@ proof)
-      val res = bv.performOneOperation(Lookup(ADKey @@ key))
-      BooleanConstant.fromBoolean(res.isSuccess) // TODO should we also check res.get.isDefined
+
+
+    case IsMember(tree: EvaluatedValue[AvlTreeData], key: EvaluatedValue[SByteArray], proof: EvaluatedValue[SByteArray]) =>
+      val keyBytes = key.matchCase(cc => cc.value, c => c.value)
+      val proofBytes = proof.matchCase(cc => cc.value, c => c.value)
```
ergomorphic: @kushti maybe pattern-match against constant to avoid type cast

src/test/scala/sigmastate/utxo/examples/FsmExampleSpecification.scala
```diff
@@ -1,20 +1,254 @@
...
+    * state3 | script_hash4 | -
+    *
+    * where state3 is a final state, which denotes end of a contract execution (and by executing a script with the hash
+    * script_hash4 it is possible to create any box as a result).
+    *
+    * We put the table description into an authenticated data structure (such as the Merkle tree, we're using
+    * authenticated AVL+ trees though). We require root digest of the tree to be preserved. We require FSM framework
+    * script to be preserved as well. So to trigger a transition from one state a coin remembers to another one,
+    * a spending transaction should create a coin with preserved root digest, and some (valid) new state, show
+    * a proof for a corresponding transition description in the tree, and show a proof for transition script (e.g. a
```
ergomorphic: @kushti why not use Box based terminology. It is more consistent with the language. It also sounds better in the sentences above where. It seems natural for boxes to remember state.

ergomorphic: What can be done with the box in final state? Which is final state: state3 or - ?

PR #167, opened 2018-06-12
Add optional default value to `ByIndex` (new)

### LINKED ISSUE #99: ByIndex problems

greenhat:
Continue from the #127 
Closes #99
Todo:
- [x] fixes after rebase;
- [x] cover all code paths (const/value) in typer;
- [x] update `ByIndex.function/cost/transformationReady`  to use an optional default value;
- [x] update (de)serialization;
- [x] revert changes to `SigmaTyper`;
- [x] implement new syntax;
- [x] rename `SProduct.fields` to `methods` and switch from tuple to a case class;

### DIFF
--- a/src/main/scala/sigmastate/lang/SigmaTyper.scala
+++ b/src/main/scala/sigmastate/lang/SigmaTyper.scala
@@ -150,30 +150,13 @@ class SigmaTyper {
           args match {
             case Seq(Constant(index, _: SNumericType)) =>
               ByIndex[SType](new_f.asCollection, SInt.upcast(index.asInstanceOf[AnyVal]), None)
-            case Seq(Constant(index, _: SNumericType), dvConst@Constant(_, dvtpe)) =>
-              if (dvtpe == elemType)
-                ByIndex[SType](new_f.asCollection, SInt.upcast(index.asInstanceOf[AnyVal]), Some(dvConst))
-              else
-                error(s"Invalid default value(const) type in array application $app: expected collection element type; actual: $dvtpe")
             case Seq(index) =>
               val typedIndex = assignType(env, index)
               typedIndex.tpe match {
                 case _: SNumericType =>
                   ByIndex[SType](new_f.asCollection, typedIndex.upcastTo(SInt), None)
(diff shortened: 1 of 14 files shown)


src/main/scala/sigmastate/lang/SigmaTyper.scala
```diff
@@ -145,18 +145,35 @@ class SigmaTyper {
...
-            case Seq(Constant(i, tpe: SNumericType)) =>
-              ByIndex[SType](new_f.asCollection, SInt.upcast(i.asInstanceOf[AnyVal]))
-            case Seq(arg) =>
-              val newArg = assignType(env, arg)
-              newArg.tpe match {
-                case targ: SNumericType =>
-                  ByIndex[SType](new_f.asCollection, newArg.upcastTo(SInt))
+            case Seq(Constant(index, _: SNumericType)) =>
+              ByIndex[SType](new_f.asCollection, SInt.upcast(index.asInstanceOf[AnyVal]), None)
+            case Seq(Constant(index, _: SNumericType), dvConst@Constant(_, dvtpe)) =>
```
ergomorphic: @greenhat I cannot accept this syntax, I only agreed with IR part of #127 
Please remove all this special cases in SigmaTyper, they are required because the syntax is ad-hoc.
Please add method `applyOrElse` next to  `"size" -> SInt` in `object SCollection`.
Check where SProduct.fields is used.
I expect no changes in SigmaTyper are necessary to add `applyOrElse`, but you will have to add a rule in SigmaSpecializer.
See for example `case Apply(Select(col,"exists", _),...)`.

greenhat: @ergomorphic Agreed. I find `getOrElse` more appealing, especially for developers with less experience it better conveys the notion that you are getting something out of the collection.

greenhat:
Restarted the build, cause "No output has been received in the last 10m0s,"

PR #178, opened 2018-06-15
Optimized serialization of Array[Boolean] 

### LINKED ISSUE #166: Optimized serialization of Array[Boolean]
We can use single bit encoding of Boolean in some often cases.

- [ ] optimize DataSerializer for serialization of Array[Boolean]
- [ ] optimize serialization of ConstantCollection[SBoolean.type]
- [ ] optimize serialization of ConcreteCollection[SBoolean.type] when all values are BooleanConstant

greenhat:
Closes #166 

Todo:
- [x] test with predefined/expected output;
- [x] round trip serialization test for `ConcreteCollection`;
- [x] implement (de)serializaiton packing boolean/bit;
- [x] remove `BooleanConstant` in favor of `Constant[SBoolean]`;
- [x] fix `LogicalTransformerSerializer.serializeBody` (test expressions)


src/main/scala/sigmastate/Values.scala
```diff
@@ -424,7 +426,12 @@ object Values {
 
   case class ConcreteCollection[V <: SType](items: IndexedSeq[Value[V]], elementType: V)
     extends EvaluatedCollection[V] {
-    override val opCode: OpCode = ConcreteCollectionCode
+    override val opCode: OpCode =
```
greenhat: @ergomorphic I don't like having separate op code for ConcreteCollection with boolean constants, but I could not found a better way. How does it look to you?

src/main/scala/sigmastate/Values.scala
```diff
@@ -424,7 +426,12 @@ object Values {
...
+      if (elementType == SBoolean
+        && items.forall(i => i.isInstanceOf[Constant[_]] || i.isInstanceOf[BooleanConstant]))
```
greenhat: @ergomorphic what's the difference between `BooleanConstant` and `Constant[SBoolean.type]`? Both might be in ConcreteCollection.

ergomorphic: @greenhat I thought it will be harder but #179.

src/main/scala/sigmastate/Values.scala
ergomorphic: @greenhat We need to decide about serialization format for the following cases:
- `Constant[SCollection[SBoolean.type]`
- `ConcreteCollection(True, False, True,...)` when all values are constants
- `ConcreteCollection(True, SomeValue, False, SomeOtherValue,...)`

greenhat: @ergomorphic 
* `Constant[SCollection[SBoolean.type]` is bit encoded (see my comment https://github.com/ScorexFoundation/sigmastate-interpreter/pull/178/files#r196445633 )
* `ConcreteCollection(True, False, True,...)` is bit encoded and handled via separate op code here (just below this exact comment) with `ConcreteCollectionBooleanConstantSerializer` 
* as for `ConcreteCollection(True, SomeValue, False, SomeOtherValue,...)` I suggest leave it as is (each value is encoded separately).

ergomorphic: @greenhat agree

src/main/scala/sigmastate/serialization/DataSerializer.scala
```diff
@@ -77,8 +77,11 @@ object DataSerializer {
       if (len > 0xFFFF)
         sys.error(s"Length of array $arr exceeds ${0xFFFF} limit.")
       w.putUShort(len.toShort)
-      for (x <- arr)
-        DataSerializer.serialize(x, tCol.elemType, w)
+      if (arr.isInstanceOf[Array[Boolean]])
+        w.putBits(arr.asInstanceOf[Array[Boolean]])
```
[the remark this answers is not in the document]
ergomorphic: @greenhat I suggest to use tCol.elemType == SBoolean to be consistent with other cases where only type descriptors are used for type dependent logic.

src/main/scala/sigmastate/serialization/transformers/LogicalTransformerSerializer.scala
```diff
@@ -9,33 +9,21 @@ import sigmastate.{SBoolean, SCollection}
...
 
   override val opCode: OpCode = code
 
   override def parseBody(bytes: Array[Byte], pos: Position): (Transformer[I, O], Consumed) = {
-    val r = Serializer.startReader(bytes, pos)
-    val size = r.getUShort()
-    val values =  (1 to size).map(_ => r.getValue().asInstanceOf[Value[SBoolean.type]])
-    cons(values) -> r.consumed
+    val (input, c1) = ValueSerializer.deserialize(bytes, pos)
+    val i = input.asInstanceOf[Value[SCollection[SBoolean.type]]]
```
ergomorphic: @greenhat there are asValue extension methods, check them out please

src/main/scala/sigmastate/serialization/transformers/LogicalTransformerSerializer.scala
```diff
@@ -9,33 +9,21 @@ import sigmastate.{SBoolean, SCollection}
...
-      w.toBytes
-    },
-      _ => {
-        // to be implemented as part of #166
-        ???
-      }
-    )
+  override def serializeBody(obj: Transformer[I, O]): Array[Byte] = {
+    val w = Serializer.startWriter()
+    val coll = obj.input.matchCase(cc => cc, collConst => collConst)
```
ergomorphic: @greenhat why we need this line?

src/main/scala/sigmastate/utils/ByteReader.scala
```diff
@@ -122,6 +130,15 @@ class ByteBufferReader(buf: ByteBuffer) extends ByteReader {
   }
 
   @inline override def getBytes(size: Int): Array[Byte] = buf.getBytes(size)
+
+  @inline override def getBits(size: Int): Array[Boolean] = {
+    if (size == 0) return Array[Boolean]()
+    val bitSet = BitSet.valueOf(buf.getBytes((size + 7) / 8))
+    (0 until size)
+      .map(i => bitSet.get(i))
```
ergomorphic: @greenhat I think we should forbid functional style operations in performance critical code. Please change to code that avoid unnecessary intermediate allocations.

src/test/scala/sigmastate/serialization/RelationsSpecification.scala
```diff
@@ -14,8 +14,8 @@ class RelationsSpecification extends TableSerializationSpecification {
       (LE(LongConstant(2), LongConstant(3)), Array[Byte](LeCode, SLong.typeCode, 4, SLong.typeCode, 6)),
       (GT(LongConstant(6), LongConstant(5)), Array[Byte](GtCode, SLong.typeCode, 12, SLong.typeCode, 10)),
       (GE(LongConstant(6), LongConstant(5)), Array[Byte](GeCode, SLong.typeCode, 12, SLong.typeCode, 10)),
-      (EQ(TrueLeaf, FalseLeaf), Array[Byte](EqCode, TrueCode, FalseCode)),
-      (NEQ(TrueLeaf, FalseLeaf), Array[Byte](NeqCode, TrueCode, FalseCode))
+      (EQ(TrueLeaf, FalseLeaf), Array[Byte](EqCode, SBoolean.typeCode, 1, SBoolean.typeCode, 0)),
+      (NEQ(TrueLeaf, FalseLeaf), Array[Byte](NeqCode, SBoolean.typeCode, 1, SBoolean.typeCode, 0))
```
ergomorphic: @greenhat we need to keep number of bytes minimal. Please consider how to stay within 3 bytes at most

src/main/scala/sigmastate/serialization/ConcreteCollectionBooleanConstantSerializer.scala
```diff
@@ -0,0 +1,34 @@
...
+    val ccSize = cc.items.size
+    require(ccSize <= Short.MaxValue, s"max collection size is Short.MaxValue = ${Short.MaxValue}")
+    val size = ccSize.toShort
+    val w = Serializer.startWriter()
+      .putUShort(size)
+      .putBits(
+      cc.items.map {
+        case v: BooleanConstant => v.value
+        case _ =>
+          ??? // fail with error (this should never happen, see guard in ConcreteCollection.opcode
```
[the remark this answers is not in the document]
greenhat: That was the plan when I was writing this comment. I just forgot about it. :) Thanks for spotting this one.

src/main/scala/sigmastate/utils/ByteReader.scala
greenhat: I got carried away. Will do. Should we consider using Scalan? :)

src/test/scala/sigmastate/serialization/RelationsSpecification.scala
greenhat: @ergomorphic Is it safe to assume that both types in `Relation` sub-types are the same? In this case, I suggest we write a value type only once. As for boolean constants, we can pack them into a bit array.

greenhat: @ergomorphic Oh, I see for `EQ/NEQ` serializer there is a constraint for types to be equal.

greenhat:
@ergomorphic Done fixing all the issues.

src/main/scala/sigmastate/utils/ByteReader.scala
ergomorphic: not yet 😄

src/test/scala/sigmastate/serialization/RelationsSpecification.scala
ergomorphic: @greenhat It look like this is hidden non trivial part of removing TrueLeaf objects
Maybe we should return TrueCode back and use them.

src/main/scala/sigmastate/serialization/trees/Relation2Serializer.scala
```diff
@@ -12,15 +14,30 @@ case class Relation2Serializer[S1 <: SType, S2 <: SType, R <: Relation[S1, S2]]
...
+  override def parseBody(bytes: Array[Byte], pos: Position): (R, Position) = {
+    if (bytes(pos) == ConcreteCollectionBooleanConstantCode) {
+      val r = Serializer.startReader(bytes, pos)
+      val _ = r.getByte() // skip collection op code
+      val booleans = r.getBits(2)
+      val firstArg = BooleanConstant.fromBoolean(booleans(0)).asValue[S1]
+      val secondArg = BooleanConstant.fromBoolean(booleans(1)).asValue[S2]
+      (constructor(firstArg, secondArg), r.consumed)
+    } else {
+      val (firstArg, consumed) = deserialize(bytes, pos)
```
ergomorphic: @greenhat don't hesitate to rewrite old style serialization into new style :)

greenhat:
@ergomorphic Done. I've restored `TrueCode` and `FalseCode` but I did not find a use for them so far. Did you have something in mind?

greenhat:
@ergomorphic wait, I've spotted that passed `constraints` are not enforced in `Relation2Serializer.serializeBody`. I've added an assert and looking into the failing test.

greenhat:
@ergomorphic Done.

PR #180, opened 2018-06-19
Sigserializer challenges update

kushti:
Cryptography-related code refactoring.


src/main/scala/org/ergoplatform/ErgoLikeTransaction.scala
```diff
@@ -137,8 +136,8 @@ object ErgoLikeTransaction {
         (outs :+ bc) -> (p + cs)
       }
 
-      val (proofs, finalPos) = (0 until inputsCount).foldLeft(Seq[SerializedProverResult]() -> posBeforeProofs) { case ((prs, p), _) =>
-        val (pr, cs) = SerializedProverResult.serializer.parseBody(bytes, p)
+      val (proofs, finalPos) = (0 until inputsCount).foldLeft(Seq[ProverResult]() -> posBeforeProofs) { case ((prs, p), _) =>
```
ergomorphic: @kushti let's use Writer/Reader API for all new and changed serializers. Especially this one will be much easier to write and read. It will be also more efficient.

kushti: @ergomorphic reworked!

PR #184, opened 2018-06-24
Tuple operations implemented

ergomorphic:


src/main/scala/sigmastate/Values.scala
```diff
@@ -380,11 +378,15 @@ object Values {
...
   }
 
-  case class Tuple(items: IndexedSeq[Value[SType]]) extends EvaluatedValue[STuple] {
+  case class Tuple(items: IndexedSeq[Value[SType]]) extends EvaluatedValue[STuple] with EvaluatedCollection[SAny.type, STuple] {
     override val opCode: OpCode = TupleCode
-    val cost: Int = value.size
-    val tpe = STuple(items.map(_.tpe))
-    lazy val value = items
+    override def elementType = SAny
+    lazy val cost: Int = value.size
```
kushti: Cost should incorporate cost of the tuple elements, like in a concrete collection: 

def cost[C <: Context[C]](context: C): Long = Cost.ConcreteCollection + items.map(_.cost(context)).sum

PR #186, opened 2018-06-25
Avoid serializing conversion Upcast nodes (TYPER REUSE)

### LINKED ISSUE #164: Avoid serializing conversion Upcast nodes
Reduce serialization size of the scripts by not serializing conversion Upcast nodes, because they can be automatically added as necessary during deserialization.

Considerations for implementation:

Type checking during deserialization should work in the following way.

1. We can think about it as having kind of 'smart constructor' for each IR node, which is used instead of direct construction of the nodes.
2. Each 'smart constructor' checks typing constraints before constructing the node and throws an exception in the case of violations.
3. Deserializer never construct nodes directly using standard constructor.
Thus at the end we obtain an IR which satisfies the set of constraints placed in smart constructors.

4. An observation is that in deserializer we expect the arguments of smart constructors should be valid. This is in contrast to SigmaTyper, where we expect the original tree is partially typed.

5. The 'checking' part of the smart constructors should be decoupled from 'constructing' part.

6. It make sense to collect all the smart constructors in a separate interface SigmaBuilder.
7. We can have two implementations of SigmaBuilder: 1) doing just constructions 2) doing checking and then constructing nodes.
8. Decoration by subclassing can be used to create specialized builders.
9. Deserializer should work up to SigmaBuilder interface.

greenhat:
Close #164 

Todo:
- [x] round trip (de)serialization test for an arbitrary tree with nodes using `Upcast` (mimic real upcast cases);
- [x] remove/skip `Upcast` nodes serialization and re-apply them on deserialization;
- [ ] find a place where to enforce `Constraints` removed from `Relation2Serializer`;

### DIFF
--- a/src/main/scala/sigmastate/lang/SigmaTyper.scala
+++ b/src/main/scala/sigmastate/lang/SigmaTyper.scala
@@ -329,6 +329,7 @@ class SigmaTyper {
     case v: ByteArrayToBigInt => v
     case v: CalcBlake2b256 => v
     case v: CalcSha256 => v
+    case v: Upcast[_, _] => v
     case v => error(s"Don't know how to assignType($v)")
   }
 
--- a/src/main/scala/sigmastate/serialization/ValueSerializer.scala
+++ b/src/main/scala/sigmastate/serialization/ValueSerializer.scala
@@ -128,12 +128,13 @@ object ValueSerializer extends SigmaSerializerCompanion[Value[SType]] {
     else {
       val handler = table(c)
       val (v: Value[SType], consumed) = handler.parseBody(bytes, pos + 1)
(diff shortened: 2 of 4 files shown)


greenhat:
@ergomorphic I've "restored" upcast on deserialization by running a typer pass on a parsed tree. It probably sounds kinda like a hack. What do you think? Upcasts are applied in the typer, and adding a separate upcast restoration phase in deserialization seemed too fragile (should be "in sync" with the typer) and non-DRY.

ergomorphic:
@greenhat I don't think we should reuse typechecker here.
SigmaTyper is not just typechecker, but in fact transformer from partially typed to fully typed tree.
Please see amendment to description of #164.

PR #188, opened 2018-06-26
Avoid serializing conversion Upcast nodes (SigmaBuilder)

### LINKED ISSUE #164: Avoid serializing conversion Upcast nodes

greenhat:
Close #164 

Todo:
- [x] iron out the `SigmaBuilder` API; 
- [x] EQ/NEQ/GT/GE/LT/LE in `SigmaBuilder`;
- [x] ArithOp in `SigmaBuilder`;
- [x] remove upcast logic in `SigmaTyper.bimap`;
- [x] tests for `SigmaTyper` binary operators parameter types;
- [x] remove Constraints in ValueSerializer;
- [x] address `SigmaTyper` vs. `SigmaBuilder` DRY; 
- [x] address `upcast` in `ByIndex` (see `SigmaTyper`);
- [x] tests for `SigmaBuilder`;
- [x] switch `sigmastate.Plus/*` to SigmaBuilder calls;
- [x] remove typer pass in `UpcastDeserializationSpec` after switching to smart constructors in expr gen;
- [x] parameteraize IR node builder users;
- [x] hide builder return type;


ergomorphic:
@greenhat I think we should keep UpcastSerializer just in case not every Upcast node is removed.

greenhat:
@ergomorphic yep makes sense as a fallback.

src/main/scala/sigmastate/serialization/ValueSerializer.scala
```diff
@@ -100,7 +102,13 @@ object ValueSerializer extends SigmaSerializerCompanion[Value[SType]] {
     AppendSerializer
   ).map(s => (s.opCode, s)).toMap
 
-  def serialize(v: Value[SType]): Array[Byte] = v match {
+  private def serializable(v: Value[SType]): Value[SType] = v match {
+    case upcast: Upcast[SType, _]@unchecked =>
+      upcast.input
```
greenhat: @ergomorphic OTOH, regarding restoring `UpcastSerializer` as a fallback, I don't see a way for Upcast to sneak in through this guard and got serialized.

src/main/scala/sigmastate/lang/SigmaTyper.scala
```diff
@@ -339,17 +340,11 @@ class SigmaTyper {
           (newNode: (Value[T], Value[T]) => SValue): SValue = {
     val l1 = assignType(env, l).asValue[T]
     val r1 = assignType(env, r).asValue[T]
-    (l1.tpe, r1.tpe) match {
```
greenhat: @ergomorphic check out my take on reusing type constraining logic from `SigmaBuilded` in `SigmaTyper`.

src/main/scala/sigmastate/lang/SigmaTyper.scala
```diff
@@ -339,17 +340,11 @@ class SigmaTyper {
...
-      case (t1, t2) =>
-        if (t1 == t2)
-          newNode(l1, r1)
-        else
-          error(s"Invalid binary operation $op ($l1, $r1): type mismatch $t1 != $t2")
+    try {
+      newNode(l1, r1)
+    } catch {
+      case e: Throwable =>
+        throw new InvalidBinaryOperationParameters(s"operation $op: $e")
```
greenhat: @ergomorphic substring matching as error detection in sigma typer test seemed too fragile to me. How do you like the idea of specialized exceptions?

greenhat:
@ergomorphic Ready for review, with only one question left: do we really need `UpcastSerializer`? See my comment https://github.com/ScorexFoundation/sigmastate-interpreter/pull/188#discussion_r198365811

src/main/scala/sigmastate/lang/SigmaBuilder.scala
```diff
@@ -0,0 +1,171 @@
...
+import sigmastate.Values.Value
+import sigmastate._
+import sigmastate.lang.Constraints.{TypeConstraint2, onlyNumeric2, sameType2}
+import sigmastate.lang.Terms._
+import sigmastate.lang.exceptions.{BuilderException, ConstraintFailed}
+import sigmastate.serialization.OpCodes
+
+trait SigmaBuilder {
+
+  def EQ[S <: SType](left: Value[S], right: Value[S]): EQ[S]
```
aslesarenko: @greenhat we need to further abstract from internal workings of builders by hiding actual return type.
So I suggest to have signatures like this `def EQ[S <: SType](left: Value[S], right: Value[S]): Value[SBoolean.type]`
We can add synonims `type SBoolValue = Value[SBoolean.type]` to make life better.
Also why not to use typical naming for type arguments like `T,U,V,R` or `A,B,C`.
However `S` is also ok.

src/main/scala/sigmastate/lang/SigmaBuilder.scala
```diff
@@ -0,0 +1,171 @@
...
+  def NEQ[S <: SType](left: Value[S], right: Value[S]): NEQ[S]
```
ergomorphic: @greenhat we need to further abstract from internal workings of builders by hiding actual return type.
So I suggest to have signatures like this def EQ[S <: SType](left: Value[S], right: Value[S]): Value[SBoolean.type]
We can add synonims type SBoolValue = Value[SBoolean.type] to make life better.
Also why not to use typical naming for type arguments like T,U,V,R or A,B,C.
However S is also ok.

src/main/scala/sigmastate/lang/SigmaParser.scala
```diff
@@ -59,18 +59,19 @@ object SigmaParser extends Exprs with Types with Core {
...
     case "&&" => AND(l.asValue[SBoolean.type], r.asValue[SBoolean.type])
     case "==" => EQ(l, r)
+    case "!=" => NEQ(l, r)
     case ">=" => GE(l, r)
     case ">"  => GT(l, r)
     case "<=" => LE(l, r)
     case "<"  => LT(l, r)
-    case "+"  => Plus(l.asValue[SLong.type], r.asValue[SLong.type])
-    case "-"  => Minus(l.asValue[SLong.type], r.asValue[SLong.type])
+    case "+"  => StdSigmaBuilder.Plus(l.asValue[SLong.type], r.asValue[SLong.type])
```
ergomorphic: @greenhat Let's pass the builder as a parameter to the parser. So we can configure parsing with different builders. This is one of the reasons to introduce builders in a first place. :)

src/main/scala/sigmastate/lang/SigmaTyper.scala
```diff
@@ -208,7 +209,7 @@ class SigmaTyper {
         case nl: SNumericType => (m, newArgs) match {
           case ("*", Seq(r)) => r.tpe match {
             case nr: SNumericType =>
-              bimap(env, "*", newObj.asNumValue, r.asNumValue)(Multiply)(tT, tT)
+              bimap(env, "*", newObj.asNumValue, r.asNumValue)(TransformingSigmaBuilder.Multiply)(tT, tT)
```
ergomorphic: @greenhat Same here, let's pass a builder to parser, binder, typer, etc. everywhere we need to construct new IR nodes

src/main/scala/sigmastate/lang/SigmaTyper.scala
ergomorphic: @greenhat Yeah, that would be better. You can also think about encapsulating  some pretty printing logic there.
However, the missing piece of the puzzle is source text location info. 
We need to thread it somehow  from parser down to specializer, so that pretty printer would be able to show location in the original text.

src/main/scala/sigmastate/serialization/ValueSerializer.scala
ergomorphic: @greenhat  We can think about our IR as instructions of some virtual machine, for which we have an interpreter.
So the question is, whether Upcast is valid instruction or not? 
For every valid instruction we need to have a serializer.
Our nearest goal is to finalize this IR, serialization format and interpreter.
Then we will have to describe all this in specification docs, so alternative implementations would be possible.
We are just giving reference implementation here.

src/main/scala/sigmastate/lang/SigmaParser.scala
greenhat: @ergomorphic Damn it, that makes so much sense. :)

src/main/scala/sigmastate/serialization/ValueSerializer.scala
greenhat: @ergomorphic This: "We are just giving reference implementation here." when it all clicked for me. Thanks for the nice write-up!

src/main/scala/sigmastate/lang/SigmaBuilder.scala
greenhat: @ergomorphic hiding return type seems like a good idea, however, I'm struggling to shove it down into `Relation2Serializer` which wants a `Relation[_,_]` (and for a good reason). Any ideas?

src/main/scala/sigmastate/lang/SigmaParser.scala
```diff
@@ -64,18 +64,23 @@ object SigmaParser extends Exprs with Types with Core {
...
-    case "/"  => StdSigmaBuilder.Divide(l.asValue[SLong.type], r.asValue[SLong.type])
-    case "%"  => StdSigmaBuilder.Modulo(l.asValue[SLong.type], r.asValue[SLong.type])
+    case "/"  => builder.Divide(l.asValue[SLong.type], r.asValue[SLong.type])
+    case "%"  => builder.Modulo(l.asValue[SLong.type], r.asValue[SLong.type])
     case _ => error(s"Unknown binary operation $opName")
   }
 
-  def apply(str: String): core.Parsed[Value[_ <: SType], Char, String] = (StatCtx.Expr ~ End).parse(str)
+  private var builder: SigmaBuilder = StdSigmaBuilder
+
```
greenhat: @ergomorphic I found this super ugly but it's all I could come up with. Any ideas on how to make eyes to not bleed are very welcome. :)

src/main/scala/sigmastate/lang/SigmaBuilder.scala
ergomorphic: @greenhat could you please describe the problem?

greenhat: @ergomorphic Ha, while describing the problem I found a solution. Check it out. 
And I changed type parameter name from `S`(leaked from the case classes) to `T`.

src/main/scala/sigmastate/lang/SigmaBuilder.scala
```diff
@@ -9,75 +9,75 @@ import sigmastate.serialization.OpCodes
...
-  def Plus[S <: SNumericType](left: Value[S], right: Value[S]): ArithOp[S]
-  def Minus[S <: SNumericType](left: Value[S], right: Value[S]): ArithOp[S]
-  def Multiply[S <: SNumericType](left: Value[S], right: Value[S]): ArithOp[S]
-  def Divide[S <: SNumericType](left: Value[S], right: Value[S]): ArithOp[S]
-  def Modulo[S <: SNumericType](left: Value[S], right: Value[S]): ArithOp[S]
+  def Plus[T <: SNumericType](left: Value[T], right: Value[T]): TwoArgumentsOperation[T,T,T]
+  def Minus[T <: SNumericType](left: Value[T], right: Value[T]): TwoArgumentsOperation[T,T,T]
+  def Multiply[T <: SNumericType](left: Value[T], right: Value[T]): TwoArgumentsOperation[T,T,T]
+  def Divide[T <: SNumericType](left: Value[T], right: Value[T]): TwoArgumentsOperation[T,T,T]
+  def Modulo[T <: SNumericType](left: Value[T], right: Value[T]): TwoArgumentsOperation[T,T,T]
```
ergomorphic: @greenhat why not just return Value[T]?

src/main/scala/sigmastate/serialization/trees/Relation2Serializer.scala
```diff
@@ -9,7 +9,7 @@ import sigmastate.serialization.{Serializer, ValueSerializer}
 
 case class Relation2Serializer[S1 <: SType, S2 <: SType, R <: Relation[S1, S2]]
 (override val opCode: Byte,
```
ergomorphic: @greenhat why not `R <: Value[SBoolean.type]`?

greenhat: @ergomorphic need to cast to `Relation` in `serializeBody` stopped me earlier. Let me try to go down that path again. I get that narrow type does not matter much here since in `ValueSerializer` it's all `Value[SType]` anyway.

greenhat:
@ergomorphic Done.

PR #189, opened 2018-06-28
Assets support

kushti:
This PR is introducing a new mandatory register (R2 now, but could be changed easily, thanks to refactoring done), which is storing assets(like for a token stored in R0, for an asset value preservation law should hold, though we do not check the preservation in our testing validation code). The register has a type of Collection[(TokenId, Long)], where TokenId is a byte array constant (of fixed size). Serialization of the register must be efficient.


build.sbt
```diff
@@ -5,7 +5,7 @@ name := "sigma-state"
 lazy val commonSettings = Seq(
   organization := "org.scorexfoundation",
   version := "0.9.4-SNAPSHOT",
-  scalaVersion := "2.12.6",
+  scalaVersion := "2.12.4",
   licenses := Seq("CC0" -> url("https://creativecommons.org/publicdomain/zero/1.0/legalcode")),
```
ergomorphic: @kushti why ?

kushti: With latest versions of Scala & SBT, the latter hangs sometimes after "Done compiling" message (and not on my machine only, reported also by Leo). I dunno what is causing this though, Scala or SBT.

src/main/scala/org/ergoplatform/ErgoBox.scala
```diff
@@ -38,46 +39,41 @@ import scala.runtime.ScalaRunTime
...
-    override val additionalRegisters: Map[NonMandatoryIdentifier, _ <: EvaluatedValue[_ <: SType]] = Map(),
-    val transactionId: Array[Byte],
-    val index: Short
-) extends ErgoBoxCandidate(value, proposition, additionalRegisters) {
+                       override val value: Long,
+                       override val proposition: Value[SBoolean.type],
+                       override val additionalTokens: Seq[(TokenId, Long)] = Seq(),
+                       override val additionalRegisters: Map[NonMandatoryRegisterId, _ <: EvaluatedValue[_ <: SType]] = Map(),
+                       val transactionId: Array[Byte],
+                       val index: Short
```
ergomorphic: @kushti why we need index? Each box's position is uniquely defined by `outputCandidates`. At least maybe we don't need to serialize it and save 2 bytes for each box?

src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
```diff
@@ -1,54 +1,68 @@
...
-      case R1 => Some(ByteArrayConstant(propositionBytes))
-      case R2 => None
-      case n: NonMandatoryIdentifier => additionalRegisters.get(n)
+      case ValueRegId => Some(LongConstant(value))
+      case ScriptRegId => Some(ByteArrayConstant(propositionBytes))
+      case TokensRegId =>
+        val tokenTuples = additionalTokens.map { case (id, amount) =>
+          Tuple(ByteArrayConstant(id), LongConstant(amount))
+        }.toIndexedSeq
+        Some(ConcreteCollection(tokenTuples, STuple(SCollection[SByte.type], SLong)))
```
ergomorphic: @kushti just FYI, there is a shortcut `val SByteArray = SCollection(SByte)`. It's systematic usage will make code shorter, and also more efficient, because equals uses eq to check for reference equality. See https://stackoverflow.com/questions/4526706/what-code-is-generated-for-an-equals-hashcode-method-of-a-case-class

src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
```diff
@@ -1,54 +1,68 @@
...
+      case n: NonMandatoryRegisterId => additionalRegisters.get(n)
     }
   }
 
   override def equals(arg: Any): Boolean = arg match {
     case x: ErgoBoxCandidate =>
-      value == x.value &&
-        proposition == x.proposition &&
-        additionalRegisters == x.additionalRegisters
+      bytesWithNoRef sameElements x.bytesWithNoRef
```
ergomorphic: @kushti it is more efficient to use `Arrays.equals`. Currently `Arrays.equals` is used in 10 places and `sameElements` in 3 places. It is better to not use `sameElements` at all and avoid polluting the code cache.

src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
```diff
@@ -63,26 +77,51 @@ object ErgoBoxCandidate {
...
+      val tokensCount = obj.additionalTokens.size.toByte
+      val tokenBytes = if (obj.additionalTokens.nonEmpty) {
+        obj.additionalTokens.map { case (id, amount) =>
+          id ++ Longs.toByteArray(amount)
+        }.reduce(_ ++ _)
+      } else {
+        Array.emptyByteArray
+      }
+
+      Longs.toByteArray(obj.value) ++ propBytes ++ (tokensCount +: tokenBytes) ++ (regNum +: regBytes)
```
ergomorphic: @kushti use Writer to serialize data in a byte array and avoid allocation of intermediate arrays. Also code will be shorter and easier to understand.

src/test/scala/sigmastate/serialization/generators/ValueGenerators.scala
```diff
@@ -209,4 +191,33 @@ trait ValueGenerators extends TypeGenerators {
...
     ))
   } yield Tuple(values)
+
+  val ergoBoxGen: Gen[ErgoBox] = for {
+    l <- arbLong.arbitrary
+    p <- proveDlogGen
+    b <- Gen.oneOf(TrueLeaf, FalseLeaf, p)
+    tId <- Gen.listOfN(32, arbByte.arbitrary)
+    boxId <- arbShort.arbitrary
+    regNum <- Gen.chooseNum[Byte](0, ErgoBox.nonMandatoryRegistersCount)
```
ergomorphic: @kushti this breaks an assumption in DataSerializer for SBox that all registers are densely packed.

src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
greenhat: @ergomorphic @kushti FWIW I've switched the `ErgoBoxCandidate` serializer to reader/writer API here - https://github.com/ScorexFoundation/sigmastate-interpreter/blob/e35a8eea77939443bfcff7a2f62b7648ee18992d/src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala#L74

src/test/scala/sigmastate/utxo/AVLTreeScriptsSpecification.scala
```diff
@@ -33,9 +35,9 @@ class AVLTreeScriptsSpecification extends SigmaTestingCommons {
     val treeData = new AvlTreeData(digest, 32, None)
 
     val env = Map("key" -> key, "proof" -> proof)
-    val prop = compile(env, """isMember(SELF.R3[AvlTree].value, key, proof)""").asBoolValue
+    val prop = compile(env, """isMember(SELF.R4[AvlTree].value, key, proof)""").asBoolValue
```
ergomorphic: @kushti we should probably be able to write script like this?
`SELF.reg[AvlTree](reg1).value`

src/test/scala/sigmastate/utxo/examples/AssetsAtomicExchangeSpecification.scala
```diff
@@ -0,0 +1,111 @@
...
+import sigmastate.utxo._
+
+/**
+  * An example of an atomic ergo <=> asset exchange.
+  * Let's assume that Alice is willing to buy 60 assets of type "token1" for 100 ergo coins, and Bob
+  * is willing to be a counterparty in this deal.
+  *
+  * Alice is creating a box of 100 coins protected by the script "I demand a spending transaction to create a box
+  * which is protected by my public key pubkey_A and contains at least 60 assets of type "token1" and also 1 ergo coin"
+  * (the last condition ensures that the box is easily spendable).
```
ergomorphic: @kushti Why 1 ergo coin makes the box easily spendable?

src/test/scala/sigmastate/utxo/examples/AssetsAtomicExchangeSpecification.scala
```diff
@@ -0,0 +1,111 @@
...
+  *
+  * Similarly, Bob is creating a box protected by a script like "I demand a spending transaction to create a box
```
ergomorphic: @kushti what is inside the box which is created by Bob?

src/main/scala/org/ergoplatform/ErgoBox.scala
kushti: @ergomorphic a box is a first-class citizen in Ergo. It is living in a state, which is basically an id -> box dictionary. Thus we need for an unique id for every box (so two boxes of the same value and proposition should have different ids), thus we store (unique) transaction id and also an output number in the box.

src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
kushti: fixed

src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
kushti: fixed

Issue #44, opened 2018-04-17
Limited one level inheritance


ergomorphic:
This is needed for #43:
- [ ] Sigma type, so that ProveDlog and ProveDiffieHellmanTuple are subclasses of Sigma

PR #195, opened 2018-07-06
SigmaProp type added 

### LINKED ISSUE #43: Implement Sigma type
Subtasks:
- [ ] refactoring of the interpreter to handle Sigma values

ergomorphic:
Close #43, #44 
This allows to pass ProveDlog and DHT values as context variables and in box registers and have access to them from scripts. (before this scripts didn't support this)
In particular current implementation of CrowdFunding implements passing publicImage as constants embedded in the script body, not as parameters.


src/main/scala/org/ergoplatform/ErgoLikeContext.scala
```diff
@@ -25,6 +25,9 @@ class ErgoLikeContext(val currentHeight: Height,
 
   def withTransaction(newSpendingTransaction: ErgoLikeTransactionTemplate[_ <: UnsignedInput]): ErgoLikeContext =
     ErgoLikeContext(currentHeight, lastBlockUtxoRoot, boxesToSpend, newSpendingTransaction, self, extension)
+
+  def withSelf(newSelf: ErgoBox): ErgoLikeContext =
```
kushti: Why can we have need for this?

ergomorphic: @kushti I added it in some intermediate version, now it is not used. Remove?

ergomorphic:
@kushti 
1) renamed Proof -> SigmaProp type
2) all existing examples remain unchanged, i.e. SigmaProp is not necessary to use in most cases
3) It solves the problem with passing ProveDlog values via context vars and registers is the scenario require this.
4) test cases are provided

build.sbt
```diff
@@ -45,6 +45,7 @@ libraryDependencies ++= Seq(
 
 publishMavenStyle := true
 
+parallelExecution in Test := false
```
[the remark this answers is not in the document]
ergomorphic: @greenhat Completely and quite repeatedly. And I also recalled your fix, but it doesn't solve the problem, unfortunately. Also I don't like switching from SCollection to SCollectionType, just to make SBT happy. So we need more principled solution.

kushti:
@ergomorphic approved, but please address my comment

PR #204, opened 2018-07-12
Add the rest of the nodes to SigmaBuilder

### LINKED ISSUE #193: Add the rest of the nodes to SigmaBuilder

greenhat:
Close #193 

- [x] nodes from `trees.scala`;
- [x] prefix `SigmaBuilder` methods with `mk` and use `import builder._`;
- [x] nodes from `transformers.scala`;
- [x] pass constructor into serializers  (ditch hard coded `DeserializationSigmaBuilder`);
- [x] `Tuple`;
- [x] `ProveDiffieHellmanTuple`;
- [x] `ProveDlog`;
- [x] `ConcreteCollection`;
- [x] `TaggedVariable`;
- [x] `Upcast`;
- [x] `Option`;
- [x] `Terms.*`;
- [x] `Constant`;
- [x] too much type casting, try to generalize the type of the enclosing functions (maybe most of the functions don't need to return concrete type);


src/main/scala/sigmastate/lang/SigmaSpecializer.scala
```diff
@@ -12,7 +12,7 @@ import sigmastate.lang.Terms.{Apply, Block, Ident, Lambda, Let, Select, ValueOps
 import sigmastate.lang.exceptions.SpecializerException
 import sigmastate.utxo._
 
-class SigmaSpecializer {
+class SigmaSpecializer(val builder: SigmaBuilder = TransformingSigmaBuilder) {
```
ergomorphic: @greenhat this default argument looks dangerous. Let’s make configuration explicit.

greenhat: @ergomorphic Agreed. Done.

greenhat:
@ergomorphic Yep, confusion with case classes is a real issue, using package name helps a bit, but it takes a cognitive toll anyway. I'll do it with `mk` prefix.

greenhat:
@ergomorphic Ready for review. Do you want me to squash commits for individual nodes? I personally found it easy to do a review on a per-commit basis.

src/main/scala/sigmastate/Values.scala
```diff
@@ -251,9 +273,9 @@ object Values {
 
   implicit class CollectionConstantOps[T <: SType](c: CollectionConstant[T]) {
     def toConcreteCollection: ConcreteCollection[T] = {
-      implicit val tElem = c.tpe.elemType
+      val tElem = c.tpe.elemType
       val items = c.value.map(v => tElem.mkConstant(v.asInstanceOf[tElem.WrappedType]))
-      ConcreteCollection(items)
+      mkConcreteCollection(items, tElem).asInstanceOf[ConcreteCollection[T]]
```
ergomorphic: @greenhat I don't think we should use builder here, because this helper method works as typecast operation.
But let's leave it as it is.

ergomorphic:
@greenhat 1) in the `apply` I think we don't need to use builder at all.
2) for generators, the cast should be in generators code
3) for TaggedVariables, I think the constructors can return Value[...]. I don't think it is necessary to have more specific type, i.e. that it is TaggedVariable.

greenhat:
@ergomorphic Thanks for the hints! Done.

src/main/scala/sigmastate/lang/Terms.scala
```diff
@@ -31,7 +33,7 @@ object Terms {
     def tpe: SType = givenType ?: body.tpe
   }
   object Let {
-    def apply(name: String, value: SValue): Let = Let(name, NoType, value)
+    def apply(name: String, value: SValue): Let = mkLet(name, NoType, value).asInstanceOf[Let]
```
ergomorphic: @greenhat Shouldn't use builder.

greenhat: @ergomorphic Good catch! Sorry, missed this one. Fixed.

src/main/scala/sigmastate/utxo/transformers.scala
```diff
@@ -211,13 +211,13 @@ case class Fold[IV <: SType](input: Value[SCollection[IV]],
 
 object Fold {
   def sum[T <: SNumericType](input: Value[SCollection[T]])(implicit tT: T) =
-    Fold(input, 21, Constant(tT.upcast(0.toByte), tT), 22, Plus(TaggedVariable(22, tT), TaggedVariable(21, tT)))
+    mkFold(input, 21, Constant(tT.upcast(0.toByte), tT), 22, Plus(mkTaggedVariable(22, tT), mkTaggedVariable(21, tT))).asInstanceOf[Fold[T]]
```
ergomorphic: @greenhat I don't think we need builders in concrete factory methods.
They are usefull in more generic context, where the builder come as a parameter and can be configured.

greenhat: @ergomorphic Agree. Fixed.

src/main/scala/sigmastate/lang/SigmaBinder.scala
```diff
@@ -100,10 +99,10 @@ class SigmaBinder(env: Map[String, Any]) {
       val newBinds = for (Let(n, t, b) <- binds) yield {
         if (env.contains(n)) error(s"Variable $n already defined ($n = ${env(n)}")
         val b1 = eval(b, env)
-        Let(n, if (t != NoType) t else b1.tpe, b1)
+        mkLet(n, if (t != NoType) t else b1.tpe, b1).asInstanceOf[Let]
```
ergomorphic: @greenhat It looks like we need to introduce new traits
```
trait Let extends Value[SType] { name: String, givenType: SType, body: SValue }
case class LetNode(...) extends Let {...}
```

src/main/scala/sigmastate/lang/SigmaSpecializer.scala
```diff
@@ -12,12 +11,13 @@ import sigmastate.lang.Terms.{Apply, Block, Ident, Lambda, Let, Select, ValueOps
...
 
-class SigmaSpecializer {
+class SigmaSpecializer(val builder: SigmaBuilder) {
   import SigmaSpecializer._
+  import builder._
 
   /** Create name -> TaggedXXX(tag) pair to be used in environment. */
   def mkTagged(name: String, tpe: SType, tag: Byte): TaggedVariable[SType] = {
-    val tagged = TaggedVariable(tag, tpe)
+    val tagged = mkTaggedVariable(tag, tpe).asInstanceOf[TaggedVariable[SType]]
```
ergomorphic: @greenhat Introduce new trait. The same pattern as with Let.

src/main/scala/sigmastate/serialization/ConstantSerializer.scala
```diff
@@ -15,7 +17,7 @@ object ConstantSerializer extends ByteBufferSerializer[Constant[SType]]  {
   override def deserialize(r: ByteReader): Constant[SType] = {
     val tpe = r.getType()
     val obj = DataSerializer.deserialize(tpe, r)
-    Constant(obj, tpe)
+    builder.mkConstant(obj, tpe).asInstanceOf[Constant[SType]]
```
ergomorphic: @greenhat  Same as with Let

greenhat:
@ergomorphic Done traits. A lot of type casts are gone. Nice!

greenhat:
@ergomorphic I like the idea in general, but that would require adding a lot of traits. If our goal is to eventually "hide" every node's case class behind the trait, I'm all in. However, I'd rather create a ticket, and deal with it outside of this PR (there are other PRs "queued" after this one).

PR #205, opened 2018-07-14
Workaround to allow Transaction have id of type different from Array[Byte]

catena2w:


ergomorphic:
@catena2w could you please give some context and motivation?

greenhat:
Close in favor of #218.

PR #213, opened 2018-07-20
Malicious test suite for deserializer

### LINKED ISSUE #198: Malicious test suite for deserializer
The goal is to fool deserializer into doing something bad and check that it catches this and responds with a clear exception throwing.
The test scripts can be handcrafted bytes.

greenhat:
Close #198 

- [x] empty;
- [x] zeros;
- [x] collection of a crazy size;
- [x] max bytes size limit;
- [x] tuple of tuples crazy deep;
- [x] add range check assert to `getUInt`;
- [x] crazy deep nodes  (AND/OR, etc)
- [x] look into LibFuzzer Corpus;
- [x] use exceptions instead of assert;
- [x] invalid type code;
- [x] invalid op code;
- [x] glue nested call depth tracker to the reader;

### DIFF
--- a/src/main/scala/sigmastate/serialization/SigmaSerializer.scala
+++ b/src/main/scala/sigmastate/serialization/SigmaSerializer.scala
@@ -22,6 +22,9 @@ object Serializer {
   type Position = Int
   type Consumed = Int
 
+  val MaxInputSize: Int = 1024 * 1024 * 1
+  val MaxTreeDepth: Int = 100
+
     /** Helper function to be use in serializers.
     * Starting position is marked and then used to compute number of consumed bytes.
     * val r = Serializer.startReader(bytes, pos)
--- a/src/main/scala/sigmastate/serialization/ValueSerializer.scala
+++ b/src/main/scala/sigmastate/serialization/ValueSerializer.scala
@@ -9,6 +9,9 @@ import sigmastate.serialization.OpCodes._
 import sigmastate.serialization.transformers._
 import sigmastate.serialization.trees.{QuadrupleSerializer, Relation2Serializer, Relation3Serializer}
 import sigmastate.utils.Extensions._
+import org.ergoplatform._
+import sigmastate.lang.DeserializationSigmaBuilder
+import sigmastate.lang.exceptions.ValueDeserializeCallDepthExceeded
 import sigmastate.utils.{ByteReader, ByteWriter, SparseArrayContainer}
 
 import scala.collection.concurrent.TrieMap
@@ -116,8 +119,8 @@ object ValueSerializer extends SigmaSerializerCompanion[Value[SType]] {
   override def deserialize(r: ByteReader): Value[SType] = {
     val depthKey = r.hashCode()
     val depth = nestedValuesDepthPerReader.getOrElseUpdate(depthKey, 0)
-    assert(depth <= Serializer.MaxTreeDepth,
-      s"nested value deserialization call depth($depth) exceeds allowed maximum ${Serializer.MaxTreeDepth}")
+    if (depth > Serializer.MaxTreeDepth)
+      throw new ValueDeserializeCallDepthExceeded(s"nested value deserialization call depth($depth) exceeds allowed maximum ${Serializer.MaxTreeDepth}")
     nestedValuesDepthPerReader.update(depthKey, depth + 1)
     val firstByte = r.peekByte()
     val v = if (firstByte.toUByte <= LastConstantCode) {
--- a/src/main/scala/sigmastate/serialization/ValueSerializer.scala
+++ b/src/main/scala/sigmastate/serialization/ValueSerializer.scala
@@ -11,7 +11,7 @@ import sigmastate.serialization.trees.{QuadrupleSerializer, Relation2Serializer,
 import sigmastate.utils.Extensions._
 import org.ergoplatform._
 import sigmastate.lang.DeserializationSigmaBuilder
-import sigmastate.lang.exceptions.ValueDeserializeCallDepthExceeded
+import sigmastate.lang.exceptions.{InvalidOpCode, ValueDeserializeCallDepthExceeded}
 import sigmastate.utils.{ByteReader, ByteWriter, SparseArrayContainer}
 
 import scala.collection.concurrent.TrieMap
@@ -100,7 +100,8 @@ object ValueSerializer extends SigmaSerializerCompanion[Value[SType]] {
 
   override def getSerializer(opCode: Tag): ValueSerializer[_ <: Value[SType]] = {
     val serializer = serializers.get(opCode)
-    if (serializer == null) sys.error(s"Cannot find serializer for Value with opCode=$opCode")
+    if (serializer == null)
+      throw new InvalidOpCode(s"Cannot find serializer for Value with opCode=$opCode")
     serializer
   }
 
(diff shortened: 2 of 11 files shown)


src/test/scala/sigmastate/serialization/DeserializationResilience.scala
```diff
@@ -12,6 +12,11 @@ class DeserializationResilience extends PropSpec
     an[ArrayIndexOutOfBoundsException] should be thrownBy ValueSerializer.deserialize(Array[Byte]())
   }
 
+  property("max size limit") {
+    val size = 1024 * 1024 * 10
```
greenhat: @ergomorphic should we assume that it's being taken care of by tx max size limit and don't bother checking?

src/test/scala/sigmastate/serialization/TypeSerializerSpecification.scala
```diff
@@ -84,4 +85,9 @@ class TypeSerializerSpecification extends SerializationSpecification {
     roundtrip(STuple(SLong, SLong, SByte, SBoolean, SInt),
       Array[Byte](TupleTypeCode, 5, SLong.typeCode, SLong.typeCode, SByte.typeCode, SBoolean.typeCode, SInt.typeCode))
   }
+
+  property("tuple of tuples crazy deep") {
```
greenhat: @ergomorphic it's throwing `StackOverflow` now. We should have some depth limit.

greenhat:
@ergomorphic Do you have other ideas on what to feed into deserializer? I've looked into fuzzing but it might be an overkill.

src/test/scala/sigmastate/serialization/DeserializationResilience.scala
ergomorphic: @greenhat I think we should check everything we can.

src/test/scala/sigmastate/serialization/TypeSerializerSpecification.scala
ergomorphic: yes, that is what I mean by malicious tests. It is better to have synthetic tests which deterministically reporduce specific cases, like very deep trees or invalid codes etc. 
They should cover all existing error throwing cases.

ergomorphic: @greenhat We need to have something similar to Corpus described here http://www.llvm.org/docs/LibFuzzer.html

greenhat: @ergomorphic I've successfully tested the proof-of-concept for coverage-guided fuzzying in scala (based on scoverage). I've started the project https://github.com/greenhat/scalafuzz-scalac-plugin and I plan to have a random mutations working this week.

src/main/scala/sigmastate/serialization/ValueSerializer.scala
```diff
@@ -107,16 +109,28 @@ object ValueSerializer extends SigmaSerializerCompanion[Value[SType]] {
       getSerializer(opCode).asInstanceOf[ValueSerializer[v.type]].serializeBody(v, w)
   }
 
+  private val nestedValuesDepthPerReader = TrieMap[Int, Int]()
```
greenhat: @ergomorphic I found it super ugly, but the alternative (passing the depth through `parseBody`) seems even uglier. I'm open to any ideas. :)

src/main/scala/sigmastate/serialization/SigmaSerializer.scala
```diff
@@ -72,6 +75,8 @@ trait SigmaSerializerCompanion[TFamily] {
   def getSerializer(opCode: Tag): SigmaSerializer[TFamily, _ <: TFamily]
 
   final def deserialize(bytes: Array[Byte], pos: Position): (TFamily, Consumed) = {
+    if (bytes.length > Serializer.MaxInputSize)
+      throw new InputSizeLimitExceeded(s"input size ${bytes.length} exceeds ${ Serializer.MaxInputSize}")
```
ergomorphic: @greenhat we still have this old-style methods? Let's get rid of them completely. Even if we need the methods like this, they should be wrappers and they shouldn't have any logic other than wrapping.

src/main/scala/sigmastate/serialization/ValueSerializer.scala
ergomorphic: @greenhat We already passing Reader along the tree (i.e. along all recursive calls). This means we can add  `getLevel` `setLevel` methods in the Reader interface so that serializers can update the corresponding state field if they need it.

src/main/scala/sigmastate/utils/Helpers.scala
```diff
@@ -53,6 +53,12 @@ object Helpers {
       case (Some(a1), Some(a2)) => deepHashCode(a1) == deepHashCode(a2)
       case _ => false
     }
+
+  /**
+    * Helper to construct a byte array from a bunch of bytes. The inputs are actually ints so that I
+    * can use hex notation and not get stupid errors about precision.
+  */
+  def bytesFromInts(bytesAsInts: Int*): Array[Byte] = bytesAsInts.map(_.toByte).toArray
```
ergomorphic: @greenhat Maybe check overflow?

src/main/scala/sigmastate/serialization/ValueSerializer.scala
greenhat: @ergomorphic I've ruled that out initially due to probable SRP violation, but it's least of all other evils. :) Thanks.

greenhat: @ergomorphic Done `get/setLevel`.

src/main/scala/sigmastate/utils/ByteReader.scala
```diff
@@ -172,6 +172,10 @@ class ByteBufferReader(buf: ByteBuffer) extends ByteReader {
   @inline override def position_=(p: Int): Unit = buf.position(p)
 
   @inline override def remaining: Int = buf.remaining()
+
+  private val lvl: AtomicInteger = new AtomicInteger(0)
```
ergomorphic: @greenhat Our Reader/Writer implementations are not thread safe, we should probably mention this in ScalaDoc. Because of this making `level` safe doesn't make much sense IMO. 
I also cannot make up a practical case for two threads working with the same Reader instance.

src/main/scala/sigmastate/utils/Helpers.scala
```diff
@@ -58,7 +58,12 @@ object Helpers {
     * Helper to construct a byte array from a bunch of bytes. The inputs are actually ints so that I
     * can use hex notation and not get stupid errors about precision.
   */
-  def bytesFromInts(bytesAsInts: Int*): Array[Byte] = bytesAsInts.map(_.toByte).toArray
+  def bytesFromInts(bytesAsInts: Int*): Array[Byte] =
+    bytesAsInts.map{ i =>
+      // values from unsigned byte range will be encoded as negative values which is expected here
+      assert(i >= Byte.MinValue && i <= 0xFF, s"$i is out of the signed/unsigned Byte range")
+      i.toByte
+    }.toArray
```
ergomorphic: @greenhat I'm suspect here we have boxing of all the bytes in a generic sequence. 
Please check this and avoid by using imperative loop. 
In any case this should be imperative loop, because `map` uses CanBuildFrom with growing intermediate buffers, which are not necessary here because we know the resuting length.

src/main/scala/sigmastate/utils/ByteReader.scala
greenhat: Agree. That was an overkill.

greenhat:
@ergomorphic All done.

PR #219, opened 2018-07-27
Save two byte on serialization for const parameters in PDHT

### LINKED ISSUE #182: Optimize serialization of ops where values have the same type (write once)
Look for the cases of context-dependent serialization, like the case with EQ/NEQ (see #178) we know both sides have the same type and can use bit encoding.

greenhat:
Close #182

### DIFF
--- a/src/main/scala/sigmastate/serialization/transformers/ProveDiffieHellmanTupleSerializer.scala
+++ b/src/main/scala/sigmastate/serialization/transformers/ProveDiffieHellmanTupleSerializer.scala
@@ -2,12 +2,12 @@ package sigmastate.serialization.transformers
 
 import scapi.sigma.ProveDiffieHellmanTuple
 import sigmastate.SGroupElement
-import sigmastate.Values.{SigmaBoolean, Value}
+import sigmastate.Values.{Constant, GroupElementConstant, SigmaBoolean, Value}
 import sigmastate.lang.Terms._
 import sigmastate.serialization.OpCodes.OpCode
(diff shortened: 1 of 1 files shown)


src/main/scala/sigmastate/serialization/transformers/ProveDiffieHellmanTupleSerializer.scala
```diff
@@ -18,18 +18,45 @@ case class ProveDiffieHellmanTupleSerializer(cons:
...
+  private val constCodePrefix: Byte = 0
+
+  override def serializeBody(obj: ProveDiffieHellmanTuple, w: ByteWriter): Unit = obj match {
+    case ProveDiffieHellmanTuple(Constant(_, _), Constant(_, _), Constant(_, _), Constant(_, _)) =>
+      val tpe = obj.gv.tpe
+      assert(tpe == obj.hv.tpe && tpe == obj.uv.tpe && tpe == obj.vv.tpe,
+        s"constants with the same tpe are expected, got $obj")
+      w.put(constCodePrefix)
+      w.putType(tpe)
+      DataSerializer.serialize(obj.gv.asInstanceOf[GroupElementConstant].value, tpe, w)
```
ergomorphic: @greenhat why not extract the value in the pattern?

greenhat: Yeah, not good enough. Fixed. Thanks.

PR #223, opened 2018-07-31
Implement Downcast conversion operation for integer types

### LINKED ISSUE #147: Implement Downcast conversion operation for integer types
- [ ] introduce Downcast IR node similar to Upcast but in reverse direction with overflow checks
- [ ] replace IntToByte, etc with Downcast

greenhat:
Close #147 

- [x] remove `IntToByte`;
- [x] add `toByte`, `toShort`, `toInt` and `toLong`;
- [x] implement in the parser -> ... pipeline;
- [x] (de)serialization;


src/main/scala/sigmastate/lang/SigmaSpecializer.scala
```diff
@@ -56,15 +56,26 @@ class SigmaSpecializer(val builder: SigmaBuilder) {
...
+    case Select(obj, field, _) if obj.tpe.isNumType =>
+      val numValue = obj.asNumValue
+      val iField = numValue.tpe.methodIndex(field)
+      val tRes = if (iField != -1) {
+        numValue.tpe.methods(iField).stype.asInstanceOf[SNumericType]
+      } else
+      // todo make a specializer exception
+        throw new MethodNotFound(s"Cannot find method '$field' in in the object $obj of Product type with methods ${numValue.tpe.methods}")
+      // todo mkCast? (makes either Upcast or Downcast)
+      Some(mkDowncast(numValue, tRes))
```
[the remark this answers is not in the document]
ergomorphic: @greenhat I think it is better to keep the logic for Upcast and Downcast separated for now. Then we can refactor if it will work out well.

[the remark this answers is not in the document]
greenhat: @ergomorphic I did not mean to introduce a new node `NumericCast`. I meant that `builder.mkNumericCast` will return either `Upcast` or `Downcast` node.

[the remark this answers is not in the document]
ergomorphic: @greenhat I've got your point, but these two ops are fundamentally different. Upcast is always safe, Downcast is in general effectful operation, throwing error on overflow. Thus their semantics is different, so it is better to have separate builder methods. The fact that they look the same in the frontend should not mislead, they are different opcodes in the VM, thus different builder methods are necessary.

src/main/scala/sigmastate/lang/SigmaPredef.scala
```diff
@@ -29,7 +29,6 @@ object SigmaPredef {
     "byteArrayToBigInt" -> mkLambda(Vector("input" -> SByteArray), SBigInt, None),
     "longToByteArray" -> mkLambda(Vector("input" -> SLong), SByteArray, None),
 //    "intToBigInt" -> mkLambda(Vector("input" -> SLong), SBigInt, None),
```
ergomorphic: @greenhat  please remove these commented lines as well

src/main/scala/sigmastate/lang/SigmaTyper.scala
```diff
@@ -345,6 +345,7 @@ class SigmaTyper(val builder: SigmaBuilder) {
     case v: EvaluatedValue[_] => v
     case v: SigmaBoolean => v
     case v: Upcast[_, _] => v
+    case v @ Select(_, _, Some(_)) => v
```
ergomorphic: @greenhat looks suspicious

src/main/scala/sigmastate/types.scala
```diff
@@ -194,9 +195,24 @@ trait SPrimType extends SType with SPredefType {
...
 /** Marker trait for all numeric types. */
-trait SNumericType extends SType {
+trait SNumericType extends SProduct {
   import SNumericType._
+
+  override def ancestors: Seq[SType] = Nil
+  val toByte = "toByte"
+  val toShort = "toShort"
+  val toInt = "toInt"
+  val toLong = "toLong"
```
ergomorphic: @greenhat  Let's keep constants capitalized

src/main/scala/sigmastate/types.scala
```diff
@@ -258,6 +274,13 @@ case object SShort extends SPrimType with SEmbeddable with SNumericType {
     case x: Short => x
     case _ => sys.error(s"Cannot upcast value $v to the type $this")
   }
+  override def downcast(v: AnyVal): Short = v match {
+    case b: Byte => b.toShort
```
ergomorphic: @greenhat I'm not sure it is downcast.

src/main/scala/sigmastate/lang/SigmaTyper.scala
greenhat: I made this shortcut for typed Select cause for the expression "1L * 1.toLong" `bimap` runs `r` through typer but it's already typed in [MethodCall handling](http://github.com/ScorexFoundation/sigmastate-interpreter/blob/db66cb4065aa37ac4b3763266e135b10b5256631/src/main/scala/sigmastate/lang/SigmaTyper.scala#L178)

PR #224, opened 2018-08-03
Add `max` and `min` operations

### LINKED ISSUE #145: Add `max` and `min` operations
Add `Max` and `Min` operations defined for numeric types (Byte, Short, Int, Long, BigInt).
`Max(0,1)==1`

greenhat:
Close #145 

Todo:
- [ ] parser -> ... -> interpreter pipeline;
- [ ] serializer;

### DIFF
--- a/src/main/scala/sigmastate/trees.scala
+++ b/src/main/scala/sigmastate/trees.scala
@@ -404,3 +404,16 @@ case class If[T <: SType](condition: Value[SBoolean.type], trueBranch: Value[T],
   override lazy val second = trueBranch
   override lazy val third  = falseBranch
 }
+
+
+case class Min[T <: SNumericType](left: Value[T], right: Value[T])
+  extends TwoArgumentsOperation[T, T, T] with NotReadyValue[T] {
+  override def tpe: T = left.tpe
+  override val opCode: OpCode = OpCodes.MinCode
+}
+
+case class Max[T <: SNumericType](left: Value[T], right: Value[T])
+  extends TwoArgumentsOperation[T, T, T] with NotReadyValue[T] {
+  override def tpe: T = left.tpe
+  override val opCode: OpCode = OpCodes.MaxCode
+}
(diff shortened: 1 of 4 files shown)


src/main/scala/sigmastate/trees.scala
```diff
@@ -404,3 +404,16 @@ case class If[T <: SType](condition: Value[SBoolean.type], trueBranch: Value[T],
   override lazy val second = trueBranch
   override lazy val third  = falseBranch
 }
+
+
+case class Min[T <: SNumericType](left: Value[T], right: Value[T])
```
greenhat: @ergomorphic Any objections to making `min`, `max` nodes case classes? I've ruled out using `ArithOp`, too much difference. Maybe `ComparisonSelector` parametrized by an op code?

ergomorphic: @greenhat looks like ArithOp to me, what problem you see?

greenhat: @ergomorphic I cannot remember now, probably not that significant of a difference. I've been trying to pull it through the `Apply(Ident...` which felt very unnatural (parametrization with concrete numeric type, etc.). I'll try `ArithOp` now. Thanks.

src/test/scala/sigmastate/serialization/TwoArgumentSerializerSpecification.scala
```diff
@@ -15,6 +15,8 @@ class TwoArgumentSerializerSpecification extends TableSerializationSpecification
       (Minus(LongConstant(2), LongConstant(3)), Array[Byte](MinusCode, SLong.typeCode, 4, SLong.typeCode, 6)),
       (Plus(LongConstant(2), LongConstant(3)), Array[Byte](PlusCode, SLong.typeCode, 4, SLong.typeCode, 6)),
       (Multiply(LongConstant(2), LongConstant(3)), Array[Byte](MultiplyCode, SLong.typeCode, 4, SLong.typeCode, 6)),
+      (Min(LongConstant(2), LongConstant(3)), Array[Byte](MinCode, SLong.typeCode, 4, SLong.typeCode, 6)),
+      (Max(LongConstant(2), LongConstant(3)), Array[Byte](MaxCode, SLong.typeCode, 4, SLong.typeCode, 6)),
```
ergomorphic: @greenhat I think it is better to avoid such magic constants. ZigZag encoding, why not use is here explicitly?

greenhat: Done.

PR #229, opened 2018-08-07
Add an array of Digest32 values in ErgoLikeTransaction

greenhat:
Todo:
- [x] (de)serialization of box candidates with token id indices;
- [x] add token ids indices in tx (de)serialization;


greenhat:
@ergomorphic Having finished box (de)serialization optimization(write/read token id index), I'd rather don't go the route of introducing `digests` property in the transaction for the following reasons:
- `ErgoBoxCandidate` -> `ErgoBox` sub-typing (candidate should store full token id, but a box should store only index);
- `ErgoLikeTransaction` and `ErgoBox` coupling (to resolve a token id index to the token id).

Issue #230, opened 2018-08-07
Add String type and string literal syntax


aslesarenko:
Currently we don't have Strings in the language.
This doesn't allow to create ErgoScript which contains some of the constants of IR.
For example it is not possible to do for Array[T], ProveDlog, DHT, BigInteger, etc.
In particular we cannot have a script which contains specific public keys.

The solution:
- [x] Add usual string literals (#231)
- [x] Add string concatenation + (#231)
- [x] add fromBase64, fromBase58, etc decoding function (#238)
- [x] add PK: String -> Value[SigmaProp] decoding of public keys e.g. `PK("2394083")`
- [ ] add generic `deserialize` function which can only deserialize constants (for security reasons)

greenhat:
@ergomorphic 
> add PK: String -> Value[SigmaProp] decoding of public keys e.g. PK("2394083")
What format is expected here? Hex/baseX/other?

ergomorphic:
@greenhat please consult with @catena2w and @kushti. Probably the format should be the same as public keys are serialized in Ergo (in JSON for example). Or how they are shown in Ergo Explorer, to be able to copy-and-paste them.

greenhat:
@ergomorphic  In the explorer we use Base16. If that's the case for JSON API as well then we can quite safely assume Base16 input in `PK`. But I'd rather encode it in type somehow and don't leave as a convention. 
We already have `fromBase58` and 'fromBase64'. We can add `fromBase16` and make `PK(Array[Byte]): Value[SigmaProp]`. Call site will look more verbose `PK(fromBase16("123"))`. We can called it not `fromBase16' but `fromHex` to save more space. Given that string value will be quite long I don't see a problem.

greenhat:
@ergomorphic 
>add generic `deserialize` function which can only deserialize constants (for security reasons)

What signature does it have? I'm struggling to find one that makes sense.
If we're talking about `deserialize[T](s: String): T', where `s` is a byte array encoded to a string, then why not to put the constant in the code itself?

ergomorphic:
@greenhat we need deserialize from string to be able to put constants of any type in the script. Currently we have literals only for some types in a syntax. How you write in a script a constant of type `Array[Byte]`?
With deserialize we can use String as universal literal. 
Deserialize actually works as compiler intrinsic.
During compilation the compiler checks that value can be deserialized to a given type, otherwise it is a compiler error.
```
[...]

PR #231, opened 2018-08-08
Add String type and string literal syntax

greenhat:
Ref #230 

Todo:
- [x] add `SString` type and `StringConstant`;
- [x] enable string literals parsing in the parser;
- [x] strip parsed literal of single/triple quotes;
- [x] (de)serialization of SString constant;
- [x] Add string concatenation + (parse, interpret, (de)serialize);

### DIFF
--- a/src/main/scala/sigmastate/types.scala
+++ b/src/main/scala/sigmastate/types.scala
@@ -354,6 +354,8 @@ case object SString extends SProduct {
   override def methods: Seq[SMethod] = Seq()
   override type WrappedType = String
   override val typeCode: TypeCode = 101: Byte
+  override def mkConstant(v: String): Value[SString.type] = StringConstant(v)
+  override def dataCost(v: SType#WrappedType): Long = Cost.StringConstantDeclaration
 }
 
   case object SGroupElement extends SProduct with SPrimType with SEmbeddable {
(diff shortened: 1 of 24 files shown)


src/main/scala/sigmastate/types.scala
```diff
@@ -349,7 +349,14 @@ case object SBigInt extends SPrimType with SEmbeddable with SNumericType {
   }
 }
 
-case object SGroupElement extends SProduct with SPrimType with SEmbeddable {
+case object SString extends SProduct {
+  override def ancestors: Seq[SType] = Nil
+  override def methods: Seq[SMethod] = Seq()
+  override type WrappedType = String
+  override val typeCode: TypeCode = 101: Byte
+}
```
greenhat: @ergomorphic to choose a type code I just searched for max one already in use and incremented it. Does it sound good? It's quite an error prone technique. Should we at least gather them at one place (like `OpCodes`)?

src/test/scala/sigmastate/lang/SigmaBinderTest.scala
```diff
@@ -23,10 +23,10 @@ class SigmaBinderTest extends PropSpec with PropertyChecks with Matchers with La
   property("simple expressions") {
     bind(env, "x") shouldBe IntConstant(10)
     bind(env, "b1") shouldBe ByteConstant(1)
-    bind(env, "x+y") shouldBe Plus(10, 11)
+    bind(env, "x-y") shouldBe Minus(10, 11)
```
ergomorphic: @greenhat  I don't think this is good idea. We still need tests for most often used operation +. The fact that is becomes overloaded makes these tests even more important.

PR #233, opened 2018-08-15
Implementation of Threshold sigma primitive

ergomorphic:


src/main/scala/sigmastate/lang/SigmaTyper.scala
```diff
@@ -129,6 +129,8 @@ class SigmaTyper(val builder: SigmaBuilder) {
           val newArgs = new_f match {
             case AllSym | AnySym =>
               adaptSigmaPropToBoolean(new_args, argTypes)
+            case AtLeastSym => // TODO: is this correct way to handle atleast?
+              new_args.head+:adaptSigmaPropToBoolean(new_args.tail, argTypes.tail)
```
ergomorphic: @reyzin yes, this is correct.

PR #237, opened 2018-08-26
Comments and changes to wpaper

ergomorphic:


kushti:
@ergomorphic please update the comments and updates

PR #240, opened 2018-08-31
add PK: String -> Value[SigmaProp] decoding of public keys

greenhat:
Ref #230 

Todo
- [x] parser;
- [x] typer;
- [x] nodes, builder;
- [x] specializer;
- [x] interpreter (parse PK from an address, see [wallet](https://github.com/ergoplatform/ergo/pull/362/files#diff-2e7ea9bed5d0b24720f7645525f3ce7bR175));
- [x] (de)serialization;

### DIFF
--- a/src/main/scala/org/ergoplatform/ErgoAddress.scala
+++ b/src/main/scala/org/ergoplatform/ErgoAddress.scala
@@ -7,13 +7,13 @@ import com.google.common.primitives.Ints
 import scapi.sigma.DLogProtocol.ProveDlog
 import scorex.crypto.hash.{Blake2b256, Digest32}
 import scorex.util.encode.Base58
-import sigmastate.Values.{IntConstant, TaggedByteArray, Value}
 import sigmastate._
+import sigmastate.Values.{IntConstant, TaggedByteArray, Value}
 import sigmastate.serialization.{DataSerializer, ValueSerializer}
 import sigmastate.utils.ByteBufferReader
 import sigmastate.utxo.{DeserializeContext, Slice}
 
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
 
 /**
@@ -73,7 +73,10 @@ sealed trait ErgoAddress {
   val script: Value[SBoolean.type]
 }
 
-case class P2PKAddress(pubkey: ProveDlog, pubkeyBytes: Array[Byte]) extends ErgoAddress {
+class P2PKAddress(val pubkey: ProveDlog,
+                  val pubkeyBytes: Array[Byte])
+                 (implicit val encoder: ErgoAddressEncoder) extends ErgoAddress {
+
   override val addressTypePrefix: Byte = P2PKAddress.addressTypePrefix
 
   override val contentBytes: Array[Byte] = pubkeyBytes
@@ -81,23 +84,25 @@ case class P2PKAddress(pubkey: ProveDlog, pubkeyBytes: Array[Byte]) extends Ergo
   override val script: ProveDlog = pubkey
 
   override def equals(obj: Any): Boolean = obj match {
-    case P2PKAddress(_, pkb) => util.Arrays.equals(pubkeyBytes, pkb)
+    case p2pk: P2PKAddress => util.Arrays.equals(pubkeyBytes, p2pk.pubkeyBytes)
     case _ => false
   }
 
   override def hashCode(): Int = Ints.fromByteArray(pubkeyBytes.takeRight(4))
+
+  override def toString: String = encoder.toString(this)
 }
 
 object P2PKAddress {
   val addressTypePrefix: Byte = 1: Byte
 
-  def apply(pubkey: ProveDlog): P2PKAddress = {
+  def apply(pubkey: ProveDlog)(implicit encoder: ErgoAddressEncoder): P2PKAddress = {
     val bs = pubkey.h.getEncoded(true)
-    P2PKAddress(pubkey, bs)
+    new P2PKAddress(pubkey, bs)
   }
 }
 
-case class Pay2SHAddress(scriptHash: Array[Byte]) extends ErgoAddress {
+class Pay2SHAddress(val scriptHash: Array[Byte])(implicit val encoder: ErgoAddressEncoder) extends ErgoAddress {
   override val addressTypePrefix: Byte = Pay2SHAddress.addressTypePrefix
 
   override val contentBytes: Array[Byte] = scriptHash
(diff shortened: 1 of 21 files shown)


src/main/scala/sigmastate/utxo/transformers.scala
```diff
@@ -289,6 +290,30 @@ case class SigmaPropBytes(input: Value[SSigmaProp.type])
...
 
+/** Decode a SigmaProp(ProveDlog) from an ergo address */
+case class ErgoAddressToSigmaProp(input: Value[SString.type])
+  extends Transformer[SString.type, SSigmaProp.type] with NotReadyValue[SSigmaProp.type] {
+  override val opCode: OpCode = OpCodes.ErgoAddressToSigmaPropCode
+
+  override def function(intr: Interpreter, ctx: Context[_], bal: EvaluatedValue[SString.type]): Value[SSigmaProp.type] =
+    intr match {
+      case ergoInterpreter: ErgoLikeInterpreter =>
+        ErgoAddressEncoder(ergoInterpreter.networkPrefix)
```
greenhat: @ergomorphic what do you think of interpreter type check and adding `ErgoLikeInterpreter.networkPrefix`? I'm not sure it's the best way, but I wanted to express address bounds to a particular network instance as well as network type(ergo).

src/test/scala/sigmastate/utxo/ErgoLikeInterpreterSpecification.scala
```diff
@@ -588,4 +588,53 @@ class ErgoLikeInterpreterSpecification extends SigmaTestingCommons {
...
+    val prop1 = ErgoAddressToSigmaProp(StringConstant(p2s.toString)).isValid
+
+    val ctx = ErgoLikeContext(
+      currentHeight = 50,
+      lastBlockUtxoRoot = AvlTreeData.dummy,
+      boxesToSpend = IndexedSeq(),
+      ErgoLikeTransaction(IndexedSeq(), IndexedSeq()),
+      self = ErgoBox(20, TrueLeaf, Seq(), Map()))
+
+    val proof1 = prover.prove(prop1, ctx, fakeMessage).get.proof
```
greenhat: @ergomorphic Should I spent time on P2S and P2SH address types now? I kind of stuck with `sigmastate.utxo.SigmaPropIsValid cannot be cast to sigmastate.Values$SigmaBoolean` error here. I wonder if it worth spending much time or should I skip them and get back to it later.

greenhat: @ergomorphic nevermind, I scraped that off. I saw your comment about missing sigma prop types for P2S and P2SH.

src/main/scala/org/ergoplatform/ErgoAddress.scala
```diff
@@ -0,0 +1,217 @@
...
+  override val contentBytes: Array[Byte] = pubkeyBytes
+
+  override val script: ProveDlog = pubkey
+
+  override def equals(obj: Any): Boolean = obj match {
+    case p2pk: P2PKAddress => util.Arrays.equals(pubkeyBytes, p2pk.pubkeyBytes)
+    case _ => false
+  }
+
+  override def hashCode(): Int = Ints.fromByteArray(pubkeyBytes.takeRight(4))
```
ergomorphic: @greenhat let's do simple optimization and void array allocation here

src/main/scala/org/ergoplatform/ErgoAddress.scala
```diff
@@ -0,0 +1,217 @@
...
+    new P2PKAddress(pubkey, bs)
+  }
+}
+
+class Pay2SHAddress(val scriptHash: Array[Byte])(implicit val encoder: ErgoAddressEncoder) extends ErgoAddress {
+  override val addressTypePrefix: Byte = Pay2SHAddress.addressTypePrefix
+
+  override val contentBytes: Array[Byte] = scriptHash
+
+  //similar script checked in "P2SH - 160 bits" test in sigma repository, but here we use 192 bits
```
ergomorphic: @greenhat let's clarify this comment. I don't understand this.

src/main/scala/org/ergoplatform/ErgoAddress.scala
```diff
@@ -0,0 +1,217 @@
...
+    val scriptIsCorrect = DeserializeContext(scriptId, SBoolean)
+    AND(hashEquals, scriptIsCorrect)
+  }
+
+  override def equals(obj: Any): Boolean = obj match {
+    case p2sh: Pay2SHAddress => util.Arrays.equals(scriptHash, p2sh.scriptHash)
+    case _ => false
+  }
+
+  override def hashCode(): Int = Ints.fromByteArray(scriptHash.takeRight(4))
```
ergomorphic: @greenhat same allocation problem

src/main/scala/sigmastate/lang/SigmaPredef.scala
```diff
@@ -39,6 +39,7 @@ object SigmaPredef {
        "tree" -> SAvlTree, "key" -> SByteArray, "proof" -> SByteArray), SBoolean, None),
     "fromBase58" -> mkLambda(Vector("input" -> SString), SByteArray, None),
     "fromBase64" -> mkLambda(Vector("input" -> SString), SByteArray, None),
+    "PK" -> mkLambda(Vector("input" -> SString), SBoolean, None),
```
ergomorphic: @greenhat PK should return SigmaProp type

src/main/scala/sigmastate/utxo/transformers.scala
ergomorphic: @greenhat I think such kind of information belongs to meta-store. In all modern DBs there is special database called `master` which contains meta-information about main other main databases. This information is typically available for SQL queries (modulo access rights). In our case we can put it in context variables, or better into new context property of Metadata type, which is a structure with potentially many fields.

src/main/scala/org/ergoplatform/ErgoAddress.scala
greenhat: The comment references http://github.com/ScorexFoundation/sigmastate-interpreter/blob/i230-add-pk/src/test/scala/sigmastate/utxo/ErgoLikeInterpreterSpecification.scala#L358
test, which I suppose was written before the address implementation. I've added the test name to the comment.

greenhat:
@ergomorphic Thanks! I've addressed your comments.

PR #245, opened 2018-09-05
Scala like lambda syntax

### LINKED ISSUE #227: Make ErgoScript syntax subset of Scala
To make ErgoScript cut-and-paste compatible with future SigmaDsl for Scala.
 - [ ] Change 'let' keyword to 'val' keyword
 - [ ] replace `fun (x: T) = {...}` with `(x: T) => {...}` or `{ (x: T) => ... }` or `{ x: T => ... }` 
 - [x] Option[T].value => Option[T].get

These are the only syntactic differences, everything else is exactly the same.

greenhat:
Close #227 

Todo:
- [x] single param in parenthesis;
- [x] body of multiple expressions;
- [x] multiple parameters;
- [x] switch to new lambda syntax in code (tests, examples, etc.);
- [x] error on multiple argument lists in lambda;
- [x] fix parsing of a tuple apply inside a lambda (`SigmaParserTest."tuple operations"` test);
- [x] fix parsing of function calls inside a tuple (`BasicOpsSpec."tuple *"`, `ContextEnrichingSpec."prover*` tests);
- [x] fix `Rule110Specification."rule110 - one layer in register"`;
- [x] change `let` to `val`
- [x] remove logging in parsers;
- [x] switch to new lambda syntax in wpaper;
- [x] switch to new lambda syntax in `LangSpec.md`;
- [x] clean up commented code in parser;
- [x] avoid `({...})`, allow to either omit ~~curly braces~~ or parenthesis; use `OUTPUTS.exists { (box: Box) => box.value > 1000 }` syntax;
- [x] complex tests (nested lambdas, etc.);
- [x] negative tests;

### DIFF
--- a/docs/LangSpec.md
+++ b/docs/LangSpec.md
@@ -122,10 +122,10 @@ The following properties and methods can be used with arrays
 Function  | Description
 --------- | ------------
 `Array[T].size` |  number of items in the array
-`fun Array[T].exists(p: T => Boolean): Boolean ` | Returns true if there exists an item `x` in the array for which `p(x) == true`  
-`fun Array[T].forall(f: T => Boolean): Boolean ` | Returns true if for all items `x` in the array `p(x) == true`   
-`fun Array[T].map[R](f: T => R): Array[R] ` | Applies function `f` for each element of array collecting results in a new array of type `R`. 
-`fun Array[T].reduce(f: (T, T) => T): T ` | For an array `Array(a0, ..., aN)` computes `f(f( ...f(f(a0, a1), a2) ...), aN)`. 
+`def Array[T].exists(p: T => Boolean): Boolean ` | Returns true if there exists an item `x` in the array for which `p(x) == true`  
+`def Array[T].forall(f: T => Boolean): Boolean ` | Returns true if for all items `x` in the array `p(x) == true`   
+`def Array[T].map[R](f: T => R): Array[R] ` | Applies function `f` for each element of array collecting results in a new array of type `R`. 
+`def Array[T].reduce(f: (T, T) => T): T ` | For an array `Array(a0, ..., aN)` computes `f(f( ...f(f(a0, a1), a2) ...), aN)`. 
 
 #### AvlTree
 TBD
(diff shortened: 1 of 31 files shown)


docs/LangSpec.md
```diff
@@ -138,32 +138,32 @@ The following function declarations are automatically imported into any script:
 
 ```
 /** Returns true if all the conditions are true */
-fun allOf(conditions: Array[Boolean]): Boolean
+def allOf(conditions: Array[Boolean]): Boolean
```
greenhat: @ergomorphic I suggest using `def` since `fun` is removed. Or we can omit it altogether.

docs/LangSpec.md
```diff
@@ -17,14 +17,14 @@
...
-- let declarations: `let h = blake2b256(pubkey)`
+- val declarations: `val h = blake2b256(pubkey)`
 - if-then-else clause: `if (x > 0) 1 else 0`
 - array literals: `Array(1, 2, 3, 4)`
 - generic high-order array operations: `map`, `fold`, `exists`, `forall`, etc.
 - accessing fields of any predefined structured objects: `box.value`
 - function invocations (predefined and user defined): `proveDlog(pubkey)` 
-- user defined functions: `let isValid(pk: GroupElement) = proveDlog(pk)`
-- lambdas and high-order methods: `OUTPUTS.exists(fun (out: Box) = out.value >= minToRaise)`
+- user defined functions: `val isValid(pk: GroupElement) = proveDlog(pk)`
```
ergomorphic: @greenhat this should be `def`

docs/LangSpec.md
```diff
@@ -100,32 +100,32 @@ As in many languages Array is a collection of items of the same type.
...
-let size = OUTPUTS.size
+val size = OUTPUTS.size
 ```
 
 Even though ErgoScript is not object-oriented language, Array operations use the syntax of method invocations. 
 For example the following script check an existence of some element in the array satisfying some predicate (condition)
 
 ```
-let ok = OUTPUTS.exists(fun (box: Box) = box.value > 1000)
+val ok = OUTPUTS.exists({ (box: Box) => box.value > 1000 })
```
ergomorphic: @greenhat do we support `OUTPUTS.exists { (box: Box) => box.value > 1000 }` ?

docs/LangSpec.md
ergomorphic: @greenhat let's use def

docs/LangSpec.md
```diff
@@ -180,11 +180,11 @@ After the timeout output could be spent by backer only.
...
-  let c2 = allOf(Array(
+  val c1 = HEIGHT >= timeout && backerPubKey
+  val c2 = allOf(Array(
     HEIGHT < timeout,
     projectPubKey,
-    OUTPUTS.exists(fun (out: Box) = {
+    OUTPUTS.exists({ (out: Box) => 
       out.value >= minToRaise && out.propositionBytes == projectPubKey.propBytes
     })
   ))
```
ergomorphic: @greenhat these double parenthesis/brace looks ugly

src/main/scala/sigmastate/lang/syntax/Core.scala
```diff
@@ -38,7 +38,7 @@ trait Core extends syntax.Literals {
   val `@` = O("@")
   val `_` = W("_")
   val `type` = W("type")
-  val `let` = W("let")
+  val `val` = W("val")
```
ergomorphic: @greenhat so`def` is not supported?

greenhat: No.

docs/LangSpec.md
greenhat: @ergomorphic Totally forgot and did not even try it. Let me check.

greenhat: @ergomorphic Yes, `OUTPUTS.exists { (box: Box) => box.value > 1000 }` works. I've used this syntax in some tests, wpaper and language spec.
I've tried to implement `OUTPUTS.exists ( (box: Box) => box.value > 1000 )` with no luck.

greenhat: @ergomorphic Done.

src/main/java/gf2t/GF2_192.java
```diff
@@ -509,7 +509,7 @@ public static void power2To2ToK(GF2_192 res, GF2_192 z, int k) {
             // so that's what we compute.
             // Note that we have no precomputed table for k=7 (i.e., 2^128), because we don't expect
             // this to be needed -- only up to k=6 is used in inversion.
-            // Here's the proof: let m = 64. 2^{2^k} mod (2^{192}-1) = 2^{mn} mod (2^{3m}-1) for n = 2^{k-6}.
+            // Here's the proof: val m = 64. 2^{2^k} mod (2^{192}-1) = 2^{mn} mod (2^{3m}-1) for n = 2^{k-6}.
```
ergomorphic: @greenhat please rollback this change

greenhat: @ergomorphic Thanks for spotting it. Bulk search & replace gone wrong. :) Will fix in my #248 PR.

PR #246, opened 2018-09-11
ByIndex showcase for bug

daron666:


greenhat:
Yes, that's a bug. Filed in #247

aslesarenko:
It should be merged as regression test cases

aslesarenko:
Fixed in #378

PR #248, opened 2018-09-12
Correct handling of optional context variables and registers

### LINKED ISSUE #244: More accurate handling of optional context variables and registers
### Problem
Currently optional values are not fully supported, however this is required for `getVar` and `box.R3[T]` operations.
The way it works now is based on an ad-hoc rule which recognizes expressions of the form `box.R3[Int].get` and `getVar[Array[Byte]](1).get`.
However, if we use `val` declaration `val x = getVar[Array[Byte]](1); x.get` which is equivalent, the rule will not work, so this script will result in compiler error, but it shouldn't.

In particular `isDefined` currently is not implemented at all, so the solution below will solve this as well.

### Solution
1. `ExtractRegisterAs` should be changed to return `Option[V]` (something like `extends Extract[SOption[V]] with NotReadyValue[SOption[V]]`)
2. The same change have to be done for TaggedVariable
3. Add new IR nodes `OptionGet`, `OptionIsDefined`

greenhat:
Close #244 

Todo:

- [x] test new nodes in the serializer;

`GetVar` IR node:
- [x] compiler pipeline;
- [x] interpreter;
- [x] (de)serialization;
- [x] (de)serialization tests;
- [x] handling of wrong type (test);
- [x] move `GetVarType` helper methods to `Values` object (and remove `import`);
- [x] throw on invalid type;

`OptionGet` IR node:
- [x] compiler pipeline;
- [x] interpreter;
- [x] (de)serialization;
- [x] (de)serialization tests;
- [x] switch from wrapping with `OptionGet(v)` to `v.get` syntax;

change `ExtractRegisterAs` to return `Option[T]`:
- [x] compiler pipeline;
- [x] rewrite `.get` in specializer in two passes (`ExtractRegisterAs` and then `OptionGet`);
- [x] interpreter (test `NoneValue`);
- [x] (de)serialization;
- [x] handling of wrong type (test);

`OptionGetOrElse()` IR node :
- [x] rewrite `.getOrElse` for `ExtractRegisterAs` in specializer in terms of `OptionGetOrElse`;
- [x] interpreter;
- [x] (de)serialization;
- [x] remove default value in `ExtractRegisterAs`;
- [x] test with `GetVar`;

`OptionIsDefined` IR node:
- [x] compiler pipeline;
- [x] interpreter;
- [x] (de)serialization;
- [x] test with `GetVar`;


src/main/scala/sigmastate/utxo/transformers.scala
```diff
@@ -414,3 +415,31 @@ case class DeserializeRegister[V <: SType](reg: RegisterId, tpe: V, default: Opt
...
+
+case class GetVar[V <: SType](varId: Byte, override val tpe: SOption[V]) extends NotReadyValue[SOption[V]] {
+  override val opCode: OpCode = OpCodes.GetVarCode
+  override def cost[C <: Context[C]](context: C): Long = context.extension.cost(varId) + 1
+}
+
+object GetVar {
+  def apply[V <: SType](varId: Byte, innerTpe: V): GetVar[V] = GetVar[V](varId, SOption(innerTpe))
+
+  def GetVarBoolean(varId: Byte): GetVar[SBoolean.type] = GetVar(varId, SBoolean)
```
greenhat: @ergomorphic Do you think it's these methods better fit in `Values` object along with `TaggedInt` and friends? This would save one import line, although I think they make more sense here.

src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -93,6 +94,17 @@ trait Interpreter extends ScorexLogging {
       else
         null
 
+    case t: GetVar[_] =>
+      if (context.extension.values.contains(t.varId)) {
+        val v = context.extension.values(t.varId)
+        if (v.tpe != t.tpe.elemType)
+          NoneValue(t.tpe.elemType)
```
ergomorphic: @kushti Should we throw an exception if context variable is defined but has wrong type? /cc @greenhat

src/main/scala/sigmastate/lang/SigmaTyper.scala
```diff
@@ -361,6 +361,8 @@ class SigmaTyper(val builder: SigmaBuilder) {
       if (i >= 0 && i <= Byte.MaxValue) ByteConstant(i.toByte)
       else error(s"Value $i of type Long cannot be converted to Byte.")
     case v: ContextVariable[_] => v
+    case v: GetVar[_] => v
```
ergomorphic: @greenhat we should check that id is of type Byte

src/main/scala/sigmastate/utxo/transformers.scala
ergomorphic: @greenhat Yes, probably, `GetVar.GetVarBoolean(10)` looks too verbose.

src/main/scala/sigmastate/lang/SigmaBuilder.scala
```diff
@@ -147,6 +146,10 @@ trait SigmaBuilder {
   def mkBase64ToByteArray(input: Value[SString.type]): Value[SByteArray]
 
   def mkPK(input: Value[SString.type]): Value[SSigmaProp.type]
+  def mkGetVar[T <: SType](varId: Byte, tpe: SOption[T]): Value[SType]
```
ergomorphic: @greenhat Why not `def mkGetVar[T <: SType](varId: Byte, tpe: T): Value[SOption[T]]`?

src/main/scala/sigmastate/lang/SigmaBuilder.scala
```diff
@@ -147,6 +146,10 @@ trait SigmaBuilder {
...
+  def mkOptionGet[T <: SType](input: Value[SOption[T]]): Value[SType]
```
ergomorphic: @greenhat this can return `Value[T]`, why not?

src/main/scala/sigmastate/lang/SigmaBuilder.scala
```diff
@@ -147,6 +146,10 @@ trait SigmaBuilder {
...
+  def mkOptionGetOrElse[T <: SType](input: Value[SOption[T]], default: Value[T]): Value[SType]
```
ergomorphic: @greenhat the same here

src/main/scala/sigmastate/lang/SigmaTyper.scala
greenhat: @ergomorphic we do it in http://github.com/ScorexFoundation/sigmastate-interpreter/blob/f04e99518095b7787a6dc97e9a89b9500cb04eef/src/main/scala/sigmastate/lang/SigmaBinder.scala#L102 
I've added a few checks and a throw.

src/main/scala/sigmastate/lang/SigmaBuilder.scala
greenhat: @ergomorphic Good catch! Thanks. Makes much more sense. Also synced `mkGetVar` and `apply` behavior  (wrap `tpe` in `SOption`).

src/main/scala/sigmastate/interpreter/Interpreter.scala
greenhat: @ergomorphic @kushti Let me fix it, then. To be sure I get it, we want `getVar[Int](1).isDefined` to throw an exception if the defined variable is not Int, right?

greenhat: @ergomorphic @kushti Implemented throwing on invalid type in both context variable and register extraction.

PR #253, opened 2018-09-24
Tree operations

### LINKED ISSUE #67: AVL tree operations
- [ ] proof of non-existance
- [ ] proof of modification

catena2w:
close #67


src/test/scala/sigmastate/utxo/AVLTreeScriptsSpecification.scala
```diff
@@ -1,21 +1,121 @@
...
+      ByteArrayConstant(proof)), SomeValue(ByteArrayConstant(endDigest)))
+    //    TODO uncomment after fix of "Option constructors are not supported"
+    //    val env = Map("ops" -> opsBytes, "proof" -> proof, "endDigest" -> endDigest)
+    //    val propCompiled = compile(env, """treeModifications(SELF.R4[AvlTree].get, ops, proof) == Some(endDigest)""").asBoolValue
+    //    prop shouldBe propCompiled
+
+    val newBox1 = ErgoBox(10, pubkey)
+    val newBoxes = IndexedSeq(newBox1)
+
+    val spendingTransaction = ErgoLikeTransaction(IndexedSeq(), newBoxes)
```
ergomorphic: @catena2w @kushti maybe we should require boxesToSpend to be a subset of inputs in transaction? 
I always thought that SELF is just a reference to one of the inputs, for which the script is executing. 
I.e. SELF is always part of INPUTS.
Otherwise to execute all scripts in all inputs we will have to remove corresponding input from the transaction to create a correct context. 
The concrete decision about this should be documented somewhere, if not yet.

PR #254, opened 2018-09-24
Include height into R3

kushti:


src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
```diff
@@ -67,6 +69,7 @@ object ErgoBoxCandidate {
                                         w: ByteWriter): Unit = {
       w.putULong(obj.value)
       w.putValue(obj.proposition)
+      w.putUInt(obj.creationHeight)
```
greenhat: @kushti `putULong`?

src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
```diff
@@ -105,6 +108,7 @@ object ErgoBoxCandidate {
     def parseBodyWithIndexedDigests(digestsInTx: Option[Array[Digest32]], r: ByteReader): ErgoBoxCandidate = {
       val value = r.getULong()
       val prop = r.getValue().asBoolValue
+      val creationHeight = r.getUInt()
```
greenhat: @kushti `getULong`?

src/main/scala/org/ergoplatform/ErgoBox.scala
```diff
@@ -46,36 +44,40 @@ class ErgoBox private(
                        override val additionalTokens: Seq[(TokenId, Long)] = Seq(),
```
ergomorphic: @kushti I think it is time to revise scaladoc for ErgoBox and sync it with the implementation.

src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
```diff
@@ -51,11 +52,12 @@ class ErgoBoxCandidate(val value: Long,
     }
   }
 
-  override def hashCode() = ScalaRunTime._hashCode((value, proposition, additionalTokens, additionalRegisters))
-
+  override def hashCode(): Int =
+    ScalaRunTime._hashCode((value, proposition, additionalTokens, additionalRegisters, creationHeight))
```
ergomorphic: @kushti It is better to avoid allocations in hashCode. Because this method is usually hotspot and can be called many times during object lifecycle.

src/test/scala/sigmastate/utxo/BlockchainSimulationSpecification.scala
```diff
@@ -31,12 +31,13 @@ class BlockchainSimulationSpecification extends PropSpec
...
 
-  def generateBlock(state: ValidationState, miner: ErgoLikeProvingInterpreter, height: Int): Block = {
+  def generateBlock(state: ValidationState, miner: ErgoLikeTestProvingInterpreter, height: Int): Block = {
     val minerPubKey = miner.dlogSecrets.head.publicImage
     val boxesToSpend = state.boxesReader.byHeightRegValue(height)
 
     val txs = boxesToSpend.map { box =>
-      val newBoxCandidate = new ErgoBoxCandidate(10, minerPubKey, Seq(), Map(heightReg -> LongConstant(height + windowSize)))
+      val newBoxCandidate =
+        new ErgoBoxCandidate(10, minerPubKey, Seq(), Map(heightReg -> LongConstant(height + windowSize)), height)
```
ergomorphic: @kushti I remember height register has Tuple type, but here it is Long

src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
kushti: @ergomorphic what to do instead, use Statics.mix directly instead? Btw, I realized that there's no hashCode impl for Value, probably it should be implemented?

src/test/scala/sigmastate/utxo/BlockchainSimulationSpecification.scala
kushti: @ergomorphic despite the name, it is just first non-mandatory register in the test, so the type could be any

Issue #256, opened 2018-09-25
ErgoAddress derivation from proposition bytes


oskin1:

ergomorphic:
@oskin1 Could you please give some description with motivation?

greenhat:
@ergomorphic There was some discussion in PR itself (see 
https://github.com/ScorexFoundation/sigmastate-interpreter/pull/257#discussion_r220555067 in particular). If I remember correctly the reason was to parse from a raw proposition bytes and not from a string representation.

PR #257, opened 2018-09-25
Method for `ErgoAddress` derivation from script.

oskin1:


src/main/scala/org/ergoplatform/ErgoAddress.scala
```diff
@@ -203,6 +205,17 @@ case class ErgoAddressEncoder(networkPrefix: Byte) {
       }
     }
   }
+
+  def fromPropositionBytes(bytes: Array[Byte]): Try[ErgoAddress] = Try {
```
greenhat: @oskin1 Could you provide some context, where do you want to use this operation? It looks like an address deserializer to me. We have a special approach to (de)serializer implementation (see [ProveDlog serializer](http://github.com/ScorexFoundation/sigmastate-interpreter/blob/ae50313900d65fcfaaf58e35efd9c9f93c07fda5/src/main/scala/sigmastate/serialization/ProveDlogSerializer.scala#L11))
Also guessing the address type seems fragile. We have address type prefixes in the `toString/fromString`.

oskin1: @greenhat, this method will be used in explorer, where we have only serialised script bytes - we need to match these bytes by leading opcodes or some mask to distinct different address types and convert particular script to suitable ErgoAddress type.

oskin1:
Closes #256

src/main/scala/org/ergoplatform/ErgoAddress.scala
```diff
@@ -203,6 +205,17 @@ case class ErgoAddressEncoder(networkPrefix: Byte) {
...
+    val script: Value[SType] = ValueSerializer.deserialize(bytes)
+    bytes.head match {
```
greenhat: @oskin1 Did you tried matching on the deserialized 'script' itself? It should work here. You should get AST nodes `ProveDlog`, `AND`, `Value[SBoolean.type]`.

oskin1: @greenhat, yes i did, but in case of P2SH we need to match by mask, which contains numerous opcodes - it would be inconvenient to match such structure (AST) in deserialised form

greenhat: @oskin1 I see matching by `OpCodes.AndCode`. I've tried: 
```
  def fromPropositionBytes(bytes: Array[Byte]): Try[ErgoAddress] = Try {
    ValueSerializer.deserialize(bytes) match {
      case d @ ProveDlog(_) =>
        P2PKAddress(d)
      case a @ AND(_) =>
        Pay2SHAddress(a)
      case b: Value[SBoolean.type] @unchecked if b.tpe == SBoolean =>
        Pay2SAddress(b)
      case _ => throw new RuntimeException()
    }
```
and the test is still green. Am I missing something?

src/test/java/org/ergoplatform/ErgoAddressEncoderSpec.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import org.scalatest.{Matchers, PropSpec, TryValues}
+import scapi.sigma.DLogProtocol
+import scapi.sigma.DLogProtocol.DLogProverInput
+import sigmastate.Values
+import sigmastate.serialization.ValueSerializer
+
+class ErgoAddressEncoderSpec extends PropSpec with Matchers with TryValues {
+
+  property("scriptToAddress() should properly distinct all types of addresses from plain script bytes") {
```
greenhat: @oskin1 I've sneaked in `ErgoAddressSpecification` from ergo yesterday after your PR. It's in the master already. Could you please move this test there. And I suppose it's `fromProposition()` instead of `scriptToAddress()`.

src/main/scala/org/ergoplatform/ErgoAddress.scala
greenhat: @oskin1 Excellent! Thanks! Only minor note with test location left.

src/test/java/org/ergoplatform/ErgoAddressEncoderSpec.scala
oskin1: @greenhat, done

PR #259, opened 2018-09-26
Generic `deserialize` for constants

### LINKED ISSUE #230: Add String type and string literal syntax

greenhat:
Close #230 

Todo:
- [x] parser;
- [x] build a deserialized constant and introduce it in the binder;
- [x] test for failing in binder (non-constant string, decoding failed, etc.);
- [x] test interpreter;
- [x] update the docs;

### DIFF
--- a/src/main/scala/sigmastate/lang/SigmaBinder.scala
+++ b/src/main/scala/sigmastate/lang/SigmaBinder.scala
@@ -8,10 +8,10 @@ import sigmastate.lang.Terms._
 import sigmastate._
 import Values._
 import org.ergoplatform._
-import sigmastate.utils.Extensions._
+import scorex.util.encode.Base58
 import sigmastate.interpreter.CryptoConstants
-import sigmastate.lang.exceptions.{BinderException, InvalidArguments}
-import sigmastate.utxo.GetVar
+import sigmastate.lang.exceptions.{BinderException, InvalidArguments, InvalidTypeArguments}
+import sigmastate.serialization.ValueSerializer
 
 class SigmaBinder(env: Map[String, Any], builder: SigmaBuilder) {
   import SigmaBinder._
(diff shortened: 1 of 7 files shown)


src/main/scala/sigmastate/lang/SigmaBinder.scala
```diff
@@ -129,6 +129,20 @@ class SigmaBinder(env: Map[String, Any], builder: SigmaBuilder) {
...
+        error(s"Wrong number of arguments in $e: expected one argument")
+      val str = args.head match {
+        case StringConstant(s) => s
+        case _ =>
+          error(s"invalid argument in $e: expected a string constant")
+      }
+      val bytes = Base58.decode(str).get
+      // todo handle decoding failure
+      Some(
+        ValueSerializer.deserialize(bytes))
```
greenhat: @ergomorphic Is that what you had in mind for `deserialize`? Or should it be done by the interpreter?

ergomorphic: @greenhat yes, you are right. However, this will be currently a little ad-hoc. In C++ it would be constexpr function, but we don't have such modifier.

PR #260, opened 2018-10-01
STypeIdent and ValDef (de)serialization

greenhat:


src/main/scala/sigmastate/types.scala
```diff
@@ -742,13 +742,14 @@ object STypeApply {
...
 case class STypeIdent(name: String) extends SType {
+  require(name.length <= 255, "name is too long")
   override type WrappedType = Any
   override val typeCode = STypeIdent.TypeCode
   override def isConstantSize = false
   override def toString = name
 }
 object STypeIdent {
-  val TypeCode = 95: Byte
+  val TypeCode = 102: Byte
```
greenhat: @ergomorphic I've changed your type code for `STypeIdent` to 102 (next free) otherwise it goes through tuples branch on deserialization.

src/main/scala/sigmastate/serialization/TypeSerializer.scala
```diff
@@ -98,6 +100,12 @@ object TypeSerializer extends ByteBufferSerializer[SType] {
         // `Tuple` type with more than 4 items `(Int, Byte, Box, Boolean, Int)`
         serializeTuple(tup, w)
     }
+    case typeIdent: STypeIdent => {
+      w.put(typeIdent.typeCode)
+      val bytes = typeIdent.name.getBytes(StandardCharsets.UTF_8)
```
ergomorphic: @greenhat Oh, no, we  need to replace type identifiers with integers. See ValDefs for inspiration.

greenhat: @ergomorphic I knew that something was off with this type. :)

PR #265, opened 2018-10-12
Store serialized constants in separate section referenced by indices (placeholders)

### LINKED ISSUE #264: A new format for optimized script execution
#### Massive script validation
Transaction (tx) have INPUTS - an array of boxes to spend.
Every input box can have a script protecting it. 
This script should be executed in a context of the current transaction.
The simplest transaction have 1 input box. 
Thus if we want to have a sustained block validation of 1000 transactions per second 
we need to be able to validate 1000 scripts per second.

#### Operations per script
For every script (of input box) the following is done in order to validate it.
1. Context is created with SELF := box
2. The script is deserialized into ErgoTree (IR)
3. IR is traversed to build costGraph and calcGraph, two graphs for 
cost estimation function and script calculation function.
4. Cost estimation is computed by evaluating costGraph with current context data
5. If cost and data size limits are not exceeded, calcGraph is evaluated 
using context data to obtain either true/false or SigmaTree
6. In the case of SigmaTree, verification procedure is executed

#### Script normalization
Before a script can be stored in a blockchain it should be first compiled from its source text 
into IR and then serialized into byte array.
Because the language is purely functional and IR is graph-based, the compilation process has an 
effect of normalization/unification. 
This means than different original scripts may have identical IR trees and as a result identical
serialized bytes. 
Because of normalization, and also because of script reusability, the number of conceptually (or logically) 
different scripts is much less than the number of individual scripts in a blockchain. 
For example we may have 1000s of _different_ scripts in a blockchain with millions of boxes. 
The average reusability ratio is 1000 in this case.
And even those different scripts may have different usage frequency.

#### Potential Script processing optimization
Having big reusability ratio we can optimize script evaluation by performing steps 1 - 4 only once per unique script.
The compiled calcGraph can be cached in Map[Array[Byte], Context => SigmaBoolean].
Every script extracted from an input box can be used as a key in this map.
 
#### Problem with embedded constants
There is one obstacle to the optimization by caching.
In many cases it is very natural to embed constants in the script body, most notable scenario is
when public keys are embedded.
As result two functionally identical scripts may serialize to different byte arrays because they have different 
embedded constants.

#### Solution: Constant-less lambdas
The solution to the problem with embedded constants is simple, we don't need to embed constants.
Each constant in the body of a lambda can be replaced with indexed placeholder (PH).
Each PH have an index field.
The index of the PH is assigned by breadth-first topological order of the graph traversal.
The transformation is part of compilation and is performed ahead of time.
Each script of the box have an array of all the constants extracted from its body.
Each PH refers to the constant by the constant's index in the array.
The format of serialized script should have 3 sections: `<num constants><constants array><script tree>`
The `<constants array>` contains serialized constant data (using ConstantSerializer) one after another.
The `<script tree>` is a serialized ErgoTree with placeholders.

Using this new script format we can use <script tree> part as a key in the cache.
An observation is that after the constants are extracted, what remains is a template.
Thus instead of applying steps 1-4 to constant-full scripts we can apply them to constant-less templates.
Before applying steps 4 and 5 we need to bind placeholders with actual values taken from `<constants array>`.

greenhat:
Close #264 

Todo:
- [x] store serialized constants in separate section referenced by indices;
- [ ] separate API for this "placeholdered" (de)serialization?
- [ ] boolean constants should be left untouched? (think optimized `AND`/`OR` serialization with bitsets);
- [ ] test `SerializedConstantPlaceholderStore`;
- [ ] fix tests failing due to predefined bytes changed (ignored `tablePredefinedBytesTest`);
- [ ] fix test in `DeserializationResilience` for input of zeroes (ignored);

### DIFF
--- a/src/main/scala/sigmastate/serialization/SerializedConstantPlaceholderStore.scala
+++ b/src/main/scala/sigmastate/serialization/SerializedConstantPlaceholderStore.scala
@@ -44,15 +44,8 @@ class SerializedConstantPlaceholderStore(builder: SigmaBuilder) {
     store.foreach { c => constantSerializer.serialize(c, w) }
   }
 
-  /**
-    * Deserializes constants.
-    * @param r reader
-    * @return self (with deserialized constants)
-    */
-  def deserialize(r: ByteReader): SerializedConstantPlaceholderStore = {
-    // todo Should we rather throw here? Or better, make it part of initialization.
-    // There is no scenario of loading constants in addition to already stored.
-    if (store.nonEmpty) store.clear()
+  protected def deserialize(r: ByteReader): SerializedConstantPlaceholderStore = {
+    require(store.isEmpty, "already have constants")
     val constantsCount = r.getUInt().toInt
     val constantSerializer = ConstantSerializer(builder)
     for (_ <- 0 until constantsCount) {
@@ -63,3 +56,13 @@ class SerializedConstantPlaceholderStore(builder: SigmaBuilder) {
   }
 }
 
+object SerializedConstantPlaceholderStore {
+
+  /**
+    * Deserializes constants.
+    * @param r reader
+    * @return store with deserialized constants
+    */
+  def deserialize(builder: SigmaBuilder, r: ByteReader): SerializedConstantPlaceholderStore =
+    new SerializedConstantPlaceholderStore(builder).deserialize(r)
+}
--- a/src/main/scala/sigmastate/serialization/SerializedConstantPlaceholderStore.scala
+++ b/src/main/scala/sigmastate/serialization/SerializedConstantPlaceholderStore.scala
@@ -5,51 +5,59 @@ import sigmastate.Values.Constant
 import sigmastate.lang.SigmaBuilder
 import sigmastate.utils.{ByteReader, ByteWriter}
 
-import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
 
 class SerializedConstantPlaceholderStore(builder: SigmaBuilder) {
 
-  private val store = mutable.Map[Constant[SType], Int]()
-  private val reversedStore = mutable.Map[Int, Constant[SType]]()
-  private var lastFreeIndex = 0
-
+  private val store = ArrayBuffer[Constant[SType]]()
   def nonEmpty: Boolean = store.nonEmpty
 
-  def getIndex(c: Constant[SType]): Int = {
-    store.get(c) match {
-      case Some(v) => v
-      case None =>
-        val usedIndex = lastFreeIndex
-        store.put(c, usedIndex)
-        reversedStore.put(usedIndex, c)
-        lastFreeIndex = lastFreeIndex + 1
-        usedIndex
-    }
+  /**
+    * Adds the constant and assign the placeholder index.
+    *
+    * Indices are not re-used for the same constant, since we want scripts to be equivalent if they
+    * use different constant values. Every call with the same constant will return new index.
+    * @param c constant
+    * @return placeholder index
+    */
+  def put(c: Constant[SType]): Int = this.synchronized {
+    store.append(c)
+    store.length - 1
   }
 
-  def getConstant(placeholderIndex: Int): Constant[SType] =
-    reversedStore(placeholderIndex)
+  /**
+    * Gets the deserialized constant by the index assigned on serialization.
+    * @param placeholderIndex placeholder index of constant in the script (assigned via `put`)
+    * @return constant
+    */
+  def get(placeholderIndex: Int): Constant[SType] =
+    store(placeholderIndex)
 
+  /**
+    * Serializes the accumulated (via `put`) constants.
+    * @param w writer
+    */
   def serialize(w: ByteWriter): Unit = {
     val constantSerializer = ConstantSerializer(builder)
     if (store.isEmpty) return
     w.putUInt(store.size)
-    store.foreach { case (c, index) =>
-      constantSerializer.serialize(c, w)
-      w.putUInt(index)
-    }
+    store.foreach { c => constantSerializer.serialize(c, w) }
   }
 
+  /**
+    * Deserializes constants.
+    * @param r reader
+    * @return self (with deserialized constants)
+    */
   def deserialize(r: ByteReader): SerializedConstantPlaceholderStore = {
+    // todo Should we rather throw here? Or better, make it part of initialization.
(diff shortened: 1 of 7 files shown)


src/main/scala/sigmastate/serialization/ValueSerializer.scala
```diff
@@ -146,13 +163,33 @@ object ValueSerializer extends SigmaSerializerCompanion[Value[SType]] {
     v
   }
 
+  private val SerializedConstantsStartCode = ConstantCode
+
   def serialize(v: Value[SType]): Array[Byte] = {
```
greenhat: @ergomorphic Should we introduce a new API for (de)serialization using placeholders (like `pack`/`unpack`)? I've made it work in `serialize(v)`/`deserialize(bytes)` treating it as an entry point for the whole tree (de)serialization.

src/main/scala/sigmastate/serialization/ValueSerializer.scala
```diff
@@ -146,13 +163,33 @@ object ValueSerializer extends SigmaSerializerCompanion[Value[SType]] {
     v
   }
 
+  private val SerializedConstantsStartCode = ConstantCode
```
greenhat: @ergomorphic I'm pretty sure we need a separate code here. It marks the serialized constants section. It works using `ConstantCode` (zero) although I sense it should be fragile.

src/main/scala/sigmastate/serialization/ValueSerializer.scala
```diff
@@ -115,9 +115,17 @@ object ValueSerializer extends SigmaSerializerCompanion[Value[SType]] {
     serializer
   }
 
+  private val ConstantPlaceholderIndexCode = ConstantCode
```
greenhat: @ergomorphic Not so sure about this one. We might need a separate code to denote an upcoming constant placeholder index.

src/main/scala/sigmastate/serialization/SerializedConstantPlaceholderStore.scala
```diff
@@ -0,0 +1,57 @@
...
+import sigmastate.SType
+import sigmastate.Values.Constant
+import sigmastate.lang.SigmaBuilder
+import sigmastate.utils.{ByteReader, ByteWriter}
+
+import scala.collection.mutable
+
+class SerializedConstantPlaceholderStore(builder: SigmaBuilder) {
+
+  private val store = mutable.Map[Constant[SType], Int]()
```
ergomorphic: @greenhat we cannot collapse equal constants from different places in the tree. They may have different semantics.

src/main/scala/sigmastate/serialization/SerializedConstantPlaceholderStore.scala
```diff
@@ -0,0 +1,57 @@
...
+  private val reversedStore = mutable.Map[Int, Constant[SType]]()
```
ergomorphic: @greenhat we just need an array here

src/main/scala/sigmastate/serialization/ValueSerializer.scala
ergomorphic: @greenhat yes it is better no have different codes

src/main/scala/sigmastate/serialization/ValueSerializer.scala
ergomorphic: @greenhat no, we don't need this code at all <num constants><constants array><script tree>

src/main/scala/sigmastate/serialization/ValueSerializer.scala
ergomorphic: @greenhat I think it is better we do additional pass, which optimizes the constants away and replaces them with placeholders.

src/main/scala/sigmastate/serialization/SerializedConstantPlaceholderStore.scala
greenhat: @ergomorphic Nice catch! Thanks. Good ol' array, how could I forget about you. :)

src/main/scala/sigmastate/serialization/ValueSerializer.scala
greenhat: @ergomorphic Do you mean introducing an IR node for constant placeholder and for constants array? I'm a bit confused.
I like the idea of ErgoScript top-level object (IR node?). Its serializer can clip constant placeholder store onto the writer/reader.

ergomorphic: @greenhat I mean we need to have some ErgoScript object which is not value. This object is responsible for top-level format of serialization. I particular it will check whether GZIP is used, how many constants in array, etc.

src/main/scala/sigmastate/serialization/SerializedConstantPlaceholderStore.scala
ergomorphic: @greenhat Easy. `HEIGHT > 5 && SELF.value > 5`

src/main/scala/sigmastate/serialization/ValueSerializer.scala
greenhat: @ergomorphic Let's assume we make:
```
object ErgoScriptSerializer {
  def serialize(script: Value[SType]): Array[Byte] = ???
  def deserialize(bytes: Array[Byte]): Value[SType] = ???
}
```
What do you mean by 

> it is better we do additional pass, which optimizes the constants away and replaces them with placeholders.

Should we replace the constants with placeholders **after** the tree is serialized with `ValueSerializer.serialize`? If not, and we first traverse the tree and replace the constant with placeholders(special value type?) then we need to teach `ValueSerializer.serialize` how to serialize placeholders. 
Both approaches look quite complicated. Why don't we parametrize `ValueSerializer` with a constant serializer (either regular or placeholder-based)? `ErgoScriptSerialize` would run then run `ValueSerializer` with a placeholder-based constant serializer.

ergomorphic: @greenhat 1) Constant separation is optional optimization, so everything should work without it.
2) This optimization can be done at compile time, so before serializer.
3) Placeholder is a special value type, which is similar to ValDef nodes. 
ValDef nodes also store an index to common subexpression, however many ValDef nodes can refer to the same subtree, whereus placeholders don't and we should have one-to-one correspondence between constants and placeholders.

greenhat:
Closing in favor of #266.

PR #266, opened 2018-10-17
New coster: Store serialized constants in separate section referenced by indices (placeholders)

### LINKED ISSUE #206: Implement (de)serialization for the remaining nodes
Added in #250 :
- `ValUse` (done in #266)
- `BlockValue` (done in #266)
- `FuncValue` (done in #266)

### LINKED ISSUE #264: A new format for optimized script execution

### LINKED ISSUE #284: Reserve 1 byte in ErgoTree serialized header for various flags
For example we may implement in future GZipping (see #263)
Or we want to change script format in ErgoScript 2.0
etc.

greenhat:
Close #264 

- [x] utilize `buildTree` to swap constants with placeholders;
- [x] add test for script without constants;
- [x] swap constants with placeholders on deserialization;
- [x] (de)serialize the whole script via `ErgoTreeSerializer`;
- [x] test with "deep tree" scripts;
- [x] add missing (de)serializer (close #206)
- [x] add header ( close #284 )
- [x] extract `ConstantPlaceholder` (de)serialization from `ValueSerializer`;
- [x] extract constants during serialization;
- [x] subclass(?) `SigmaByteReader` and move all stores (constants, valdefs) under there;
- [x] switch from `ByteReader/Writer` to `SigmaByteReader/Writer` everywhere and **remove casts**;
- [x] move `ConstantPlaceholderIndexCode` out of the constant range and remove special handling in `ValueSerializer`;
- [x] implement deserializitation of a tree preserving constant placeholders(do not inject constants);  
- [x] missing serializers for `SigmaAnd`, `SigmaOr`;
- [x] add (de)serialization round trip in `CompilerItTest` pipeline;


src/main/scala/sigmastate/serialization/ErgoScriptSerializer.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  def extractConstants(script: Value[SType]): (Seq[Constant[_]], Value[SType]) = {
+    val IR: RuntimeIRContext = new RuntimeIRContext {}
+    import IR._
+    implicit val eAny = stypeToElem(script.tpe).asInstanceOf[Elem[Context]]
+
+    val env = Map[String, Any]()
+    val IR.Pair(calcF, _) = doCosting(env, script)
+
+    val f = fun { in: Rep[Context] =>
```
greenhat: @ergomorphic What's the best approach to traversing the script tree? I'm thinking of building the graph out of it and implement the traversing and constant swapping in `fun` wrapped around the `calcF` and passing it to the `buildTree`. I feel like there should be an easier way.

src/main/scala/sigmastate/serialization/ErgoScriptSerializer.scala
```diff
@@ -0,0 +1,28 @@
...
+import sigmastate.SType
+import sigmastate.Values.{Constant, Value}
+import sigmastate.eval.RuntimeIRContext
+
+object ErgoScriptSerializer {
+  def serialize(script: Value[SType]): Array[Byte] = ???
+  def deserialize(bytes: Array[Byte]): Value[SType] = ???
+
+  def extractConstants(script: Value[SType]): (Seq[Constant[_]], Value[SType]) = {
+    val IR: RuntimeIRContext = new RuntimeIRContext {}
```
ergomorphic: @greenhat context is not lightwight object, so it is better to make is a parameter of the serializer class.

src/main/scala/sigmastate/serialization/ErgoScriptSerializer.scala
ergomorphic: @greenhat You don't need to do new traversal of the script tree. I think the constant replacement should be done in TreeBuilding.buildTree. see `case Def(Const(x)) =>`. Instead of mkConstant you need to do mkPlaceholder(nPlaceHolder). Keep track of nPlaceHolder counters and constants. 
Currently all constants in Scalan graphs are unified, this should be changed for our specific use case. It is not hard to do. So you can assume that different constants subexpressions in source script are different IR nodes in Scalan graph.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -990,6 +989,11 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting {
 //        val inputC = evalNode(ctx, env, input)
 //        withDefaultSize(inputC.value, inputC.cost + costOf(node))
 
+      case ConstantPlaceholder(_, tpe) =>
+        // todo make a proper Rep type
+        val res = toRep(null)(NothingElement.asElem[Null])
```
greenhat: @ergomorphic For swapping constant placeholders with constants on script deserialization my plan is to use `buildTree`. Seems like we need a rep type for a constant placeholder. Am I on the right track?

src/main/scala/sigmastate/eval/TreeBuilding.scala
```diff
@@ -106,29 +106,52 @@ trait TreeBuilding extends RuntimeCosting { IR: Evaluation =>
...
-  def buildValue(mainG: PGraph, env: DefEnv, s: Sym, defId: Int): SValue = {
+  def buildValue(mainG: PGraph,
+                 env: DefEnv,
+                 s: Sym,
+                 defId: Int,
+                 constantsProcessing: Option[ConstantsProcessing]): SValue = {
     import builder._
-    def recurse[T <: SType](s: Sym) = buildValue(mainG, env, s, defId).asValue[T]
-    object In { def unapply(s: Sym): Option[SValue] = Some(buildValue(mainG, env, s, defId)) }
+    def recurse[T <: SType](s: Sym) = buildValue(mainG, env, s, defId, constantsProcessing).asValue[T]
```
ergomorphic: @greenhat Why we need this parameter?

src/main/scala/sigmastate/eval/TreeBuilding.scala
```diff
@@ -106,29 +106,52 @@ trait TreeBuilding extends RuntimeCosting { IR: Evaluation =>
...
         block
       case Def(Const(x)) =>
         val tpe = elemToSType(s.elem)
-        mkConstant[tpe.type](x.asInstanceOf[tpe.WrappedType], tpe)
+        constantsProcessing match {
+          case Some(s: ExtractConstants) =>
+            val constant = mkConstant[tpe.type](x.asInstanceOf[tpe.WrappedType], tpe)
+              .asInstanceOf[ConstantNode[SType]]
+            s.extract(constant)(builder)
+          case Some(_: InjectConstants) =>
```
ergomorphic: @greenhat this looks like overcomplication. Note that we may have both constants to extract and also Placeholders from previous layers

src/main/scala/sigmastate/eval/TreeBuilding.scala
```diff
@@ -219,29 +242,59 @@ trait TreeBuilding extends RuntimeCosting { IR: Evaluation =>
...
 
-  def buildTree[T <: SType](f: Rep[Context => Any]): Value[T] = {
+  sealed trait ConstantsProcessing
+
+  class ExtractConstants() extends ConstantsProcessing {
+
+    private val store: ArrayBuffer[Constant[SType]] = new ArrayBuffer[Constant[SType]]
+
+    def extract(c: ConstantNode[SType])(implicit builder: SigmaBuilder): sigmastate.Values.ConstantPlaceholder[SType] =
+      this.synchronized {
```
ergomorphic: @greenhat does it make sense to have this thread-safe paradise island in an ocean of non-thread-safe code?

src/main/scala/sigmastate/eval/TreeBuilding.scala
```diff
@@ -219,29 +242,59 @@ trait TreeBuilding extends RuntimeCosting { IR: Evaluation =>
...
-    val rhs = buildValue(mainG, curEnv, root, curId)
+    val rhs = buildValue(mainG, curEnv, root, curId, constantsProcessing)
     val res = if (valdefs.nonEmpty) BlockValue(valdefs.toIndexedSeq, rhs) else rhs
     res
   }
 
-  def buildTree[T <: SType](f: Rep[Context => Any]): Value[T] = {
+  sealed trait ConstantsProcessing
+
+  class ExtractConstants() extends ConstantsProcessing {
```
ergomorphic: @greenhat I think we just need to have single repository of constants which is use both as the point where the constant is extracted and in all other places where we need constants

src/main/scala/sigmastate/serialization/ConstantPlaceholderSerializer.scala
```diff
@@ -0,0 +1,25 @@
...
+import sigmastate.utils.{ByteReader, ByteWriter}
+
+case class ConstantPlaceholderSerializer(cons: (Int, SType) => Value[SType])
+  extends ValueSerializer[ConstantPlaceholder[SType]] {
+
+  override val opCode: OpCode = ConstantPlaceholderIndexCode
+
+  override def serializeBody(obj: ConstantPlaceholder[SType], w: ByteWriter): Unit = {
+    w.putUInt(obj.id)
+    w.putType(obj.tpe)
```
ergomorphic: @greenhat we probably don't need to save type, as it may be obtained from ConstantStore

src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import sigmastate.SType
+import sigmastate.Values.{Constant, Value}
+import sigmastate.eval.IRContext
+import sigmastate.lang.DeserializationSigmaBuilder
+import sigmastate.utils.Extensions._
+
+import scala.collection.mutable
+
+class ErgoTreeSerializer(IR: IRContext) {
```
ergomorphic: @greenhat why we need IR?

src/main/scala/sigmastate/serialization/ValueSerializer.scala
```diff
@@ -140,10 +147,16 @@ object ValueSerializer extends SigmaSerializerCompanion[Value[SType]] {
...
     r.level = depth + 1
-    val firstByte = r.peekByte()
-    val v = if (firstByte.toUByte <= LastConstantCode) {
-      // look ahead byte tell us this is going to be a Constant
-      ConstantSerializer(builder).deserialize(r)
+    val firstByte = r.peekByte().toUByte
+    val v = if (firstByte <= LastConstantCode) {
+      if (firstByte == ConstantPlaceholderIndexCode) {
+        // skip opcode
+        r.getByte()
```
ergomorphic: @greenhat I see Placeholder as very similar to ValUse, the difference is that it takes its values from ConstantsStore, not from the current environment.

src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala
greenhat: @ergomorphic for constant extraction/injection using `buildTree`. Fixed  (extracted this feature).

src/main/scala/sigmastate/eval/TreeBuilding.scala
greenhat: @ergomorphic Nope. Old reflexes. :)

src/main/scala/sigmastate/eval/TreeBuilding.scala
greenhat: Fixed.

src/main/scala/sigmastate/eval/TreeBuilding.scala
greenhat: Done.

src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala
```diff
@@ -0,0 +1,65 @@
...
+    val constantsBuilder = mutable.ArrayBuilder.make[Constant[SType]]()
+    for (_ <- 0 until constantCount) {
+      constantsBuilder += constantSerializer.deserialize(r)
+    }
+    val constants = constantsBuilder.result
+    val tree = r.getValue()
+    (constants, tree)
+  }
+
+  def deserialize(bytes: Array[Byte]): Value[SType] = {
```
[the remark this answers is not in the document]
greenhat: Done.

src/main/scala/sigmastate/serialization/ValueSerializer.scala
greenhat: Done.

src/main/scala/sigmastate/serialization/ConstantPlaceholderSerializer.scala
greenhat: @ergomorphic Done in [this](https://github.com/ScorexFoundation/sigmastate-interpreter/pull/266/commits/ebccc8b638b5b00e2180c90bddae19845e3864a9) commit.
This makes optional `ByteReader.payload[T]` into required `ByteReader.constantStore`. `ConstantPlaceholderSerializer.parseBody()` will throw if it's missing. 
Also, since ByteReader/Writer API and implementation is due to be extracted into sigma-util does it make sense to move `.constantStore` into a subclass `SigmaByteReader`?

ergomorphic: @greenhat yes, we can use Sigma-specific interfaces derived from the base R/W interfaces as you suggested.

src/main/scala/sigmastate/serialization/ValUseSerializer.scala
```diff
@@ -0,0 +1,24 @@
...
+import sigmastate.utils.Extensions._
+import sigmastate.utils.{ByteReader, ByteWriter}
+
+case class ValUseSerializer(cons: (Int, SType) => Value[SType]) extends ValueSerializer[ValUse[SType]] {
+
+  override val opCode: OpCode = ValUseCode
+
+  override def serializeBody(obj: ValUse[SType], w: ByteWriter): Unit = {
+    w.putUInt(obj.valId)
+    w.putType(obj.tpe)
```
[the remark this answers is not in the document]
greenhat: Done in 
https://github.com/ScorexFoundation/sigmastate-interpreter/pull/266/commits/09b94859fe0003b938cd7974ffe2a16040acdde0

greenhat:
@ergomorphic What should we do with our optimization for serializing boolean constant collection (see [src](http://github.com/ScorexFoundation/sigmastate-interpreter/blob/09b94859fe0003b938cd7974ffe2a16040acdde0/src/main/scala/sigmastate/serialization/ConcreteCollectionBooleanConstantSerializer.scala#L13)). We write them as a bit array, thus we don't catch them for constant extraction on serialization. We have two options:
1. Leave as is and face cache misses for serialized tree lookups.
2. Remove this optimization, extract constants and face serialized size increase.
I'm in favor of option 1. We could fix the speed later.

ergomorphic:
@greenhat Yes, agree, we can leave the current implementation. Anyway this constant segregation is an optional optimization.

src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala
```diff
@@ -18,6 +21,23 @@ object ErgoTreeSerializer {
     w.toBytes
   }
 
+  def serialize(tree: Value[SType]): Array[Byte] = {
+    val constantStore = new ConstantStore()
+    val treeWriter = Serializer.startWriter(constantStore)
+    ValueSerializer.serialize(tree, treeWriter)
+    val extractedConstants = constantStore.getAll
```
ergomorphic: @greenhat looks like we are doing serialization twice?

greenhat: @ergomorphic Not exactly. We first extract constants and serialize the tree (with placeholders). Than, serialize just the extracted constants.

src/main/scala/sigmastate/utils/SigmaByteWriter.scala
```diff
@@ -0,0 +1,32 @@
...
+  @inline override def putUByte(x: Int): SigmaByteWriter = { super.putUByte(x); this }
+  @inline override def putBoolean(x: Boolean): SigmaByteWriter = { super.putBoolean(x); this }
+  @inline override def putShort(x: Short): SigmaByteWriter = { super.putShort(x); this }
+  @inline override def putUShort(x: Int): SigmaByteWriter = { super.putUShort(x); this }
+  @inline override def putInt(x: Int): SigmaByteWriter = { super.putInt(x); this }
+  @inline override def putUInt(x: Long): SigmaByteWriter = { super.putUInt(x); this }
+  @inline override def putLong(x: Long): SigmaByteWriter = { super.putLong(x); this }
+  @inline override def putULong(x: Long): SigmaByteWriter = { super.putULong(x); this }
+  @inline override def putBytes(xs: Array[Byte]): SigmaByteWriter = { super.putBytes(xs); this }
+  @inline override def putBits(xs: Array[Boolean]): SigmaByteWriter = { super.putBits(xs); this }
```
ergomorphic: @greenhat In scala this overrides can be avoided by using `def putULong(x: Long): this.type` in the base class.

greenhat: @ergomorphic Thank you! I totally forget about this feature.

PR #272, opened 2018-10-22
New coster: make the tests green again. Part 3.

greenhat:
fix ComplexSigSpec tests;


src/test/scala/sigmastate/TestingInterpreterSpecification.scala
```diff
@@ -195,10 +195,10 @@ class TestingInterpreterSpecification extends SigmaTestingCommons {
...
 
   property("failed numeric downcast (overflow)") {
-    an[ArithmeticException] should be thrownBy testEval("Array(999)(0).toByte > 0")
-    an[ArithmeticException] should be thrownBy testEval("Array(999)(0).toShort.toByte > 0")
-    an[ArithmeticException] should be thrownBy testEval(s"Array(${Int.MaxValue})(0).toShort > 0")
-    an[ArithmeticException] should be thrownBy testEval(s"Array(${Long.MaxValue}L)(0).toInt > 0")
+    an[IR.StagingException] should be thrownBy testEval("Array(999)(0).toByte > 0")
+    an[IR.StagingException] should be thrownBy testEval("Array(999)(0).toShort.toByte > 0")
+    an[IR.StagingException] should be thrownBy testEval(s"Array(${Int.MaxValue})(0).toShort > 0")
+    an[IR.StagingException] should be thrownBy testEval(s"Array(${Long.MaxValue}L)(0).toInt > 0")
```
ergomorphic: @greenhat we actually need to check that StagingException has caused by expected exception.

greenhat: @ergomorphic Done.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -872,6 +872,12 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting {
           case ModuloCode =>
             s = y.dataSize
             v = x.value.mod(y.value)
+          case MinCode =>
+            s = y.dataSize
```
ergomorphic: @greenhat the resulting data size for BigInteger ops depend on argument sizes.

src/main/scala/sigmastate/utxo/CostTable.scala
```diff
@@ -94,6 +102,38 @@ object CostTable {
 
     ("%", "(BigInt, BigInt) => BigInt", 0.0001),
     ("%_per_item", "(BigInt, BigInt) => BigInt", MinimalCost),
+
+    ("Downcast", "(Short) => Byte", MinimalCost),
```
ergomorphic: @greenhat we can simplify this by using generic type signature. I don't think the cost of Upcast and Downcast depend on concrete types. So we can have single rule like ("Upcast", "T => R").  See examples for array operations.

PR #273, opened 2018-10-23
Generic opType for Upcast and Downcast and min/max costing for BigInt fixes

greenhat:


src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -873,10 +873,10 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting {
             s = y.dataSize
             v = x.value.mod(y.value)
           case MinCode =>
-            s = y.dataSize
+            s = x.dataSize + y.dataSize
```
ergomorphic: @greenhat a better aproximation would be `x.dataSize max y.dataSize`

greenhat:
Fixed.

---
Best regards, 
Denys Zadorozhnyi

PR #282, opened 2018-10-26
Implement "-" prefix unary op for constants 

### LINKED ISSUE #281: Implement "-" prefix unary op for constants (`-57`, etc.)
Now we just throw on any prefix in the parser ([src](http://github.com/ScorexFoundation/sigmastate-interpreter/blob/955e5edbb308728c7c7c3d5d2f185b81ca38d121/src/main/scala/sigmastate/lang/SigmaParser.scala#L48))

greenhat:
Close #281

### DIFF
--- a/src/main/scala/sigmastate/lang/SigmaParser.scala
+++ b/src/main/scala/sigmastate/lang/SigmaParser.scala
@@ -45,7 +45,15 @@ object SigmaParser extends Exprs with Types with Core {
   implicit val logger = Logger(m => this.synchronized { logged.append(m) })
 
   def mkUnaryOp(opName: String, arg: Value[SType]) = opName match {
-    case _ => error(s"Unknown prefix operation $opName")
+    case "-" if arg.isInstanceOf[Constant[_]] && arg.tpe.isNumType =>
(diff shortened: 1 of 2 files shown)


src/main/scala/sigmastate/lang/SigmaParser.scala
```diff
@@ -45,7 +45,15 @@ object SigmaParser extends Exprs with Types with Core {
...
+    case "-" if arg.isInstanceOf[Constant[_]] && arg.tpe.isNumType =>
+      arg match {
+        case IntConstant(value) =>
+          builder.mkConstant[SInt.type](-value, SInt)
+        case LongConstant(value) =>
+          builder.mkConstant[SLong.type](-value, SLong)
+        case _ => error(s"cannot prefix $arg with op $opName")
+      }
+    case _ => error(s"Unknown prefix operation $opName for $arg")
   }
```
ergomorphic: @greenhat so, this is only parser level operation?

greenhat: Yes.

Issue #236, opened 2018-08-24
Add ZKProof function to designate zero-knowledge scope


ergomorphic:
# ZK scoping

## Motivation
We have `SigmaProp` type, `Boolean` type and there are connectives `&&`, `||` 
which allow to combine them.
In ErgoScript frontend we allow implicit conversion from `SigmaProp` to `Boolean` 
to satisfy the type checker.
For example the following script is valid, even though `&&` has `SigmaProp` 
on the left and `Boolean` on the right. 
```
((pkA || pkB) && HEIGHT > 1000) || pkC
```
For type checking this is made explicit by using `isValid` method of `SigmaProp`.
```
((pkA || pkB).isValid && HEIGHT > 1000) || pkC.isValid
```
However `isValid` function is revealing information, 
i.e. we may learn that the secret for pkC has not been provided or it was invalid.

To avoid such revealing the compiler is performing _push-up transformation_ of all 
nested `isValid` invocations.
The code above will be transformed to the following
```
(((pkA || pkB) && TrivialSigma(HEIGHT > 1000)) || pkC).isValid
```
We can call this process _knowledge hiding_ or _ZK-securing_.

`TrivialSigma` nodes represent embedding of classical boolean values to `SigmaProp` type.
At runtime `HEIGHT > 1000` term will be evaluated to `Boolean` (`true` or `false`) and then 
`TrivialSigma` will become either `TrueProof` or `FalseProof` of the sigma tree.

Unfortunately, in general such push-up transformation not always possible.
For example the following cannot be ZK-secured.
```
arr.where(x => someFunc(x).isValid)
``` 
Thus we have a conceptual problem, not every program in ErgoScript can be ZK evaluated 
due to non-interactive nature of sigma-protocols which are used.

## Solution in ErgoScript

In order to give programmer control over ZK-securing process the following function 
can be used to designate zero-knowledge scope in the program
```
ZKProof: SigmaProp => Boolean
``` 
The function is lazy in its argument, so the whole syntactic subtree will become a scope 
of zero knowledge proofs. This is how the first example can be written
```
ZKProof {
  val deadlinePassed = HEIGHT > 1000
  val aliceOrBob = (pkA || pkB)
  (aliceOrBob && deadlinePassed) || pkC
}
```
Here the sub-tree is the whole block in curly braces, so the whole script can be 
easily made into ZKProof, or alternatively only part of the script can be ZK-secured.

[...]

PR #293, opened 2018-11-02
add ZKProof  function (zero-knowledge scope)

greenhat:
Ref #236 

Todo: 
- [x] add `ZKProof` IR node and draft the parsing;
- [ ] `ZKProof` can only be at the root of the tree;
- [ ] `ZKProof` can only be single;
- [ ] Usage of `isValid` outside of `ZKProof` scope should lead to a compiler error (this is only in the presence of ZKProof scope. If it is not declared then no such limitation);
- [ ] In the scope of ZKProof no Boolean value or operation is allowed
- [ ] Closure of ZKProof shouldn't contain Boolean values.
...

### DIFF
--- a/src/main/scala/sigmastate/lang/SigmaBuilder.scala
+++ b/src/main/scala/sigmastate/lang/SigmaBuilder.scala
@@ -159,7 +159,7 @@ trait SigmaBuilder {
   def mkBlock(bindings: Seq[Val], result: Value[SType]): Value[SType]
   def mkBlockValue(items: IndexedSeq[BlockItem], result: Value[SType]): Value[SType]
   def mkValUse(valId: Int, tpe: SType): Value[SType]
-  def mkZKProofBlock(body: Value[SType]): Value[SType]
+  def mkZKProofBlock(body: Value[SSigmaProp.type]): Value[SBoolean.type]
   def mkVal(name: String, givenType: SType, body: Value[SType]): Val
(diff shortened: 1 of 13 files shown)


src/main/scala/sigmastate/lang/SigmaBuilder.scala
```diff
@@ -158,6 +158,7 @@ trait SigmaBuilder {
   def mkNoneValue[T <: SType](elemType: T): Value[SOption[T]]
 
   def mkBlock(bindings: Seq[Val], result: Value[SType]): Value[SType]
+  def mkZKProofBlock(body: Value[SType]): Value[SType]
```
ergomorphic: @greenhat  I think it should return BoolValue

src/main/scala/sigmastate/lang/Terms.scala
```diff
@@ -30,6 +30,14 @@ object Terms {
       Block(Seq(let), result)
   }
 
+  case class ZKProofBlock(body: Value[SType]) extends Value[SType] {
+    override val opCode: OpCode = OpCodes.Undefined
+    override def tpe: SType = body.tpe
+    override def cost[C <: Context](context: C): Long = ???
+    override def evaluated: Boolean = false
+    override def opType: SFunc = SFunc(Vector(), tpe)
```
ergomorphic: @greenhat It should have type SigmaProp => Boolean

PR #303, opened 2018-11-14
MapCollection, Fold, Exists, ForAll refactor (from TaggedVar to Lambda/FuncValue)

### LINKED ISSUE #300: Refactor MapCollection and Exists signatures to be like removed MapCollection1 and Exists1
They were removed in this commit - 
https://github.com/ScorexFoundation/sigmastate-interpreter/pull/266/commits/fd79bded1d4d0d9af1340ce34e21edce4a311dcd

### LINKED ISSUE #312: Refactor ForAll after ForAll1 (and remove it)

greenhat:
Close #300
Close #312

### DIFF
--- a/src/main/scala/sigmastate/eval/RuntimeCosting.scala
+++ b/src/main/scala/sigmastate/eval/RuntimeCosting.scala
@@ -966,15 +966,15 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
         }
         withDefaultSize(value, cost)
 
-      case MapCollection(input, id, mapper) =>
+      case MapCollection(input, Terms.Lambda(_, Seq((n, t)), _, Some(mapper))) =>
         val eIn = stypeToElem(input.tpe.elemType)
-        val xs = asRep[CostedCol[Any]](eval(input))
-        implicit val eAny = xs.elem.asInstanceOf[CostedElem[Col[Any],_]].eVal.eA
+        val inputC = evalNode(ctx, env, input).asRep[CostedCol[Any]]
+        implicit val eAny = inputC.elem.asInstanceOf[CostedElem[Col[Any],_]].eVal.eA
         assert(eIn == eAny, s"Types should be equal: but $eIn != $eAny")
         val mapperC = fun { x: Rep[Costed[Any]] =>
-          evalNode(ctx, env + (id -> x), mapper)
+          evalNode(ctx, env + (n -> x), mapper)
         }
-        val res = xs.mapCosted(mapperC)
+        val res = inputC.mapCosted(mapperC)
         res
 
       case Fold(input, id, zero, accId, foldOp) =>
--- a/src/main/scala/sigmastate/eval/RuntimeCosting.scala
+++ b/src/main/scala/sigmastate/eval/RuntimeCosting.scala
@@ -984,7 +984,7 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
         val res = inputC.mapCosted(mapperC)
         res
 
-      case Fold(input, zero, Terms.Lambda(_, Seq((accN, _), (n, _)), _, Some(foldOp))) =>
+      case Fold(input, zero, sfunc) =>
         val eItem = stypeToElem(input.tpe.elemType)
         val eState = stypeToElem(zero.tpe)
         (eState, eItem) match { case (eState: Elem[s], eItem: Elem[a]) =>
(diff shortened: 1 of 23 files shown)


greenhat:
@ergomorphic The idea is to use `Lambda` for mapper function instead of tagged var. But we need to have `FuncValue` in the final tree. I used `IR.buildTree` to go from `Lambda` to `FuncValue`. I've updated the compilation pipeline in `CollectionOperationsSpecification` to include `IR.buildTree` and it broke all other tests with a lot of "Don't know how to buildValue..." errors.  
Am I on the right track at all?

ergomorphic:
@greenhat Looks good. You need to implemented the corresponding case in buildValue function.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -966,15 +967,26 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
...
-      case MapCollection(input, id, mapper) =>
+      case MapCollection(input, Terms.Lambda(_, Seq((n, t)), _, Some(mapper))) =>
         val eIn = stypeToElem(input.tpe.elemType)
-        val xs = asRep[CostedCol[Any]](eval(input))
-        implicit val eAny = xs.elem.asInstanceOf[CostedElem[Col[Any],_]].eVal.eA
+        val inputC = evalNode(ctx, env, input).asRep[CostedCol[Any]]
+        implicit val eAny = inputC.elem.asInstanceOf[CostedElem[Col[Any],_]].eVal.eA
+        assert(eIn == eAny, s"Types should be equal: but $eIn != $eAny")
+        val mapperC = fun { x: Rep[Costed[Any]] =>
+          evalNode(ctx, env + (n -> x), mapper)
```
ergomorphic: @greenhat this `mapperC` should be obtained by recursively calling `eval(f)` where f is second argument.
So only one MapCollection rule is required.

greenhat: @ergomorphic I could not figure out how to do it, however, I did manage to combine both `MapCollection` and `Fold` into single cases handling both `Lambda` and `FuncValue`.

src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
```diff
@@ -276,7 +280,12 @@ class BasicOpsSpecification extends SigmaTestingCommons {
         |}""".stripMargin,
       {
         val data = GetVar(dataVar, dataType).get
-        Exists(data, 21, EQ(SelectField(TaggedVariable(21, STuple(SCollection(SByte), SLong)), 2), LongConstant(10)))
+        OR(
```
ergomorphic: @greenhat Why `exists` is transformed to `map`? Where is this happens?

src/test/scala/sigmastate/utxo/CollectionOperationsSpecification.scala
```diff
@@ -373,13 +442,31 @@ class CollectionOperationsSpecification extends SigmaTestingCommons {
...
+      If(
+        ValUse(7, SBoolean),
+        BlockValue(Vector(ValDef(8, List(), ValUse(3, SInt))), ValUse(8, SInt)).asIntValue,
+        Minus(ValUse(5, SInt), ValUse(2, SInt))),
+      None)
+val expectedPropTree = BlockValue(
+  Vector(
+    ValDef(1, List(), IntConstant(0)),
+    ValDef(2, List(), IntConstant(1)),
+    ValDef(3, List(), IntConstant(5)),
```
ergomorphic: @greenhat to increase effect of constant segregation we need to treat the constants specially and don't create ValDef even if the constant is used more than one time, because two equal constants don't always have the same meaning. 
This will not apply for GroupElement constants of ProveDlog, because in this case different constants have different meaning, thus it is ok for them to create ValDef

src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
greenhat: @ergomorphic I suppose it starts here - http://github.com/ScorexFoundation/sigmastate-interpreter/blob/c8b1773de7996455f7ac5e7279204b32f3fb3d0f/src/main/scala/sigmastate/eval/RuntimeCosting.scala#L963 
and then - http://github.com/ScorexFoundation/sigmastate-interpreter/blob/c8b1773de7996455f7ac5e7279204b32f3fb3d0f/src/main/scala/sigmastate/lang/SigmaBuilder.scala#L269

src/test/scala/sigmastate/utxo/CollectionOperationsSpecification.scala
greenhat: Fixed.

src/test/scala/sigmastate/utxo/CollectionOperationsSpecification.scala
```diff
@@ -373,13 +442,31 @@ class CollectionOperationsSpecification extends SigmaTestingCommons {
...
-    val indexId = 21.toByte
-    val index = TaggedInt(indexId)
-    val element = ByIndex(string, If(LE(index, 0), 5, Minus(index, 1)))
-    val elementRule = BinOr(EQ(element, 0), EQ(element, 1))
-    val expectedPropTree = ForAll(indexCollection, indexId, elementRule)
+    val element = ByIndex(
+      ValUse(4, SCollectionType(SInt)),
+      If(
+        ValUse(7, SBoolean),
+        BlockValue(Vector(ValDef(8, List(), ValUse(3, SInt))), ValUse(8, SInt)).asIntValue,
```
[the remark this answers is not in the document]
greenhat: Fixed.

src/test/scala/sigmastate/utxo/CollectionOperationsSpecification.scala
```diff
@@ -349,12 +398,32 @@ class CollectionOperationsSpecification extends SigmaTestingCommons {
...
+          LE(boundaryIndex, ValUse(3, SInt))
+        )
+      )
+    )
+    val expectedPropTree = BlockValue(
+      Vector(
+        ValDef(1,List(),IntConstant(0)),
+        ValDef(2,List(),IntConstant(1)),
+        ValDef(3,List(),IntConstant(5))),
+       AND(
```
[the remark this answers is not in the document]
greenhat: I suspect `hasManyUsagesGlobal()`. Seems like more than zero will trigger extraction. Anyway, this case is fixed by skipping constants in `ValDef` extraction.

src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
```diff
@@ -380,8 +391,17 @@ class BasicOpsSpecification extends SigmaTestingCommons {
...
+      BlockValue(
+        Vector(ValDef(1, IntConstant(3))),
+        EQ(
+          ExtractRegisterAs(Self, R6, SOption(SInt)).getOrElse(
+            BlockValue(
+              Vector(ValDef(2, ValUse(1, SInt))),
+              ValUse(2, SInt)).asIntValue
+          ),
+        ValUse(1, SInt))
+    ),
```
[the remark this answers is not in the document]
ergomorphic: Added #313 for Mainnet

PR #304, opened 2018-11-14
Use public snapshot versions of special and special.sigma

greenhat:


lock.sbt
```diff
@@ -17,6 +17,16 @@ dependencyOverrides in ThisBuild ++= Seq(
   "com.typesafe.akka" % "akka-actor_2.12" % "2.4.20",
   "com.typesafe.scala-logging" % "scala-logging_2.12" % "3.9.0",
   "commons-io" % "commons-io" % "2.5",
+  "io.github.scalan" % "common_2.12" % "snapshot-publish-b32356d5-SNAPSHOT",
+  "io.github.scalan" % "core_2.12" % "snapshot-publish-b32356d5-SNAPSHOT",
+  "io.github.scalan" % "library-api_2.12" % "snapshot-publish-b32356d5-SNAPSHOT",
+  "io.github.scalan" % "library-impl_2.12" % "snapshot-publish-b32356d5-SNAPSHOT",
+  "io.github.scalan" % "library_2.12" % "snapshot-publish-b32356d5-SNAPSHOT",
+  "io.github.scalan" % "macros_2.12" % "snapshot-publish-b32356d5-SNAPSHOT",
```
aslesarenko: @greenhat hm, it is interesting why this dependency has crept in

ergomorphic: @greenhat hm, it is interesting why this dependency has crept in

PR #315, opened 2018-11-29
Check max size of big int in ByteArrayToBigInt

### LINKED ISSUE #311: Check max size of BigInt
Byte array => BigInt conversion currently does not check that BigInt is no more than 256 bits, more, BigInt should be no more than order of the group we're using (CryproConstants.dlogGroup.order). This should be fixed.

greenhat:
Close #311 

- [x] look into `OracleExamplesSpecification` test fail;

### DIFF
--- a/lock.sbt
+++ b/lock.sbt
@@ -24,9 +24,9 @@ dependencyOverrides in ThisBuild ++= Seq(
   "io.github.scalan" % "library_2.12" % "master-98f4f086-SNAPSHOT",
   "io.github.scalan" % "macros_2.12" % "master-98f4f086-SNAPSHOT",
   "io.github.scalan" % "meta_2.12" % "master-98f4f086-SNAPSHOT",
-  "io.github.scalan" % "sigma-api_2.12" % "master-c916619f-SNAPSHOT",
-  "io.github.scalan" % "sigma-impl_2.12" % "master-c916619f-SNAPSHOT",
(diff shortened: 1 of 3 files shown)


greenhat:
@kushti I've added a runtime check for ByteArrayToBigInt operation to test if resulted big int exceeds dlog group order(required in #311). It brakes the `OracleExampleSpec` since blake hash converted to big int here -
here http://github.com/ScorexFoundation/sigmastate-interpreter/blob/a11b15927bbf2a8e92751b0c099d0b635608b456/src/test/scala/sigmastate/utxo/examples/OracleExamplesSpecification.scala#L138

and the resulted big int is larger than dlog group order. The runtime check is done in http://github.com/aslesarenko/sigma/blob/4433e65069a0e978c030dcecf4da64a8db66522b/sigma-impl/src/main/scala/special/sigma/SigmaDslOverArrays.scala#L204

Should the `OracleExampleSpec` be changed? Would taking only part of the hash will hurt the entire example here?

kushti:
@greenhat as was discussed with @ergomorphic , we need to add modular arithmetic operators, namely, unary .modQ operator (in ErgoScript, in ErgoTree it would be like ModQ(ByteArrayToBigInt(...)), and also binary plusModQ/minusModQ at least. To fix the example now, we can construct the big int in the following way: Array(0: Byte) ++ blake2b256(...).take(31)), 248 random bits instead of 252 looks okay.

greenhat:
Fixed the oracle example and extracted modular arithmetic operators into #327

PR #317, opened 2018-11-30
Switch the rest of the tests to the compileWithCosting

### LINKED ISSUE #294: Make atLeast to operate on SigmaProp values
Currently atLeast is defined as function `atLeast: Array[Boolean] => Boolean`.
However it operates on values of type SigmaProp.
Thus its type should be changed to `atLeast: Array[SigmaProp] => SigmaProp
ThresholdSpecification should be fixed accordingly.

### LINKED ISSUE #314: Switch the tests to the new compiler pipeline (with coster)

greenhat:
Close #314
Close #294 

Todo:
- [x] remove `SigmaSpecializer.specialize` from `compileWithCosting`;
- [x] make `AtLeast` to operate on SigmaProps;
- [x] look into failing `anyOf(Col(pubkeyA, pubkeyB, HEIGHT > 500))` ComplexSigSpec test (re-arranged args); 
- [x] switch tests(mostly);
- [x] `CostingBox` leaking through the `TreeBuilding`;

### DIFF
--- a/src/main/scala/sigmastate/Values.scala
+++ b/src/main/scala/sigmastate/Values.scala
@@ -647,7 +647,7 @@ object Values {
     lazy val tpe: SFunc = SFunc(args.map(_._2), body.tpe)
     val opCode: OpCode = FuncValueCode
     /** This is not used as operation, but rather to form a program structure */
-    def opType: SFunc = Value.notSupportedError(this, "opType")
+    override def opType: SFunc = SFunc(Vector(), tpe)
   }
   object FuncValue {
     def apply(argId: Int, tArg: SType, body: SValue): FuncValue =
--- a/src/main/scala/sigmastate/Values.scala
+++ b/src/main/scala/sigmastate/Values.scala
@@ -275,7 +275,7 @@ object Values {
     }
   }
 
-  implicit class AvlTreeConstantOps(c: AvlTreeConstant) {
+  implicit class AvlTreeConstantOps(val c: AvlTreeConstant) extends AnyVal {
     def createVerifier(proof: SerializedAdProof) =
       new BatchAVLVerifier[Digest32, Blake2b256.type](
         c.value.startingDigest,
@@ -358,7 +358,7 @@ object Values {
     }
   }
 
-  implicit class CollectionConstantOps[T <: SType](c: CollectionConstant[T]) {
+  implicit class CollectionConstantOps[T <: SType](val c: CollectionConstant[T]) extends AnyVal {
     def toConcreteCollection: ConcreteCollection[T] = {
       val tElem = c.tpe.elemType
       val items = c.value.map(v => tElem.mkConstant(v.asInstanceOf[tElem.WrappedType]))
@@ -555,7 +555,7 @@ object Values {
 
   trait LazyCollection[V <: SType] extends NotReadyValue[SCollection[V]]
 
-  implicit class CollectionOps[T <: SType](coll: Value[SCollection[T]]) {
+  implicit class CollectionOps[T <: SType](val coll: Value[SCollection[T]]) extends AnyVal {
     def length: Int = matchCase(_.items.length, _.value.length, _.items.length)
     def items = matchCase(_.items, _ => sys.error(s"Cannot get 'items' property of node $coll"), _.items)
 //    def isEvaluatedCollection =
@@ -578,17 +578,17 @@ object Values {
       )
   }
 
-  implicit class SigmaPropValueOps(p: Value[SSigmaProp.type]) {
+  implicit class SigmaPropValueOps(val p: Value[SSigmaProp.type]) extends AnyVal {
     def isProven: Value[SBoolean.type] = SigmaPropIsProven(p)
     def propBytes: Value[SByteArray] = SigmaPropBytes(p)
   }
 
-  implicit class SigmaBooleanOps(sb: SigmaBoolean) {
+  implicit class SigmaBooleanOps(val sb: SigmaBoolean) extends AnyVal {
     def isProven: Value[SBoolean.type] = SigmaPropIsProven(SigmaPropConstant(sb))
     def propBytes: Value[SByteArray] = SigmaPropBytes(SigmaPropConstant(sb))
   }
 
-  implicit class BoolValueOps(b: BoolValue) {
+  implicit class BoolValueOps(val b: BoolValue) extends AnyVal {
     def toSigmaProp: SigmaPropValue = BoolToSigmaProp(b)
   }
 
@@ -658,7 +658,7 @@ object Values {
       FuncValue(IndexedSeq((argId,tArg)), body)
   }
 
-  implicit class OptionValueOps[T <: SType](p: Value[SOption[T]]) {
+  implicit class OptionValueOps[T <: SType](val p: Value[SOption[T]]) extends AnyVal {
     def get: Value[T] = OptionGet(p)
     def getOrElse(default: Value[T]): Value[T] = OptionGetOrElse(p, default)
     def isDefined: Value[SBoolean.type] = OptionIsDefined(p)
--- a/src/main/scala/sigmastate/Values.scala
+++ b/src/main/scala/sigmastate/Values.scala
@@ -588,6 +588,10 @@ object Values {
     def propBytes: Value[SByteArray] = SigmaPropBytes(SigmaPropConstant(sb))
   }
 
+  implicit class BoolValueOps(b: BoolValue) {
+    def toSigmaProp: SigmaPropValue = BoolToSigmaProp(b)
+  }
+
   sealed trait BlockItem extends NotReadyValue[SType] {
     def id: Int
     def rhs: SValue
--- a/src/main/scala/sigmastate/eval/CompiletimeCosting.scala
+++ b/src/main/scala/sigmastate/eval/CompiletimeCosting.scala
@@ -11,6 +11,7 @@ import sigmastate.lang.Terms.{Apply, _}
 import sigmastate.lang.SigmaPredef._
 import sigmastate.utxo._
 import sigmastate.SType._
+import sigmastate.SCollection._
 import sigmastate.Values.Value.Typed
 import sigmastate.lang.SigmaSpecializer.error
 import sigmastate.lang.{Terms, TransformingSigmaBuilder}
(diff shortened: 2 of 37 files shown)


src/test/scala/sigmastate/utxo/ErgoLikeInterpreterSpecification.scala
```diff
@@ -31,8 +31,8 @@ class ErgoLikeInterpreterSpecification extends SigmaTestingCommons {
 
     val ctx = ErgoLikeContext.dummy(fakeSelf)
 
-    val e = compile(Map("h1" -> h1.bytes, "h2" -> h2.bytes), "h1 == h1")
-    val exp = EQ(ByteArrayConstant(h1.bytes), ByteArrayConstant(h1.bytes))
+    val e = compileWithCosting(Map("h1" -> h1.bytes, "h2" -> h2.bytes), "h1 == h1")
+    val exp = TrueLeaf
```
ergomorphic: @greenhat I think this is the result of constant propagation in Special.
It can be [turned off](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/7ba836ae5ddb9d915cfa4b9e68d83fe93bcdda0a/src/main/scala/sigmastate/eval/IRContext.scala#L15)

greenhat: @ergomorphic Thanks for the hint.  Nope, enabling constant propagation does not change this case, moreover, it breaks `1 + 1` case here http://github.com/ScorexFoundation/sigmastate-interpreter/blob/b9ccb7e8d05e95e5c3183f46dc9b81790f0b685b/src/test/scala/sigmastate/eval/ErgoTreeBuildingTest.scala#L28 (it produces `IntConstant(2)` instead of `Plus(...)`)

src/main/scala/sigmastate/eval/CompiletimeCosting.scala
```diff
@@ -50,9 +50,18 @@ trait CompiletimeCosting extends RuntimeCosting { IR: Evaluation =>
...
         eval(mkLongToByteArray(arg))
 
-      case sigmastate.Upcast(Constant(value, tpe), toTpe: SNumericType) =>
+      case Terms.Apply(ByteArrayToBigIntSym, Seq(arr: Value[SByteArray]@unchecked)) =>
+        eval(mkByteArrayToBigInt(arr))
+
+      case Terms.Apply(SigmaPropSym, Seq(bool: Value[SBoolean.type]@unchecked)) =>
+        eval(mkBoolToSigmaProp(bool))
+
+      case sigmastate.Upcast(Constant(value, _), toTpe: SNumericType) =>
```
ergomorphic: @greenhat I think we can do upcast and downcast not only constants but result of any expression of Numeric type.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -945,14 +945,7 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
...
         }
-        val condC = node.condition match {
-          case Terms.Lambda(_, Seq((id, _)), _, Some(body)) =>
-            fun { x: Rep[Costed[Any]] => evalNode(ctx, env + (id -> x), body) }
-          case FuncValue(Seq((id, _)), body) =>
-            fun { x: Rep[Costed[Any]] => evalNode(ctx, env + (id -> x), body) }
-          case v =>
-            evalNode(ctx, env, v).asRep[CostedFunc[Unit, Any, SType#WrappedType]].func
-        }
+        val condC = evalNode(ctx, env, node.condition).asRep[CostedFunc[Unit, Any, SType#WrappedType]].func
```
ergomorphic: @greenhat I use method asRep(...) instead of .toRep because the later comes with unnecessary allocation of wrapper implicit class.

src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
```diff
@@ -272,12 +272,11 @@ class BasicOpsSpecification extends SigmaTestingCommons {
         |}""".stripMargin,
       {
         val data = GetVar(dataVar, dataType).get
-        OR(
-          MapCollection(data,
-            FuncValue(
-              Vector((1, STuple(SByteArray, SLong))),
-              EQ(SelectField(ValUse(1, STuple(SByteArray, SLong)), 2), LongConstant(10)))
-          ).asCollection[SBoolean.type])
+        Exists(data,
```
greenhat: @ergomorphic `Exists` and `ForAll` are back. Made them in http://github.com/ScorexFoundation/sigmastate-interpreter/blob/94fed4472ba0b8ba09f8cea2ad54abefa9816e61/src/main/scala/sigmastate/eval/RuntimeCosting.scala#L955

src/main/scala/sigmastate/eval/RuntimeCosting.scala
greenhat: @ergomorphic Fixed.

src/main/scala/sigmastate/eval/CompiletimeCosting.scala
ergomorphic: @greenhat no I mean it can be just any Value of Numeric type like `Upcast(Plus(IntConst(10), IntConst(20)), SLong)`. Why not?

src/main/scala/sigmastate/types.scala
```diff
@@ -427,6 +427,7 @@ case object SSigmaProp extends SProduct with SPrimType with SEmbeddable with SLo
           GroupElementConstant(gv), GroupElementConstant(hv),
           GroupElementConstant(uv), GroupElementConstant(vv)) =>
       SGroupElement.dataSize(gv.asWrappedType) * 4 + 1
+    case CAND(inputs) => inputs.map(i => dataSize(i.asWrappedType)).sum
```
greenhat: @ergomorphic Could you please check my costing for `CAND`?

src/main/scala/sigmastate/Values.scala
```diff
@@ -588,6 +588,10 @@ object Values {
     def propBytes: Value[SByteArray] = SigmaPropBytes(SigmaPropConstant(sb))
   }
 
+  implicit class BoolValueOps(b: BoolValue) {
```
ergomorphic: @greenhat It is better to make all implicit classes `extends AnyVal` to avoid allocations

src/main/scala/sigmastate/types.scala
ergomorphic: @greenhat looks good.

src/main/scala/sigmastate/Values.scala
greenhat: Done.

src/main/scala/sigmastate/eval/CostingDataContext.scala
```diff
@@ -43,7 +44,9 @@ class CostingBox(
     bytesWithoutRef: Col[Byte],
     propositionBytes: Col[Byte],
     registers: Col[AnyValue],
-    var isCost: Boolean) extends TestBox(id, value, bytes, bytesWithoutRef, propositionBytes, registers) {
+    var isCost: Boolean,
+    val ergoBox: ErgoBox
```
greenhat: @ergomorphic I've added an instance of `ErgoBox` to `CostingBox` rather than add missing properties. It might look a bit dirty, but the alternative is to add all `ErgoBox` properties to `CostingBox` (except `value`) I like less. 

This needed for pulling boxes through TreeBuilding.

greenhat: Removed as being too dirty. Created https://github.com/ScorexFoundation/sigmastate-interpreter/issues/325

src/main/scala/sigmastate/Values.scala
```diff
@@ -647,14 +651,14 @@ object Values {
     lazy val tpe: SFunc = SFunc(args.map(_._2), body.tpe)
     val opCode: OpCode = FuncValueCode
     /** This is not used as operation, but rather to form a program structure */
-    def opType: SFunc = Value.notSupportedError(this, "opType")
+    override def opType: SFunc = SFunc(Vector(), tpe)
```
ergomorphic: @greenhat this method is not supposed to be used. It is not defined for this node.

src/main/scala/sigmastate/eval/CompiletimeCosting.scala
```diff
@@ -112,9 +141,34 @@ trait CompiletimeCosting extends RuntimeCosting { IR: Evaluation =>
         val index = fn.substring(1).toByte
         eval(mkSelectField(tuple.asTuple, index))
 
-      case Terms.Apply(Select(col, "slice", _), Seq(from, until)) =>
+      case Select(obj, method, Some(tRes: SNumericType))
+        if obj.tpe.isNumType && obj.asNumValue.tpe.isCastMethod(method) =>
```
ergomorphic: @greenhat Please use Array in `isCastMethod` and also avoid allocation on every method call.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -814,6 +814,9 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
         case box: ErgoBox =>
           val boxV = liftConst(box.toTestBox(false)(IR))
           RCCostedBox(boxV, costOf(c))
+        case box: special.sigma.Box =>
```
ergomorphic: @greenhat This is strange. In which example this case is required?

src/main/scala/sigmastate/Values.scala
greenhat: I've used it in `RuntimeCosting` for `FuncValue` [here, in `costOf()`](http://github.com/ScorexFoundation/sigmastate-interpreter/blob/ad98d80a40e119097a4da0f7e0bab6d63e93e7e3/src/main/scala/sigmastate/eval/RuntimeCosting.scala#L1296) . I did not find a better way than to "enable" it.

src/main/scala/sigmastate/eval/CompiletimeCosting.scala
greenhat: Fixed.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
greenhat: In array ops tests in `TestingInterpreterSpec` (access box in env). It will be resolved by #325 so I rewrote tests to avoid using a box ([see](
https://github.com/ScorexFoundation/sigmastate-interpreter/pull/317/commits/ffc1048c03fdb55c9f34c4737c55b5fb8521e625)).

PR #318, opened 2018-11-30
Add (de)serializer with generic Reader/Writer API

### LINKED ISSUE #316: Rewrite SigmaByteReader/Writer into a Decorator
Expose a generic ByteReader/Writer.

greenhat:
Close #316 

- [ ] test new API in ergo (branch PR from `new-serialization` branch);
- [x] implement;

### DIFF
--- a/src/main/scala/sigmastate/utils/SigmaByteWriter.scala
+++ b/src/main/scala/sigmastate/utils/SigmaByteWriter.scala
@@ -1,17 +1,66 @@
 package sigmastate.utils
 
-import scorex.util.ByteArrayBuilder
-import scorex.util.serialization.VLQByteBufferWriter
+import scorex.util.serialization.{ByteStringWriter, VLQByteBufferWriter, Writer}
+import scorex.util.serialization.Writer.Aux
 import sigmastate.SType
 import sigmastate.Values.Value
 import sigmastate.serialization.{ConstantStore, TypeSerializer, ValueSerializer}
 
-class SigmaByteWriter(b: ByteArrayBuilder,
-                      val constantExtractionStore: Option[ConstantStore]) extends VLQByteBufferWriter(b) {
+// todo rename to SigmaWriter?
+class SigmaByteWriter(val w: Writer,
+                      val constantExtractionStore: Option[ConstantStore]) extends Writer {
 
-  @inline def putType[T <: SType](x: T): SigmaByteWriter = { TypeSerializer.serialize(x, this); this }
-  @inline def putValue[T <: SType](x: Value[T]): SigmaByteWriter = { ValueSerializer.serialize(x, this); this }
-  @inline def putValues[T <: SType](xs: Seq[Value[T]]): SigmaByteWriter = {
+  type CH = w.CH
+
+  override def length(): Int = w.length()
+
+  override def newWriter(): Aux[CH] = w.newWriter()
+
+  override def putChunk(chunk: CH): this.type = { w.putChunk(chunk); this }
+
+  override def result(): CH = w match {
+    case wr: ByteStringWriter => wr.result().asInstanceOf[CH]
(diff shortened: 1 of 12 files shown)


ergomorphic:
@greenhat I think Decoration with Delegation relates purely to Writer/Reader. Serializer should not change.

greenhat:
@ergomorphic It’s only the external API so far. My plan is to “wrap” passed generic reader/writer into our SigmaReader/Writer for use in our serializers

src/main/scala/sigmastate/utils/SigmaByteWriter.scala
```diff
@@ -1,14 +1,63 @@
...
+import scorex.util.serialization.{ByteStringWriter, VLQByteBufferWriter, Writer}
+import scorex.util.serialization.Writer.Aux
 import sigmastate.SType
 import sigmastate.Values.Value
 import sigmastate.serialization.{ConstantStore, TypeSerializer, ValueSerializer}
 
-class SigmaByteWriter(b: ByteArrayBuilder,
-                      val constantExtractionStore: Option[ConstantStore]) extends VLQByteBufferWriter(b) {
+// todo rename to SigmaWriter?
+class SigmaByteWriter(val w: Writer,
```
greenhat: @ergomorphic Check out my first take on rewriting SigmaByteWriter to be a decorator for the underlying writer. All test are passing (except for new ergo serializer wrappers).

src/main/scala/sigmastate/utils/SigmaByteWriter.scala
```diff
@@ -1,14 +1,63 @@
...
+  type CH = w.CH
+
+  override def length(): Int = w.length()
+
+  override def newWriter(): Aux[CH] = w.newWriter()
+
+  override def putChunk(chunk: CH): this.type = { w.putChunk(chunk); this }
+
+  override def result(): CH = w match {
+    case wr: ByteStringWriter => wr.result().asInstanceOf[CH]
```
ergomorphic: @greenhat Why not just always use `wr.result().asInstanceOf[CH]`?

src/main/scala/sigmastate/utils/SigmaByteWriter.scala
```diff
@@ -1,14 +1,63 @@
...
+    this
+  }
+
+  def putShortString(s: String): this.type = { w.putShortString(s); this }
+
+  // todo move to Writer
+  def toBytes: Array[Byte] = w match {
+    case wr: ByteStringWriter => wr.result().asByteBuffer.array()
+    case wr: VLQByteBufferWriter => wr.toBytes
+  }
```
ergomorphic: @greenhat I see decoration as adding new "kind of high-level app-specific" API above basic API. This means that decorator doesn't have it's own bytes. Thus all bytes we have written are in  `w` so we can just call `w.toBytes`

src/main/scala/sigmastate/utils/SigmaByteWriter.scala
greenhat: @ergomorphic Yep. I missed that. Thanks.

src/main/scala/sigmastate/utils/SigmaByteWriter.scala
greenhat: @ergomorphic My thoughts exactly, but `Writer` does not have `toBytes`, so that's what my todo comment is about.

greenhat: @VictorMikheev Could you please add `Writer.toBytes`? Thanks.

ergomorphic: @greenhat this probably deserves an issue at scorex-utils

greenhat: @ergomorphic Done. https://github.com/ScorexFoundation/scorex-util/issues/9

PR #322, opened 2018-12-06
Atomic single chain tests

### LINKED ISSUE #298: More atomic single-chain trades tests
Implement more tests around atomic single-chain trades (AssetsAtomicExchangeSpecification.scala):

1. Failure scenarios (when some conditions are met)
2. Successful execution of several orders (w. partial filling for some of them). 
3. Show impossibility of the following attacks against a previous construction: 

"
We caution that for security, Alice cannot have two such boxes outstanding at any given time, because they can be both used in a single transaction that contains just one output with 60 tokens of type ``token1" --- the script of each box will be individually satisfied, but Alice will get less only half of what owed to her. Same for Bob.

this seems like a pretty big security vulnerability and we should highlight it and show how to fix it. Restricting the number of inputs and outputs to just 2 is not enough. For example, suppose Alice has another box C in the UTXO with value 100 Ergo and 60 token1, and the following script:

  (HEIGHT > deadline && pkA) || {
    val tokenData = OUTPUTS(0).R2[Array[(Array[Byte], Long)]].get(0)
    allOf(Array(
        tokenData._1 == token1,
        tokenData._2 >= 60L,
        OUTPUTS(0).propositionBytes == pkA.propBytes,
        OUTPUTS(0).value >= 100L
    ))
  } 
  || ...

Basically, this script allows Alice to simply reclaim what she has in B; it may also also her to do other things, which is why  "|| ...'' is there in the script. I don't know the purpose of such a script, but it seems plausible. Now, an adversary can create a transaction with two inputs: Alice's box above (which requires to give her 60 token1 in the output) and box C; and two outputs: output with value 100 Ergo + 60 token1 to Alice, and output with value 100 Ergo to the adversary). Alice just lost 100 Ergo."

oskin1:
Close #298

### DIFF
--- a/src/test/scala/sigmastate/utxo/examples/AssetsAtomicExchangeSpecification.scala
+++ b/src/test/scala/sigmastate/utxo/examples/AssetsAtomicExchangeSpecification.scala
@@ -187,9 +187,13 @@ class AssetsAtomicExchangeSpecification extends SigmaTestingCommons {
         |
         |  val outIdx = getVar[Short](127).get
         |  val out = OUTPUTS(outIdx)
+        |
+        |  val defaultTokenData = Array((Array(1.toByte), 0L))
+        |
         |  val tokenData = out.R2[Array[(Array[Byte], Long)]].get(0)
         |  val tokenId = tokenData._1
         |  val tokenValue = tokenData._2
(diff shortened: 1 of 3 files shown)


aslesarenko:
@oskin1 please actualize and rebase on top of `develop`

PR #331, opened 2018-12-14
Check AtLeast input size

### LINKED ISSUE #291: Constraint on the number of children of atLeast at compile time
See `require(children.length <= 255)` in `AtLeast.reduce`  method   
`// TODO: this constraint on the number of children of atLeast should also be ensured at ErgoScript compile time`

greenhat:
Close #291

### DIFF
--- a/src/main/scala/sigmastate/eval/RuntimeCosting.scala
+++ b/src/main/scala/sigmastate/eval/RuntimeCosting.scala
@@ -1155,6 +1155,9 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
 
       case AtLeast(bound, input) =>
         val inputC = asRep[CostedCol[Any]](evalNode(ctx, env, input))
+        val inputCount = inputC.values.length.asValue
+        if (inputCount > AtLeast.MaxChildrenCount)
+          error(s"Expected input elements count should not exceed ${AtLeast.MaxChildrenCount}, actual: $inputCount")
         val boundC = eval(bound)
         val res = sigmaDslBuilder.atLeast(boundC.value, asRep[Col[SigmaProp]](inputC.values))
         val cost = boundC.cost + inputC.cost + costOf(node)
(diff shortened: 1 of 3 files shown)


src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -1155,6 +1155,9 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
 
       case AtLeast(bound, input) =>
         val inputC = asRep[CostedCol[Any]](evalNode(ctx, env, input))
+        val inputCount = inputC.values.length.asValue
```
ergomorphic: @greenhat This usage of `asValue` is not always safe. It will only work if collection is constructed by using `fromItems` method. However this will not always be the case. Check out how this function is implemented and extract constant only when there is a constant. Otherwise do nothing.

src/test/scala/sigmastate/utxo/ThresholdSpecification.scala
```diff
@@ -400,4 +401,21 @@ class ThresholdSpecification extends SigmaTestingCommons {
...
+    val pubKeysStrExceeding = Array.fill[String](AtLeast.MaxChildrenCount + 1)(keyName).mkString(",")
+    an[CosterException] should be thrownBy compileWithCosting(env, s"""atLeast(2, Coll($pubKeysStrExceeding))""")
+    an[CosterException] should be thrownBy
+      compileWithCosting(env, s"""{ val arr = Coll($pubKeysStrExceeding); atLeast(2, arr) }""")
+
+    // max children should work fine
+    val pubKeysStrMax = Array.fill[String](AtLeast.MaxChildrenCount)(keyName).mkString(",")
+    compileWithCosting(env, s"""atLeast(2, Coll($pubKeysStrMax))""")
+    compileWithCosting(env, s"""{ val arr = Coll($pubKeysStrMax); atLeast(2, arr) }""")
+  }
```
ergomorphic: Please add test where `atLeast` uses some collection with unknown items, for example from `getVar`

PR #348, opened 2018-12-27
More scripts in ErgoScriptPredef

catena2w:


src/main/scala/org/ergoplatform/Emission.scala
```diff
@@ -0,0 +1,85 @@
...
+
+object Emission {
+
+  val CoinsInOneErgo: Long = 1000000000
+
+  /**
+    * Emission rules.
+    * Return number of coins, issued at height `h` and all previous heights
+    */
+  def issuedCoinsAfterHeight(h: Int,
```
ergomorphic: @catena2w consider renaming to `issuedCoinsAtHeight`, this is also consistent with the next method.

src/test/scala/org/ergoplatform/EmissionSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import org.scalacheck.Gen
+import sigmastate.helpers.SigmaTestingCommons
+
+import scala.annotation.tailrec
+
+class EmissionSpec extends SigmaTestingCommons {
+
+  val genesisHeight = 1
+  val blocksTotal = 2080799
```
ergomorphic: It is better to uncover the formulas of these magic numbers.

catena2w:
It should be already merged in v2.0

PR #347, opened 2018-12-27
Add missing non-method ops (typer layer)

greenhat:
#324 

Close #349 

Todo:
- [x] convert existing predefined global functions to use `PredefinedFunc`;
- [x] implement new predefined global functions (see spec);
- [x] fix ergo (pass network prefix to compiler);

Issue #324, opened 2018-12-06
Add missing operations


ergomorphic:
#### New operations in SigmaPredef
See [spec](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/v2.0/docs/LangSpec.md#PredefinedFunctions)

#### New methods for various objects
See [spec](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/v2.0/docs/LangSpec.md#DataTypes)

#### Unary operations (// Testnet2)
```scala
! : SBoolean => SBoolean  // logical Not 
- : SNumeric => SNumeric  // negation operation
~ : SNumeric => SNumeric  // bitwise inversion of all bits
```

#### Bitwise operations (// Testnet2)
```scala
^ : (SBoolean, SBoolean) => SBoolean // logical XOR  

| : (SNumeric, SNumeric) => SNumeric // bitwise OR for all numeric types 
& : (SNumeric, SNumeric) => SNumeric // bitwise AND for all numeric types 
^ : (SNumeric, SNumeric) => SNumeric // bitwise XOR for all numeric types 

/** Returns this value bit-shifted right by the specified number of bits, 
filling in the left bits with the same value as the left-most bit of this. */
>> : (SNumeric, SNumeric) => SNumeric 

/** Returns this value bit-shifted left by the specified number of bits, 
filling in the new right bits with zeroes. */
<< : (SNumeric, SNumeric) => SNumeric 

/** Returns this value bit-shifted right by the specified number of bits, 
filling in the new left bits with zeroes. */
>>> : (SNumeric, SNumeric) => SNumeric 

/** Returns this collection bit-shifted left by the specified number of bits, 
    filling in the new right bits with zeroes. The size of collection is preserved. */
<< : (Coll[Boolean], SInt) => Coll[Boolean]

/** Returns this collection bit-shifted right by the specified number of bits, 
filling in the left bits with the same value as the left-most bit of this. 
The size of collection is preserved.*/
>> : (Coll[Boolean], SInt) => Coll[Boolean]

/** Returns this collection bit-shifted right by the specified number of bits, 
filling in the new left bits with zeroes. The size of collection is preserved.*/
>>> : (Coll[Boolean], SInt) => Coll[Boolean]

```

#### Operation to be replaces/renamed
Existing `Exponentiate` operation use symbol `^` which should be used for bitwise XOR instead.
See method `SGroupElement.exp` above.

[...]

PR #358, opened 2019-01-10
Implement missing type methods (typer layer)

### LINKED ISSUE #344: Make SMethod a single point of declaration
For methods we need to improve infrastructure. 
We need to extends SMethod, so that it can be a single point of method declaration. 
And all other code should just use info from SMethod. 
This include Parser/Typer/Coster/TreeBuilder.

The goal:
To add a new method we just need to register new SMethod for the corresponding type.

greenhat:
#324 
Close #344

Todo: 
- [x] "shifting" SCollection methods;
- [x] rebase onto`v2.0` after #347 is merged;
- [x] methods from other types;
- [x] the rest of missing SCollection methods;
- [x] `SCollection.zip`, `partition` type cannot be expressed (circular dependency on `STuple`);
- [x]  `SCollection.mapReduce`;
- [ ] move existing IR building code from `CompiletimeCosting` into `SMethod.irBuilder`;
- [x] try pulling a few `MethodCall` based collection method through the costing; 
- [x] test new methods (`MethodCall` based) pulled through the costing in interpreter;
- [ ] add `costOf(node)` for `MethodCall` ops (RuntimeCosting);
- [ ] move to `special.sigma` appending of the result type of the lambda to arg list on `MethodCall` eval and also for other methods with RType implicit argument (if any) [cmt](https://github.com/ScorexFoundation/sigmastate-interpreter/pull/358/files#r252321622)

### DIFF
--- a/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
+++ b/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
@@ -9,7 +9,7 @@ import scalan.RType.{TupleType, PairType}
 import sigmastate.Values._
 import sigmastate._
 import sigmastate.eval._
-import sigmastate.interpreter.{ContextExtension, Context}
+import sigmastate.interpreter.{ContextExtension, Context => ErgoContext}
 import sigmastate.serialization.OpCodes
 import sigmastate.serialization.OpCodes.OpCode
 import special.collection.{Coll, CollType}
@@ -30,7 +30,7 @@ class ErgoLikeContext(val currentHeight: Height,
                       val spendingTransaction: ErgoLikeTransactionTemplate[_ <: UnsignedInput],
                       val self: ErgoBox,
                       override val extension: ContextExtension = ContextExtension(Map())
-                 ) extends Context {
+                 ) extends ErgoContext {
   assert(self == null || boxesToSpend.exists(box => box.id == self.id), s"Self box if defined should be among boxesToSpend")
   override def withExtension(newExtension: ContextExtension): ErgoLikeContext =
     ErgoLikeContext(currentHeight, lastBlockUtxoRoot, minerPubkey, boxesToSpend, spendingTransaction, self, newExtension)
--- a/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
+++ b/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
@@ -163,3 +163,9 @@ case object Self extends NotReadyValueBox {
   override val opCode: OpCode = OpCodes.SelfCode
   def opType = SFunc(SContext, SBox)
 }
+
+case object Context extends NotReadyValue[SContext.type] {
+  override val opCode: OpCode = OpCodes.ContextCode
+  override def tpe: SContext.type = SContext
+  override def opType: SFunc = SFunc(SUnit, SContext)
+}
--- a/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
+++ b/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
@@ -164,8 +164,3 @@ case object Self extends NotReadyValueBox {
   def opType = SFunc(SContext, SBox)
 }
 
-case object Context extends NotReadyValue[SContext.type] {
-  override val opCode: OpCode = OpCodes.ContextCode
-  override def tpe: SContext.type = SContext
-  override def opType: SFunc = SFunc(SUnit, SContext)
-}
--- a/src/main/scala/sigmastate/eval/Evaluation.scala
+++ b/src/main/scala/sigmastate/eval/Evaluation.scala
@@ -259,7 +259,17 @@ trait Evaluation extends RuntimeCosting { IR =>
 
           // NOTE: This is a fallback rule which should be places AFTER all other MethodCall patterns
           case mc @ MethodCall(obj, m, args, _) =>
-            val dataRes = invokeUnlifted(obj.elem, mc, dataEnv)
+            val dataRes = obj.elem match {
+              case _: CollElem[_, _] => mc match {
+                case CollMethods.flatMap(xs, f) =>
+                  val newMC = mc.copy(args = mc.args :+ f.elem.eRange)(mc.selfType, mc.isAdapterCall)
+                  invokeUnlifted(obj.elem, newMC, dataEnv)
+                case _ =>
+                  invokeUnlifted(obj.elem, mc, dataEnv)
+              }
+              case _ =>
+                invokeUnlifted(obj.elem, mc, dataEnv)
+            }
             val res = dataRes match {
               case Constant(v, _) => v
               case v => v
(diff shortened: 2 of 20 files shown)


src/main/scala/org/ergoplatform/ErgoLikeContext.scala
```diff
@@ -162,3 +162,9 @@ case object Self extends NotReadyValueBox {
   override val opCode: OpCode = OpCodes.SelfCode
   def opType = SFunc(SContext, SBox)
 }
+
+case object Context extends NotReadyValue[SContext.type] {
+  override val opCode: OpCode = OpCodes.ContextCode
+  override def tpe: SContext.type = SContext
+  override def opType: SFunc = SFunc(SUnit, SContext)
+}
```
aslesarenko: @greenhat We don't need such node in the tree, because any script if `f: Context => SigmaProp`, so to context is actually argument of the function.

greenhat: @aslesarenko If we remove `Context` case object we need to translate every context method call into an IR node in the binder. This means no `MethodCall`s. Or am I missing something?

aslesarenko: @greenhat We will still keep existing context operations as shortcuts for method of Context.
But we don't need new shortcuts.
And remember is the graph IR, `HEIGHT` becomes `ctx.HEIGHT` where ctx is lambda var.

greenhat: @aslesarenko Got it. Removed.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -1501,6 +1501,31 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
...
 
+      case Terms.MethodCall(obj, method, args) if obj.tpe.isCollectionLike =>
+        val xsC = asRep[CostedColl[Any]](evalNode(ctx, env, obj))
+        val argsCostVals = args.map {
+          case sfunc: Value[SFunc]@unchecked if sfunc.tpe.isFunc =>
+            val funC = asRep[CostedFunc[Unit, Any, Any]](evalNode(ctx, env, sfunc)).func
+            val (calcF, costF) = splitCostedFunc2(funC, okRemoveIsValid = true)
+            val cost = xsC.values.zip(xsC.costs.zip(xsC.sizes)).map(costF).sum(intPlusMonoid)
+            (cost, calcF)
+          case a@_ =>
```
aslesarenko: @greenhat maybe `case a => `?

src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -1501,6 +1501,31 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
...
+            val aC = eval(a)
+            (aC.cost, aC.value)
+        }
+        val argsCosts = argsCostVals.map(_._1)
```
aslesarenko: Try using argsCostVals.unzip

src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -1501,6 +1501,31 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
...
+        val xsC = asRep[CostedColl[Any]](evalNode(ctx, env, obj))
+        val argsCostVals = args.map {
+          case sfunc: Value[SFunc]@unchecked if sfunc.tpe.isFunc =>
+            val funC = asRep[CostedFunc[Unit, Any, Any]](evalNode(ctx, env, sfunc)).func
+            val (calcF, costF) = splitCostedFunc2(funC, okRemoveIsValid = true)
+            val cost = xsC.values.zip(xsC.costs.zip(xsC.sizes)).map(costF).sum(intPlusMonoid)
+            (cost, calcF)
+          case a@_ =>
+            val aC = eval(a)
+            (aC.cost, aC.value)
```
aslesarenko: Let's keep as a convention the order (value, cost, size)

src/test/scala/sigmastate/lang/SigmaCompilerTest.scala
```diff
@@ -246,24 +246,20 @@ class SigmaCompilerTest extends SigmaTestingCommons with LangTests with ValueGen
...
         ConcreteCollection(TrueLeaf, FalseLeaf),
         SCollection.IndicesMethod,
         Vector()
       )
-    )
   }
 
   property("SCollection.flatMap") {
-    testMissingCosting("OUTPUTS.flatMap({ (out: Box) => Coll(out.value >= 1L) })",
+    comp("OUTPUTS.flatMap({ (out: Box) => Coll(out.value >= 1L) })") shouldBe
```
aslesarenko: @greenhat The flatMap can accept only limited lambdas as arguments. 
Currently:
1) `x => x._1`
2) `box => MethodCall(box, name, List())`
3) `x: Coll[T] => x`

aslesarenko: please put this requirements into description of method call both LangSpec.md and special.collection.Colls.scala

src/main/scala/sigmastate/eval/RuntimeCosting.scala
greenhat: Fixed.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
greenhat: Thanks! Done.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
greenhat: Fixed.

src/test/scala/sigmastate/lang/SigmaCompilerTest.scala
greenhat: @aslesarenko Done. What's the reason for this limitation? I'm struggling to find where and how it's enforced.

src/main/scala/sigmastate/eval/Evaluation.scala
```diff
@@ -261,7 +261,17 @@ trait Evaluation extends RuntimeCosting { IR =>
 
           // NOTE: This is a fallback rule which should be places AFTER all other MethodCall patterns
           case mc @ MethodCall(obj, m, args, _) =>
-            val dataRes = invokeUnlifted(obj.elem, mc, dataEnv)
+            val dataRes = obj.elem match {
+              case _: CollElem[_, _] => mc match {
+                case CollMethods.flatMap(xs, f) =>
```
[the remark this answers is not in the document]
aslesarenko: Yes, please, and also for other methods with RType implicit argument (if any)

PR #369, opened 2019-01-22
New serialization

VictorMikheev:


src/main/scala/sigmastate/serialization/SigmaSerializer.scala
```diff
@@ -34,7 +22,7 @@ object Serializer {
   def startReader(bytes: Array[Byte], pos: Int = 0): SigmaByteReader = {
     val buf = ByteBuffer.wrap(bytes)
     buf.position(pos)
-    val r = new SigmaByteReader(buf, new ConstantStore(), resolvePlaceholdersToConstants = false)
+    val r = new SigmaByteReader(new VLQByteBufferReader(buf), new ConstantStore(), resolvePlaceholdersToConstants = false)
```
aslesarenko: @VictorMikheev I don't think this is a good idea to have VLQ/Non-VLQ logic reader-wise. The same reader instance should have BOTH. Thus this name is misleading and should be changed. Because it is in dependency please change it first and then update this PR.

Issue #374, opened 2019-01-25
A contract is causing an unexpected exception during deserialization 


kushti:
The following contract https://gist.github.com/kushti/1e9f6d0add6a5fd926743aaa26d448bd is causing the following exception https://gist.github.com/kushti/6e5afdb21a5c4d20bc49bca8b9eaf23f on line 102.

greenhat:
Fixed in #375. Confirmed green on current `master`.

PR #375, opened 2019-01-25
Recognize non-const ProveDlog and ProveDHTuple

### LINKED ISSUE #373: Value incorrectly recognized as a constant breaks (de)serialization
Code to reproduce:
```
    import sigmastate.interpreter.CryptoConstants.dlogGroup
    val bob = new ErgoLikeTestProvingInterpreter
    val env = Map("gA" -> dlogGroup.generator, "yA" -> bob.dlogSecrets.head.publicImage.value)
    val script = compileWithCosting(env,
      """{
        |    val c1 = OUTPUTS(0).R4[GroupElement].get
        |    val c2 = OUTPUTS(0).R5[GroupElement].get
        |    proveDHTuple(gA, c1, yA, c2)
        |}""".stripMargin
    ).asBoolValue // script successfully compiles
    val box = ErgoBox(10, script, 100) // this line gives error
```

greenhat:
Close #373 
Ref #374

### DIFF
--- a/src/main/scala/sigmastate/eval/TreeBuilding.scala
+++ b/src/main/scala/sigmastate/eval/TreeBuilding.scala
@@ -296,9 +296,20 @@ trait TreeBuilding extends RuntimeCosting { IR: Evaluation =>
       case Def(TrivialSigmaCtor(In(cond))) =>
         mkBoolToSigmaProp(cond.asBoolValue)
       case Def(ProveDlogEvidenceCtor(In(g))) =>
-        SigmaPropConstant(mkProveDlog(g.asGroupElement))
+        g match {
+          case gc: Constant[SGroupElement.type]@unchecked => SigmaPropConstant(mkProveDlog(gc))
+          case _ => mkProveDlog(g.asGroupElement)
+        }
       case Def(ProveDHTEvidenceCtor(In(g), In(h), In(u), In(v))) =>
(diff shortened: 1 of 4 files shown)


src/main/scala/sigmastate/eval/TreeBuilding.scala
```diff
@@ -296,9 +296,20 @@ trait TreeBuilding extends RuntimeCosting { IR: Evaluation =>
...
+          case gc: Constant[SGroupElement.type]@unchecked => SigmaPropConstant(mkProveDlog(gc))
+          case _ => mkProveDlog(g.asGroupElement)
+        }
       case Def(ProveDHTEvidenceCtor(In(g), In(h), In(u), In(v))) =>
-        SigmaPropConstant(mkProveDiffieHellmanTuple(g.asGroupElement, h.asGroupElement, u.asGroupElement, v.asGroupElement))
+        (g, h, u, v) match {
+          case (gc: Constant[SGroupElement.type]@unchecked,
+          hc: Constant[SGroupElement.type]@unchecked,
+          uc: Constant[SGroupElement.type]@unchecked,
+          vc: Constant[SGroupElement.type]@unchecked) =>
```
aslesarenko: What if one of the matched values is not Constant?

greenhat: @aslesarenko `ProveDHTuple` will not be wrapped in `ConstantNode`. See `case _`.

PR #380, opened 2019-01-29
Protocol examples 

scalahub:
Added Mixing example spec.

### DIFF
--- a/docs/sigmastate_protocols/sigmastate_protocols.bib
+++ b/docs/sigmastate_protocols/sigmastate_protocols.bib
@@ -1,3 +1,18 @@
+@InProceedings{fiatshamir,
+author="Fiat, Amos
+and Shamir, Adi",
+editor="Odlyzko, Andrew M.",
+title="How To Prove Yourself: Practical Solutions to Identification and Signature Problems",
+booktitle="Advances in Cryptology --- CRYPTO' 86",
+year="1987",
+publisher="Springer Berlin Heidelberg",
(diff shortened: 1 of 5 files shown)


src/test/scala/sigmastate/utxo/examples/DHTupleExampleSpecification.scala
```diff
@@ -0,0 +1,115 @@
...
+    val inBox = ErgoBox(10, script, 50)
+
+    // a blockchain node verifying a block containing a spending transaction
+    val verifier = new ErgoLikeTestInterpreter
+
+    val bob = new ErgoLikeTestProvingInterpreter
+
+    val y:BigInteger = bob.dlogSecrets.head.w // y is Bob's private key
+
+    val g_y = GroupElementConstant(dlogGroup.exponentiate(g, y)) // g^y
```
greenhat: @scalahub Is there a reason to get Bob's public key this way? I find the following code easier to grasp and more in line with Alice's key's definitions:
```suggestion
   val bobPubKey: ProveDlog = bob.dlogSecrets.head.publicImage
   val g_y = GroupElementConstant(bobPubKey.h) // Bob's public key g^y
```
Good job explaining how DHT works, BTW!

scalahub: No particular reason. Just wanted to show how to generate a public key from a private value y. This version also fine.

scalahub:
Already done in #441

PR #383, opened 2019-02-01
Code from aslesarenko/sigma added

aslesarenko:
This is done to simplify dependencies, development and deployment.
Hopefully this doesn't reduce modularity.
Dependency on aslesarenko/sigma is removed.


build.sbt
```diff
@@ -59,14 +64,22 @@ version in ThisBuild := {
 
 git.gitUncommittedChanges in ThisBuild := true
 
+val bouncycastleBcprov = "org.bouncycastle" % "bcprov-jdk15on" % "1.60"
+val scripto            = "org.scorexfoundation" %% "scrypto" % "2.1.4"
```
kushti: scrypto

PR #388, opened 2019-02-06
Fix publishing artifacts in sigma* subprojects;

greenhat:
+ fix assembly of scalanizer;


greenhat:
It’s already  in a 2.0 branch.

Issue #393, opened 2019-02-11
Move SigmaBoolean out of Value hierarchy 


aslesarenko:
This is critical one, and probably lead to serialization format changes.
This is a source of many problems because SigmaBooleans cannot be mixed with Values.
SigmaBoolean is data (computed during script execution) where as Value is node of ErgoTree IR.
Every time sigma boolean should be part of the ErgoTree it should be wrapped in SigmaPropConstant.
However, in current version, the compiler cannot raise an error when SigmaBoolean value appears in the position where Value is expected, thus necessary wrapping may be missed which later lead to bugs, which very hard to localize.

aslesarenko:
Resolved in v2.0

PR #399, opened 2019-02-19
New SigmaBoolean not inherited from Value[SBoolean.type]

### LINKED ISSUE #376: Add CreateProveDlog and CreateProveDHTuple primitive functions
It should resolve #374 bug.
We need to separate ProveDlog value (which is SigmaBoolean, which is node of sigma tree) from CreateProveDlog: GroupElement => SigmaProp primitive of ErgoTree, which is primitive function taking GroupElement and producing value of SigmaProp type.

aslesarenko:
Close #376, #393


src/main/scala/sigmastate/Values.scala
```diff
@@ -522,24 +533,72 @@ object Values {
...
 
   object SigmaBoolean {
     val PropBytes = "propBytes"
-    val IsProven = "isProven"
-    val fields = Seq(
-      PropBytes -> SByteArray,
-      IsProven -> SBoolean
-    )
+    val IsValid = "isValid"
+    object serializer extends SigmaSerializer[SigmaBoolean, SigmaBoolean] {
```
greenhat: Why not put the serializer into a separate file?

aslesarenko: It just continues the design used in ErgoBox, AvlTreeData etc. I agreed it should be changed though.

PR #400, opened 2019-02-20
Implement missing ops and predefined functions for 2.0

greenhat:
#324 

Todo:
- [x] add/fix (de)serialization;
- [ ] `ZKProofBlock` (de)serialization?
- [x] `byteArrayToLong`;
- [x] `xorOf`;
- [x] `Box.getReg`;
- [x] `!`(logical Not);
- [x] `-`(negation);
- [ ] `^` (logical XOR);
- [ ] `~` (bitwise inversion);
- [ ] `|` (bitwise OR for Numeric);
- [ ] `&` (bitwise AND for Numeric);
- [ ] `^` (bitwise XOR for Numeric);
- [ ] `<<`, `>>`, `>>>` (bit shifts for Numeric);
- [ ] `outerJoin`;
- [ ] `executeFromVar`;
- [ ] `substConstants`;


src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -1415,6 +1419,20 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
           withDefaultSize(res, cost)
       }
 
+      case XorOf(input) => input match {
+        case ConcreteCollection(items, tpe) =>
+          val itemsC = items.map(item => eval(adaptSigmaBoolean(item)))
```
aslesarenko: @greenhat why `adaptSigmaBoolean` is necessary. BoolValue is not related to SigmaBoolean anymore.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -1520,6 +1554,13 @@ trait RuntimeCosting extends SigmaLibrary with DataCosting with Slicing { IR: Ev
         val cost = inputC.cost + costOf(node)
         withDefaultSize(res, cost)
 
+      case ByteArrayToLong(In(arr)) =>
+        val arrC = asRep[Costed[Coll[Byte]]](arr)
+        val value = sigmaDslBuilder.byteArrayToLong(arrC.value)
+        val size = arrC.dataSize
```
aslesarenko: Since we know the size of resulting Long is constant, it is better to use withDefaultSize method here, this makes the costing graph simpler and the whole process faster.

src/main/scala/sigmastate/serialization/MethodCallSerializer.scala
```diff
@@ -16,6 +17,8 @@ case class MethodCallSerializer(opCode: Byte, cons: (Value[SType], SMethod, Inde
       assert(mc.args.nonEmpty)
       w.putValues(mc.args)
     }
+    w.putUByte(mc.typeSubst.size)
+    mc.typeSubst.foreach { case (typeIdent, tpe) => w.putType(typeIdent).putType(tpe) }
```
aslesarenko: We don't need to save type idents because they can be restored from Method.stype instance, where we already have a Seq of them. We just need to serialize tpes in the same order.

src/main/scala/sigmastate/types.scala
```diff
@@ -110,7 +111,7 @@ object SType {
     * NOTE: in the current implementation only monomorphic methods are supported (without type parameters)*/
   val types: Map[Byte, STypeCompanion] = Seq(
     SNumericType, SString, STuple, SGroupElement, SSigmaProp, SContext,
-    SAvlTree, SBox, SOption, SCollection
+    SAvlTree, SBox, SOption, SCollection, SBigInt, SOption
```
aslesarenko: Why SOption is two times in this list?

src/test/scala/sigmastate/helpers/SigmaTestingCommons.scala
```diff
@@ -79,16 +88,19 @@ trait SigmaTestingCommons extends PropSpec
     val tB = RType[B]
     val tpeA = Evaluation.rtypeToSType(tA)
     val tpeB = Evaluation.rtypeToSType(tB)
+    val getVarType = if (tpeA == SBigInt) "BigInt" else tA.name
```
aslesarenko: this hack it not necessary, tA.name should return correct type name. See `implicit val BigIntRType: RType[BigInt] = wrapperType[BigInt]`

src/test/scala/special/sigma/SigmaDslTest.scala
```diff
@@ -111,6 +111,49 @@ class SigmaDslTest extends PropSpec with PropertyChecks with Matchers with Sigma
     }
   }
 
+  property("byteArrayToLong equivalence") {
+    val eq = checkEq(func[Coll[Byte],Long]("{ (x: Coll[Byte]) => byteArrayToLong(x) }")){ x =>
+      Longs.fromByteArray(x.toArray)
```
aslesarenko: The Scala lambda should be exactly the same as Script lambda, because it should be SigmaDsl expression, which is the same as SigmaScript expression.
Here we actually test equivalence of SigmaDsl and SigmaScript.

src/test/scala/special/sigma/SigmaDslTest.scala
```diff
@@ -111,6 +111,49 @@ class SigmaDslTest extends PropSpec with PropertyChecks with Matchers with Sigma
...
+    forAll { x: Array[Byte] =>
+      whenever(x.size >= 8) {
+        eq(Builder.DefaultCollBuilder.fromArray(x))
+      }
+    }
+  }
 
+  property("xorOf equivalence") {
+    val eq = checkEq(func[Coll[Boolean], Boolean]("{ (x: Coll[Boolean]) => xorOf(x) }")) { x =>
+      x.toArray.distinct.length == 2
```
aslesarenko: the same as above.

src/main/scala/sigmastate/types.scala
greenhat: Merging artifacts I suppose. Thanks for spotting this one! Fixed.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
greenhat: Thanks. Done.

src/main/scala/sigmastate/eval/RuntimeCosting.scala
greenhat: Fixed.

src/main/scala/sigmastate/serialization/MethodCallSerializer.scala
greenhat: @aslesarenko Interesting idea. I have two questions: 
1. We have Seq of type params `SFunc`. What about `Option.get` which `stype` is `STypeIdent`?
2. Would not it be too fragile, when simple type params reordering in method declaration would break deserialization of existing data?

aslesarenko: 1) We have `SGenericType` and by mistake SOption is not inherited from it.
2) I don't think we need to support changing order paying such a big price at blockchain scale :)

src/test/scala/sigmastate/helpers/SigmaTestingCommons.scala
greenhat: I've switched to `special.sigma.BigInt` and after a couple of fixes [commit](
https://github.com/ScorexFoundation/sigmastate-interpreter/pull/400/commits/55eb2d198f02653908efd1485c65d9d93d370299) I ended up with a path type which drives our parser crazy:
```
line 3:   val res = func(getVar[special.sigma.BigInt](1).get)
                                              ^
Path types are not supported: Select(Select(Ident(special,NoType),sigma,None),BigInt,None)
```
For `scala.BigInt` case I stuck with missing ev for RType. I'll get back to that later.

src/test/scala/special/sigma/SigmaDslTest.scala
greenhat: Whoa! How can I miss this? Fixed (mostly) in [commit](
https://github.com/ScorexFoundation/sigmastate-interpreter/pull/400/commits/237de80470a2c35de96e02563a486a8c01fb9f1f) 

Now I clearly get why we need prefix methods. :)

src/test/scala/special/sigma/SigmaDslTest.scala
greenhat: Fixed.

src/test/scala/sigmastate/helpers/SigmaTestingCommons.scala
aslesarenko: Yes, scala.BigInt is not and will not be supported. There is BigIntegerRType and BigIntRType implicits is special.sigma package object. Note, that BigInt is parsed string is not the same type as special.sigma.BigInt. They correspond, but not the same things.

src/main/scala/sigmastate/serialization/MethodCallSerializer.scala
greenhat: Done in 
https://github.com/ScorexFoundation/sigmastate-interpreter/pull/400/commits/391cec0bc40fedaf2229ec4e5dafc52093f7f9c7

src/main/scala/sigmastate/serialization/MethodCallSerializer.scala
```diff
@@ -17,25 +17,38 @@ case class MethodCallSerializer(opCode: Byte, cons: (Value[SType], SMethod, Inde
...
       w.putValues(mc.args)
     }
-    w.putUByte(mc.typeSubst.size)
-    mc.typeSubst.foreach { case (typeIdent, tpe) => w.putType(typeIdent).putType(tpe) }
+    mc.method.stype match {
+      case genType: SGenericType =>
+        w.putUByte(mc.typeSubst.size)
+        mc.typeSubst.foreach { case (ti, tpe) =>
+          val tpIndex = genType.substitutedTypeParams.indexOf(STypeParam(ti))
+          w.putUByte(tpIndex)
```
aslesarenko: @greenhat we don't need to output tpIndex either. We just need to always serialize them is the order specified in SGenericType.typeParams

greenhat: @aslesarenko Damn, that's so obvious in the hindsight! Thanks. This also catches missed type substitutions. Done in 
https://github.com/ScorexFoundation/sigmastate-interpreter/pull/400/commits/eb4b3c177c7db1583329e177c030d971987af40f

src/test/scala/sigmastate/helpers/SigmaTestingCommons.scala
greenhat: @aslesarenko Ok, I'm a bit confused here. Do we have to support both `special.sigma.BigInt` and `java.math.BigInteger` in SigmaDsl? If so, does it mean that both of them in the generated script should have `BigInt` sigma type in `getVar[BigInt](1)`? 
`tA.name` for both of them renders a path type which is not supported in our parser. Do you see another way of fixing instead of my original hack above? I don't like changing `SGeneralType.name` to not return a path type.

aslesarenko: In SigmaDsl we only use special.sigma.BigInt, which is WrapperOf[BigInteger] (see WrapperOf)
BigInterger we use internally.

You can add this method.
```
  case class WrapperType[Wrapper](cWrapper: ClassTag[Wrapper]) extends RType[Wrapper] {
   ...
    override def name: String = cWrapper.runtimeClass.getSimpleName
  }
```
Note, in `def func[A: RType, B: RType](func: String)(implicit IR: IRContext): A => B` we can assume A and B both SigmaDsl types.
See  this function for details of data conversion
```
/** Generic converter from types used in ErgoTree values to types used in ErgoDsl. */
  def toDslData(value: Any, tpe: SType, isCost: Boolean)(implicit IR: Evaluation): Any
```

aslesarenko: Separation between SigmaDsl types and ErgoTree types is annoying, but maybe in future we can eliminate this distinction. Currently we need to do conversions. Search for places where conversion methods are used.

greenhat: Done in 
https://github.com/ScorexFoundation/sigmastate-interpreter/pull/400/commits/2a5524a9b61e01af4bab4613002ef883057da40f

PR #435, opened 2019-03-14
Protocol examples v2.0

scalahub:
Enhanced documentation of some protocols and added an example of using the oracle with tokens.


scalahub:
Created new alternate PR (441) with branch protocol_examples_new based on master.

PR #442, opened 2019-03-18
Costing tests

catena2w:

PR #440, opened 2019-03-15
Global groupGenerator method

aslesarenko:
closes #426

PR #446, opened 2019-03-22
Costing tests (part 1)

aslesarenko:
Replaces #442


src/main/scala/org/ergoplatform/ErgoBox.scala
```diff
@@ -44,26 +47,24 @@ import scala.runtime.ScalaRunTime
...
+         val index: Short,
+         override val creationHeight: Int
+       ) extends ErgoBoxCandidate(value, ergoTree, creationHeight, additionalTokens, additionalRegisters) {
 
   import ErgoBox._
 
-  lazy val id: BoxId = ADKey @@ Blake2b256.hash(bytes)
-
-  override def dataSize: Long = bytes.length
+  lazy val idBytes: BoxId = ADKey @@ Blake2b256.hash(bytes)
```
catena2w: I think type `BoxId` should also be renamed (or just deleted) in such a case.
Also I propose to name this fields as `adKey`, at least it would be consistent with the type

src/main/scala/org/ergoplatform/ErgoBox.scala
```diff
@@ -93,7 +94,7 @@ object ErgoBox {
     val size: Short = 32
   }
 
-  type TokenId = Digest32
```
kushti: Digest32 type was introduced for a reason introduce distinction between atbitrary-sized byte arrays and 32-bytes arrays. Not clear why new type is introduced.

src/test/scala/sigmastate/CostingSpecification.scala
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  property("SELF box operations cost") {
+    cost("{ SELF.value > 0 }") shouldBe (accessBox + extractCost + GTConstCost)
+    cost("{ SELF.id.size > 0 }") shouldBe (accessBox + extractCost + collLength + GTConstCost)
+    cost("{ SELF.tokens.size > 0 }") shouldBe (accessBox + extractCost + collLength + GTConstCost)
+    cost("{ SELF.creationInfo._1 > 0 }") shouldBe (accessBox + accessRegister + selectField + GTConstCost)
+    cost("{ SELF.R5[Int].get > 0 }") shouldBe (accessBox + RegisterAccess + GTConstCost)
+
+//    cost("{ SELF.getReg[Long](0.toByte).get > 0 }") shouldBe (accessBox + RegisterAccess + GTConstCost)
```
kushti: commented code without explanation

src/test/scala/sigmastate/lang/SigmaTyperTest.scala
```diff
@@ -275,6 +275,7 @@ class SigmaTyperTest extends PropSpec with PropertyChecks with Matchers with Lan
     typecheck(env, "SELF.R1[Int].isDefined") shouldBe SBoolean
     typecheck(env, "SELF.R1[Int].isEmpty") shouldBe SBoolean
     typecheck(env, "SELF.R1[Int].get") shouldBe SInt
+//    typecheck(env, "SELF.getReg[Int](1)") shouldBe SOption.SIntOption
```
kushti: commented code without explanation

src/test/scala/sigmastate/eval/BasicOpsTests.scala
```diff
@@ -60,16 +60,11 @@ class BasicOpsTests extends FunSuite with ContractsTestkit with Matchers {
...
-    val selfId = collection[Byte](0, 1)
-    val self = new TestBox(selfId, 10, noBytes, noBytes, noBytes, noRegisters)
-    val ctx = testContext(noInputs, noOutputs, height = 200, self, emptyAvlTree, dummyPubkey, Array())
-  }
 
-  test("box.creationInfo._1 is Int") {
-    val box = newAliceBox(1, 100, Map(3 -> toAnyValue((20 -> SigmaDsl.Colls.fromArray(Array.emptyByteArray)))))
-    box.creationInfo._1 shouldBe a [Integer]
-  }
+//  test("box.creationInfo._1 is Int") {
```
kushti: use ignore, not commenting out the code. Please provide an explanation why the test is ignored currently

src/main/scala/sigmastate/utils/Helpers.scala
```diff
@@ -46,6 +46,15 @@ object Helpers {
     result
   }
 
+  def concatArrays[T:ClassTag](arr1: Array[T], arr2: Array[T]): Array[T] = {
+    val length: Int = arr1.length + arr2.length
+    val result: Array[T] = new Array[T](length)
+    var pos: Int = 0
```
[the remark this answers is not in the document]
aslesarenko: They are used in later, I don't know how they became part of this PR, but this is necessary.

src/test/scala/sigmastate/CostingSpecification.scala
aslesarenko: made it `// TODO   cost("{ SELF.getReg[Long](...` in upcoming PR

src/test/scala/sigmastate/lang/SigmaTyperTest.scala
aslesarenko: made it `// TODO    typecheck(env, "SELF.getReg[Int](1)")` in upcoming PR

src/test/scala/sigmastate/eval/BasicOpsTests.scala
aslesarenko: made it working in upcoming PR

src/main/scala/sigmastate/utils/Helpers.scala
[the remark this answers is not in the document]
aslesarenko: val MaxOpCode : OpCode = XorOfCode
is already removed in #440

src/main/scala/org/ergoplatform/ErgoBox.scala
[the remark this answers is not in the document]
aslesarenko: This is to optimize  `org.ergoplatform.ErgoLikeContext.toSigmaContext` which takes about 32% of times mentioned here https://github.com/ergoplatform/ergo/issues/645

src/main/scala/org/ergoplatform/ErgoBox.scala
aslesarenko: 1) This is to optimize org.ergoplatform.ErgoLikeContext.toSigmaContext which takes about 32% of times mentioned here ergoplatform/ergo#645
2) Digest32 is not true new type, it is still mutable array without correct equals and hashcode. With Coll it will be more safe and also more efficient by implementing special hashCode with will use first 4 bytes from the digest

PR #459, opened 2019-03-30
Add type check for deserialized script

### LINKED ISSUE #454: Type check on deserialization
Add type checking when deserializing script from bytes

greenhat:
Close #454 

Todo:
- [x] type check in `deserializeErgoTree`;
- [x] narrow `serial*/deserial*` methods to `SigmaPropValue` and add a type check on deserialization; 
- [x] test coverage of `ErgoTreeSerializer`;
- [x] remove any ser/deser except ErgoTree in ErgoTreeSerializer;
- [x] fix ergo (`SigmaProp` only ser/deser, `Value.bytes` removed);
- [x] for type mismatch in `DeserializeContext`, ensure that incorrect scripts will be declined by our verifier if the transaction was formed using different prover;
- [x] test for constant type mismatch case on deserialization (Not applicable. For a constant placeholder we store only an index for a constant and not the constant type. When deserializing constant placeholder we get the type from the constant store, i.e from deserialized segregated constant. It means that constant placeholder type follows the segregated constant type.);
- [x] check types underneath root node of the script? (extracted into #473)

ergo PR https://github.com/ergoplatform/ergo/pull/681

### DIFF
--- a/src/main/scala/org/ergoplatform/ErgoScriptPredef.scala
+++ b/src/main/scala/org/ergoplatform/ErgoScriptPredef.scala
@@ -37,7 +37,7 @@ object ErgoScriptPredef {
   def expectedMinerOutScriptBytesVal(delta: Int, minerPkBytesVal: Value[SByteArray]): Value[SByteArray] = {
     val genericPk = ProveDlog(CryptoConstants.dlogGroup.generator)
     val genericMinerProp = rewardOutputScript(delta, genericPk)
-    val genericMinerPropBytes = DefaultSerializer.serializeErgoTree(genericMinerProp.treeWithSegregation)
+    val genericMinerPropBytes = DefaultSerializer.serializeErgoTree(genericMinerProp)
     // first segregated constant is delta, so key is second constant
     val positions = IntArrayConstant(Array[Int](1))
     val minerPubkeySigmaProp = CreateProveDlog(DecodePoint(minerPkBytesVal))
@@ -48,30 +48,30 @@ object ErgoScriptPredef {
   /**
     * Required script of the box, that collects mining rewards
     */
-  def rewardOutputScript(delta: Int, minerPk: ProveDlog): SigmaPropValue = {
+  def rewardOutputScript(delta: Int, minerPk: ProveDlog): ErgoTree = {
     SigmaAnd(
       GE(Height, Plus(boxCreationHeight(Self), IntConstant(delta))).toSigmaProp,
       SigmaPropConstant(minerPk)
-    )
+    ).treeWithSegregation
   }
 
   /**
     * Proposition that allows to send coins to a box which is protected by the following proposition:
     * prove dlog of miner's public key and height is at least `delta` blocks bigger then the current one.
     */
-  def feeProposition(delta: Int = 720): SigmaPropValue = {
+  def feeProposition(delta: Int = 720): ErgoTree = {
     val out = ByIndex(Outputs, IntConstant(0))
     AND(
       EQ(Height, boxCreationHeight(out)),
       EQ(ExtractScriptBytes(out), expectedMinerOutScriptBytesVal(delta, MinerPubkey)),
       EQ(SizeOf(Outputs), 1)
-    ).toSigmaProp
+    ).toSigmaProp.treeWithSegregation
   }
 
   /**
     * A contract that only allows to collect emission reward by a box with miner proposition.
     */
-  def emissionBoxProp(s: MonetarySettings): SigmaPropValue = {
+  def emissionBoxProp(s: MonetarySettings): ErgoTree = {
     val rewardOut = ByIndex(Outputs, IntConstant(0))
     val minerOut = ByIndex(Outputs, IntConstant(1))
 
@@ -97,7 +97,7 @@ object ErgoScriptPredef {
       heightIncreased,
       correctMinerOutput,
       OR(AND(outputsNum, sameScriptRule, correctCoinsConsumed, heightCorrect), lastCoins)
-    ).toSigmaProp
+    ).toSigmaProp.treeWithSegregation
   }
 
   /**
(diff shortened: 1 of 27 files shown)


src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala
```diff
@@ -111,10 +115,10 @@ class ErgoTreeSerializer {
     r.constantStore = new ConstantStore(constants)
     val previousResolvePlaceholderValue = r.resolvePlaceholdersToConstants
     r.resolvePlaceholdersToConstants = resolvePlaceholdersToConstants
-    val tree = ValueSerializer.deserialize(r)
+    val value = ValueSerializer.deserialize(r)
```
greenhat: @aslesarenko I did not add a type check (is `SigmaProp`) here since the `serializeWithSegregation` us used for `Value.bytes` in [src](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/ac5a03c152e66162a48ebe4a080479565c42b92a/src/main/scala/sigmastate/Values.scala#L58).

aslesarenko: @greenhat do we need this Value.bytes at all?
I suggest to remove it completely. 
It is used mostly in Specifications and should be replaced by ErgoTree serialization.
The only core usage is inside MonetarySettings where it also should be replaces with ErgoTree serialization. /cc @catena2w @kushti

greenhat: I've removed `Value.bytes` and switch ser/deser methods to `SigmaPropValue` with added type check on deserialization. See [commit](https://github.com/ScorexFoundation/sigmastate-interpreter/pull/459/commits/f72682893e1e2676a6befcc223b28948ba920cce).

aslesarenko: @greenhat I really mean completely :) 
We don't need any public means to serialize Value directly. 
The only script serialization should be via explicit construction of ErgoTree. 
Otherwise it leads to confusion, how the bytes should be deserialized, and how to interpret result of such deserialization.

greenhat: @aslesarenko Done in https://github.com/ScorexFoundation/sigmastate-interpreter/pull/459/commits/6edd1f4089a2df3ab8b4b4fc25dc8e7cdffa462e

src/main/scala/org/ergoplatform/ErgoScriptPredef.scala
```diff
@@ -37,7 +37,7 @@ object ErgoScriptPredef {
   def expectedMinerOutScriptBytesVal(delta: Int, minerPkBytesVal: Value[SByteArray]): Value[SByteArray] = {
     val genericPk = ProveDlog(CryptoConstants.dlogGroup.generator)
     val genericMinerProp = rewardOutputScript(delta, genericPk)
-    val genericMinerPropBytes = ErgoTreeSerializer.DefaultSerializer.serializeWithSegregation(genericMinerProp)
+    val genericMinerPropBytes = DefaultSerializer.serializeErgoTree(genericMinerProp.treeWithSegregation)
```
aslesarenko: @greenhat I suggest rewardOutputScript should return ErgoTree. So the decision of whether to segregate constants or not and to which extent is done inside the method.

greenhat: Done.

src/main/scala/org/ergoplatform/settings/MonetarySettings.scala
```diff
@@ -18,7 +18,7 @@ case class MonetarySettings(fixedRatePeriod: Int = 30 * 2 * 24 * 365,
                             foundersInitialReward: Long = 75L * EmissionRules.CoinsInOneErgo / 10) {
 
   val feeProposition: SigmaPropValue = ErgoScriptPredef.feeProposition(minerRewardDelay)
-  val feePropositionBytes: Array[Byte] = feeProposition.bytes
+  val feePropositionBytes: Array[Byte] = feeProposition.treeWithSegregation.bytes
```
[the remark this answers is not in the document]
greenhat: Fixed.

src/test/scala/sigmastate/helpers/SigmaTestingCommons.scala
```diff
@@ -49,9 +49,9 @@ trait SigmaTestingCommons extends PropSpec
   val compiler = SigmaCompiler(TestnetNetworkPrefix, TransformingSigmaBuilder)
 
   def checkSerializationRoundTrip(v: SValue): Unit = {
-    val compiledTreeBytes = ErgoTreeSerializer.DefaultSerializer.serializeWithSegregation(v)
+    val compiledTreeBytes = ValueSerializer.serialize(v)
```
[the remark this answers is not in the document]
greenhat: This round trip check is used extensively in our tests for all type of values, not only SigmaProp.

src/test/scala/sigmastate/serialization/ErgoTreeSerializerSpecification.scala
```diff
@@ -50,64 +49,58 @@ class ErgoTreeSerializerSpecification extends SerializationSpecification with Si
...
-      val ergoTree = extractConstants(processedTree)
-      val bytes = ErgoTreeSerializer.DefaultSerializer.serializeErgoTree(ergoTree)
-      val (_, deserializedConstants, treeBytes) = ErgoTreeSerializer.DefaultSerializer
-        .deserializeHeaderWithTreeBytes(SigmaSerializer.startReader(bytes))
-      val c = new ConstantStore(deserializedConstants)
-      val deserializedTree = ErgoTreeSerializer.DefaultSerializer.deserializeWithConstantInjection(c, treeBytes)
-      deserializedTree shouldEqual processedTree
-    }
-  }
-
```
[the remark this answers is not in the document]
greenhat: This test was the only usage of `ErgoTreeSerializer.deserializeWithConstantInjection`. I removed the method, so the need for this test was gone. The test above covers the use case.

greenhat:
@catena2w I've done everything, except "Error should be thrown by verifier, not a prover" in `DeserializeContext` type mismatch test. Does it mean that we should not apply context deserialization in prover (see [src](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/a472d96e344843ff80416fa118bf44b888dce346/src/main/scala/sigmastate/interpreter/ProverInterpreter.scala#L135))?

PR #482, opened 2019-05-01
Check script tree depth limit during deserialization

### LINKED ISSUE #270: Check script tree depth limit during deserialization
- This issue defines behavior of script interpreter when given enormously deep tree during transaction validation
-  Script trees with at least `MaxTreeDepth` levels should be supported
- When transaction is validated the script is deserialized for execution.
- The script should be checked by deserializer for it's depth using `level` field in the deserializer.
- The scripts with more than `MaxTreeDepth` levels are considered malicious which should lead to deserializer throwing the corresponding exception.
- every recursive serializer should check `MaxTreeDepth` (ValueSerializer, DataSerializer, SigmaBoolean.serializer)

greenhat:
Close #270 

- [x] check that `SigmaByteReader.level` corresponds to the actual recursive call depth of the serializer(s);
- [x] count depth in `DataSerializer`, 
- [x] count depth in `SigmaBoolean.serializer`;

### DIFF
--- a/src/main/scala/sigmastate/utils/SigmaByteReader.scala
+++ b/src/main/scala/sigmastate/utils/SigmaByteReader.scala
@@ -1,16 +1,16 @@
 package sigmastate.utils
 
-import java.nio.ByteBuffer
-
-import scorex.util.serialization.{Reader, VLQByteBufferReader}
+import scorex.util.Extensions._
+import scorex.util.serialization.Reader
 import sigmastate.SType
 import sigmastate.Values.SValue
+import sigmastate.lang.exceptions.DeserializeCallDepthExceeded
 import sigmastate.serialization.{ConstantStore, TypeSerializer, ValDefTypeStore, ValueSerializer}
-import scorex.util.Extensions._
 
 class SigmaByteReader(val r: Reader,
                       var constantStore: ConstantStore,
-                      var resolvePlaceholdersToConstants: Boolean)
+                      var resolvePlaceholdersToConstants: Boolean,
+                      val maxTreeDepth: Int)
   extends Reader {
 
   val valDefTypeStore: ValDefTypeStore = new ValDefTypeStore()
@@ -67,7 +67,12 @@ class SigmaByteReader(val r: Reader,
 
   private var lvl: Int = 0
   @inline def level: Int = lvl
-  @inline def level_=(v: Int): Unit = lvl = v
+  @inline def level_=(v: Int): Unit = {
+    if (v > maxTreeDepth)
+      throw new DeserializeCallDepthExceeded(s"nested value deserialization call depth($v) exceeds allowed maximum $maxTreeDepth")
+    lvl = v
+  }
+
   @inline def getValues(): IndexedSeq[SValue] = {
     val size = getUInt().toIntExact
     val xs = new Array[SValue](size)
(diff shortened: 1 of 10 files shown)


src/main/scala/sigmastate/utils/SigmaByteReader.scala
```diff
@@ -1,16 +1,16 @@
...
 import sigmastate.Values.SValue
-import sigmastate.serialization.{ConstantStore, TypeSerializer, ValDefTypeStore, ValueSerializer}
-import scorex.util.Extensions._
+import sigmastate.lang.exceptions.DeserializeCallDepthExceeded
+import sigmastate.serialization.{ConstantStore, SigmaSerializer, TypeSerializer, ValDefTypeStore, ValueSerializer}
 
 class SigmaByteReader(val r: Reader,
                       var constantStore: ConstantStore,
-                      var resolvePlaceholdersToConstants: Boolean)
+                      var resolvePlaceholdersToConstants: Boolean,
```
aslesarenko: @greenhat do we really need this fields to be `var`?

src/test/scala/sigmastate/serialization/DeserializationResilience.scala
```diff
@@ -47,4 +64,135 @@ class DeserializationResilience extends SerializationSpecification {
...
+        new ConstantStore(),
+        resolvePlaceholdersToConstants = false,
+        maxTreeDepth = SigmaSerializer.MaxTreeDepth) {
+      val levels: mutable.ArrayBuilder[Int] = mutable.ArrayBuilder.make[Int]()
+      override def level_=(v: Int): Unit = {
+        if (v >= super.level) {
+          // going deeper (save new depth to account added depth by the caller)
+          levels += v
+        } else {
+          // going up (save previous depth to account added depth be the caller)
```
aslesarenko: This comment is not clear

src/test/scala/sigmastate/serialization/TypeSerializerSpecification.scala
```diff
@@ -93,12 +91,6 @@ class TypeSerializerSpecification extends SerializationSpecification {
       Array[Byte](TupleTypeCode, 5, SLong.typeCode, SLong.typeCode, SByte.typeCode, SBoolean.typeCode, SInt.typeCode))
   }
 
-  property("tuple of tuples crazy deep") {
```
aslesarenko: @greenhat I don't see it is replaced by something similar?

src/main/scala/sigmastate/utils/SigmaByteReader.scala
greenhat: Yes, we're setting new values for existing reader (see https://github.com/ScorexFoundation/sigmastate-interpreter/blob/471450d25ca7e129b13193dafea7adfeca95f994/src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala#L33 )

src/test/scala/sigmastate/serialization/TypeSerializerSpecification.scala
greenhat: I replaced it with two tests in `DeserializationResilience` (see last two tests https://github.com/ScorexFoundation/sigmastate-interpreter/blob/4eee4bf45c60d7a73bf6111ebb4ae22ddeb71a59/src/test/scala/sigmastate/serialization/DeserializationResilience.scala#L185 ). It's not exactly the same test, but I don't like this one, since it's kind of `reverse-engineered` from serialization format and thus is too fragile.

src/test/scala/sigmastate/serialization/DeserializationResilience.scala
greenhat: Thanks! I rephrased the comment. 
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/3818e6f90c56a12080ac2de17078e91fc42a75f8/src/test/scala/sigmastate/serialization/DeserializationResilience.scala#L80

PR #483, opened 2019-05-02
I437 soft forkability

aslesarenko:


src/main/scala/org/ergoplatform/Validation.scala
```diff
@@ -0,0 +1,153 @@
...
+case class ReplacedRuleException(vs: ValidationSettings, replacedRule: ValidationRule, replacement: ReplacedRule)
+  extends SoftForkException(s"Rule ${replacedRule.id} was replaced with ${replacement.newRuleId}")
+
+case class ChangedRuleException(vs: ValidationSettings, changedRule: ValidationRule, change: ChangedRule)
+  extends SoftForkException(s"Rule ${changedRule.id} was changed with ${change}")
+
+/**
+  * Configuration of validation.
+  */
+abstract class ValidationSettings {
```
catena2w: How should `ValidationRule` and `RuleStatus`it be specified on proposing a soft-fork?

src/test/scala/sigmastate/SoftForkabilitySpecification.scala
```diff
@@ -0,0 +1,170 @@
...
+
+  implicit lazy val IR = new TestingIRContext
+  lazy val prover = new ErgoLikeTestProvingInterpreter()
+  lazy val verifier = new ErgoLikeTestInterpreter
+  lazy val invalidPropV1 = compile(Interpreter.emptyEnv,
+    """{
+     |  HEIGHT > 100 && OUTPUTS.size == 1
+     |}""".stripMargin).asBoolValue
+  lazy val invalidTxV1 = createTransaction(createBox(100, invalidPropV1.asSigmaProp, 1))
+  lazy val invalidTxV1bytes = invalidTxV1.messageToSign
```
catena2w: `.bytes` should be here (and later), instead of `.messageToSign`

src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala
```diff
@@ -1,42 +1,79 @@
...
     serializeHeader(ergoTree, w)
     ValueSerializer.serialize(ergoTree.root, w)
     w.toBytes
   }
 
+  /** Default serialization of ErgoTree. Doesn't apply any transformations and guarantee to preserve original
+    * structure after deserialization. */
+  def serializeErgoTree(ergoTree: ErgoTree): Array[Byte] = {
+    val bytes = serializeWithoutSize(ergoTree)
+    if (ergoTree.hasSize) {
```
catena2w: What if `ergoTree.hasSize = false` and also there is an op-code from soft-fork? Looks like new nodes will accept the script, while the old one won't and will reject the block.

src/main/scala/org/ergoplatform/Validation.scala
aslesarenko: amended the class doc.

src/test/scala/sigmastate/SoftForkabilitySpecification.scala
aslesarenko: ErgoLikeTransation doesn't have `bytes`, its id is computed like this
`  override lazy val id: ModifierId = Blake2b256.hash(messageToSign).toModifierId`

aslesarenko:
Replaced by #500

PR #501, opened 2019-05-20
Constants check

gostkin:


src/main/scala/sigmastate/trees.scala
```diff
@@ -238,7 +239,7 @@ case class AtLeast(bound: Value[SInt.type], input: Value[SCollection[SSigmaProp.
 
 object AtLeast {
 
-  val MaxChildrenCount = 255
+  val MaxChildrenCount = SigmaConstants.MaxChildrenCount.value
```
aslesarenko: @gostkin better name for the constant `MaxChildrenCountForAtLeastOp`

PR #503, opened 2019-05-21
Soft-Forkability: table of allowed opcodes in cost function

### LINKED ISSUE #499: Soft-Forkability: table of allowed opcodes in cost function
1) The validation rule `CheckCostFuncOperation` should be added 
2) it should be implemented similar to `CheckValidOpCode` rule.
3) is should be called from verifyCostFunc passing opCode of each graph node
4) New operation can be allowed in cost function by putting their op codes in block extension section (see `CheckValidOpCode`)

greenhat:
Close #499 

Todo:
- [x] issue `OpCodeExtra` codes for all ops in `getOpCodeEx`;
- [x] add `SoftForkabilitySpec` test for `OpCodeExtra`;
- [x] make op code types `TaggedType`;


src/main/scala/sigmastate/eval/Evaluation.scala
```diff
@@ -127,15 +132,75 @@ trait Evaluation extends RuntimeCosting { IR: IRContext =>
     case _ => false
   }
 
-  /** Checks if the function (Lambda node) given by the simbol `costF` contains only allowed operations
+  private val _allowedOpCodesInCosting = HashSet[OpCode](
+    Undefined,
```
aslesarenko: @greenhat is it really allowed?

src/main/scala/sigmastate/eval/Evaluation.scala
```diff
@@ -127,15 +132,75 @@ trait Evaluation extends RuntimeCosting { IR: IRContext =>
...
     * in the schedule. */
   def verifyCostFunc(costF: Rep[Any => Int]): Try[Unit] = {
     val Def(Lambda(lam,_,_,_)) = costF
     Try {
       lam.scheduleAll.forall { te =>
-        val ok = isValidCostPrimitive(te.rhs)
-        if (!ok) !!!(s"Invalid primitive in Cost function: ${te.rhs}")
-        ok
+        val isValidCostPrim = isValidCostPrimitive(te.rhs)
+        if (!isValidCostPrim) !!!(s"Invalid primitive in Cost function: ${te.rhs}")
```
aslesarenko: @greenhat this check it not necessary, validation rule should do the job.

src/main/scala/sigmastate/eval/Evaluation.scala
```diff
@@ -127,15 +132,75 @@ trait Evaluation extends RuntimeCosting { IR: IRContext =>
...
+    // todo implement
+//    true
+    _allowedOpCodesInCosting.contains(opCode)
+  }
+
+
+  private def getOpCode(d: Def[_]): OpCode = d match {
+      // todo implement
+//    case _: Const[_] => true
+    case _: OpCost | _: PerKbCostOf | _: Cast[_] => Undefined
```
aslesarenko: Undefined in not a good name for this.

src/main/scala/sigmastate/eval/Evaluation.scala
greenhat: My thoughts exactly. I'll remove it when `CheckCostFuncOperation` is ready.

src/main/scala/sigmastate/eval/Evaluation.scala
greenhat: @aslesarenko I agree the name `Undefined` is a bit confusing. But it's comment says "  // serialization is not required" which fit this case nicely. Maybe `NotForSerialization` or something similar. How does it sound to you?

aslesarenko: ok, let's leave Undefined

src/main/scala/sigmastate/eval/TreeBuilding.scala
```diff
@@ -48,7 +48,7 @@ trait TreeBuilding extends RuntimeCosting { IR: IRContext =>
   type DefEnv = Map[Sym, (Int, SType)]
 
   object IsArithOp {
-    def unapply(op: EndoBinOp[_]): Option[Byte] = op match {
+    def unapply(op: EndoBinOp[_]): Option[OpCode] = op match {
```
aslesarenko: This code is part of transaction validation. So performance is critical. 
This however will produce a lot of Some object in the heap.
Nullable will not help here as well.
So please reimplement avoiding allocations on a code pass.

src/main/scala/sigmastate/eval/Evaluation.scala
```diff
@@ -84,58 +89,117 @@ trait Evaluation extends RuntimeCosting { IR: IRContext =>
...
+      case SFuncM.sizeEnv(_) => Some()
+      case _: CSizePairCtor[_, _] | _: CSizeFuncCtor[_, _, _] | _: CSizeOptionCtor[_] | _: CSizeCollCtor[_] |
+           _: CSizeBoxCtor | _: CSizeContextCtor | _: CSizeAnyValueCtor => Some()
+      case _: CReplCollCtor[_] | _: PairOfColsCtor[_, _] => Some()
+      case CollM.sum(_, _) => Some()
+      case CBM.replicate(_, _, _) | CBM.fromItems(_, _, _) => Some()
+      case _: CostOf | _: SizeOf[_] => Some()
+      case SPCM.some(_) => Some()
+      case _ => None
+    }
```
aslesarenko: @greenhat the same performance argument applies here. 
Please avoid allocations.
Also here another possible optimization. 
Instead of checking each individual method we can check it's class
this will reduce the number of cases in `match`, hence reduce the number of the checks.

src/main/scala/sigmastate/eval/Evaluation.scala
aslesarenko: I suggest to remove this special code from the table and check it explicitly. This will also improve performance, because for all non-serializable nodes it will have table lookup time.

src/main/scala/sigmastate/eval/Evaluation.scala
```diff
@@ -84,58 +89,117 @@ trait Evaluation extends RuntimeCosting { IR: IRContext =>
...
+    PlusCode,
+    SelectFieldCode,
+    SelfCode,
+    SigmaPropBytesCode,
+    SizeOfCode,
+    SliceCode,
+    TupleCode,
+    UpcastCode,
+  )
+
```
aslesarenko: @greenhat please add 
`protected def allowedOpCodesInCosting: HashSet[OpCode]`
to make it possible to override this table in derived cakes.
This in particular will be useful in tests.

src/main/scala/org/ergoplatform/Validation.scala
```diff
@@ -263,6 +265,24 @@ object ValidationRules {
...
 
+  object CheckCostFuncOperation extends ValidationRule(1010,
+    "Check the opcode is allowed in cost function") {
+    def apply[Ctx <: IRContext, T](ctx: Ctx)(opCode: OpCode)(block: => T): T = {
+      def msg = s"Not allowed opCode = LastConstantCode + ${opCode.toUByte - OpCodes.LastConstantCode} in cost function"
+      def args = Seq(opCode)
+      validate(ctx.isAllowedOpCodeInCosting(opCode), new CosterException(msg, None), args, block)
+    }
+
+    override def isSoftFork(vs: ValidationSettings,
```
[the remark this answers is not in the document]
greenhat: Done.

src/main/scala/sigmastate/eval/TreeBuilding.scala
greenhat: Done.

src/main/scala/sigmastate/eval/Evaluation.scala
greenhat: Done.

src/main/scala/sigmastate/eval/Evaluation.scala
greenhat: Done.

src/main/scala/sigmastate/eval/Evaluation.scala
greenhat: Done.

aslesarenko: @greenhat well it doesn't make sense to rebuild this table one every method invocation.

src/main/scala/sigmastate/eval/Evaluation.scala
```diff
@@ -84,58 +89,161 @@ trait Evaluation extends RuntimeCosting { IR: IRContext =>
...
-    case CBM.replicate(_,_,_) | CBM.fromItems(_,_,_) => true
-    case BoxM.propositionBytes(_) | BoxM.bytesWithoutRef(_) | BoxM.getReg(_,_,_) => true
-    case OM.get(_) | OM.getOrElse(_,_) | OM.fold(_,_,_) | OM.isDefined(_) => true
+    case _: CReplCollCtor[_] | _: PairOfColsCtor[_, _] => true
+    case CollM.sum(_, _) => true
+    case CBM.replicate(_, _, _) | CBM.fromItems(_, _, _) => true
     case _: CostOf | _: SizeOf[_] => true
-    case _: Upcast[_,_] => true
-    case _: Apply[_,_] => true
     case SPCM.some(_) => true
```
aslesarenko: @greenhat I realised we cannot have such method if we want soft-forkability. 
This method strictly prohibit all the operations (mentioned in it) to be added in future. 
So we need to have a table of codes (which are NOT opCodes) but still some codes which we can add to block Extension section and check in SoftForkWhenCodeAdded.isSoftFork method. 
So please create a new table OpCodesEx and assign each operation a new code (as it is done in getOpCode)

src/main/scala/sigmastate/eval/Evaluation.scala
```diff
@@ -84,58 +89,161 @@ trait Evaluation extends RuntimeCosting { IR: IRContext =>
...
     case _ => false
   }
 
+  def getOpCode(d: Def[_]): OpCode = if (isNonSerializable(d)) Undefined else d match {
```
aslesarenko: isNonSerializable check should be new ValidationRule which extends SoftForkWhenCodeAdded.

greenhat: @aslesarenko Damn it, I had the same thoughts back then but dismissed them as being too complicated. :)

src/main/scala/sigmastate/eval/Evaluation.scala
greenhat: I'm struggling with implementation here. I've tried a few variants but eventually got to the dead end.

My line of thinking here:
1. *Every* op should resolve into an opcode (including `Size*Methods` ops);
2. Validation rule should check if opcode is allowed in costing;
3. Alter validation rule with block extension to allow new ops in costing via a soft-fort in the future;

What values should be in `OpCodeEx` table? All byte range values are already taken in `OpCodes`. Should we use a wider type (Int) and not intersect `OpCodeEx` values with `OpCodes`? In this case what about `ChangedRule.newValue` type (`Array[Byte]`)?

Actually, by resolving all ops in `isNonSerializable` to the `Undefined` opcode this method *allows* these ops in costing.

`isNonSerializable` name may be confusing. I assume we're talking about allowing ops in costing and not about if op is serializable. Cause this is handled in `CheckValidOpCode` rule.

aslesarenko: added descriptions

src/main/scala/org/ergoplatform/validation/ValidationRules.scala
```diff
@@ -163,6 +180,51 @@ object ValidationRules {
...
+                            ruleId: Short,
+                            status: RuleStatus, args: Seq[Any]): Boolean = (status, args) match {
+      case (ChangedRule(newValue), Seq(code: OpCodeExtra)) =>
+        decodeVLQUShort(newValue).contains(code)
+      case _ => false
+    }
+
+    @inline
+    private def unsignedOpCode(opCode: OpCodeExtra): Short =
+      if (opCode < 0) opCode.toByte.toUByte.toShort else opCode
```
aslesarenko: @greenhat Could you please explain what this means? OpCodeExtra values should never be negative.

src/main/scala/org/ergoplatform/validation/ValidationRules.scala
```diff
@@ -163,6 +180,51 @@ object ValidationRules {
...
+   * sequence.
+   * Note, we don't need to store a number of items in a sequence,
+   * because at the time of parsing we may assume that `ChangedRule.newValue`
+   * has correct length, so we just parse it until end of bytes (of cause
+   * checking consistency). */
+  object CheckCostFuncOperation extends ValidationRule(1014,
+    "Check the opcode is allowed in cost function") with SoftForkWhenCodeAdded {
+    def apply[Ctx <: IRContext, T](ctx: Ctx)(opCode: OpCodeExtra)(block: => T): T = {
+      def msg = s"Not allowed opCode $opCode in cost function"
+      def args = Seq(unsignedOpCode(opCode))
```
aslesarenko: This rule is inconsistent with the other rules checking the codes. 
They all put original code into `args`, here you modify it and put into `args`.
Also the `unsignedOpCode` function is not clear here (see my comment there)

src/main/scala/sigmastate/serialization/OpCodes.scala
```diff
@@ -31,9 +31,28 @@ trait ValueCodes extends TypeCodes {
...
+  * And we use VLQ to serialize Short values of extended codes.
+  *
+  * Examples:
+  * 1) For validation rule CheckValidOpCode we use OpCodes range, so we use single byte encoding.
+  * 2) For CheckCostFuncOperation we use 1-511 range and extended encoding (see docs)
+  */
 object OpCodes extends ValueCodes {
 
   type OpCode = Byte
+  type OpCodeExtra = Short
```
aslesarenko: @greenhat Let's make it TaggedType. And also assume all codes are positive values, so they are densely packed in 0-511 range. Please revise all the places where OpCodeExtra is used.

src/main/scala/org/ergoplatform/validation/ValidationRules.scala
greenhat: Just as I was writing an explanation I found a way to eliminate it. :) Here is the fix - https://github.com/ScorexFoundation/sigmastate-interpreter/pull/503/commits/ccbdd58c50e4e205882f8f446c243a9d76ac887d

src/main/scala/org/ergoplatform/validation/ValidationRules.scala
greenhat: Thanks! Fixed and added comments.

src/main/scala/sigmastate/serialization/OpCodes.scala
greenhat: Resolved in https://github.com/ScorexFoundation/sigmastate-interpreter/pull/503/commits/b1caa8c87d2d569591545f15061c2984c3500a2c

src/test/scala/sigmastate/SoftForkabilitySpecification.scala
```diff
@@ -313,4 +312,29 @@ class SoftForkabilitySpecification extends SigmaTestingData {
...
+    val v2vs = vs.updated(CheckCostFuncOperation.id,
+      ChangedRule(CheckCostFuncOperation.encodeVLQUShort(Seq(Height.opCode))))
+    checkRule(CheckCostFuncOperation, v2vs, {
+      val IR.Pair(calcF, _) = IR.doCostingEx(emptyEnv, exp, okRemoveIsProven = false)
+      // use calcF as costing function to have forbidden (not allowed) op (Height) in the costing function
+      CheckCostFunc(IR)(IR.asRep[Any => Int](calcF)) { }
+    })
+  }
+
+  property("CheckCostFuncOperation rule (OpCodeExtra") {
```
aslesarenko: how this test is different from the previous?
Looks like it is checking the same, so it doesn't improve code coverage.

src/main/scala/sigmastate/serialization/OpCodes.scala
```diff
@@ -23,168 +29,240 @@ trait ValueCodes extends TypeCodes {
...
 object OpCodes extends ValueCodes {
 
-  type OpCode = Byte
+  object OpCode extends TaggedType[Byte]
+  type OpCode = OpCode.Type
+
+  object OpCodeExtra extends TaggedType[Short]
+  type OpCodeExtra = OpCodeExtra.Type
+
+  implicit def opCodeToOpCodeExtra(oc: OpCode): OpCodeExtra = OpCodeExtra @@ oc.toUByte.toShort
```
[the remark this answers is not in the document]
greenhat: Fixed.

src/test/scala/sigmastate/SoftForkabilitySpecification.scala
greenhat: It tests that "OpCodeExtra" op can be enabled. First one tested "OpCode" op.

Issue #452, opened 2019-03-25
Unexpected Costing Procedure Behavior  


kushti:
It is expected that with growing input cost is also growing. However, this is not the case for current costing procedure. Take a look to "simple ico example - fundraising stage only" in the ico-example branch. Depending on "funderBoxCount" parameter value, costs are as follows:
2 | 4248
20 | "Estimated expression complexity 1711276242 exceeds the limit 1000000"
200 | 3800 
2000 | 32600
20000 | 325800

gostkin:
#505  shows that now everything is fine

PR #505, opened 2019-05-23
Ico example issue

gostkin:
outp-full
outp-small-full
We see a linear-like costing growth by founders count in `simple ico example - fundraising stage only`. It seems to be a correct thing, so it's unclear, where's the bug. https://github.com/ScorexFoundation/sigmastate-interpreter/issues/452
*Note:* adapted code for datasets production is attached, plotting code is also attached (`data/GenPlots.ipynb`)


kushti:
@gostkin seems about right, there's no need to merge I guess? If so, lets close both the pull request and the issue.

gostkin:
@kushti yes, I think datasets/graphs shouldn't be placed to sigma

PR #507, opened 2019-05-25
Constants check

gostkin:
Simple case (without soft-fork support for constants)

### DIFF
--- a/src/main/scala/org/ergoplatform/ErgoConstants.scala
+++ b/src/main/scala/org/ergoplatform/ErgoConstants.scala
@@ -7,6 +7,9 @@ case class SizeConstant[T: Numeric](value: T, id: Short, description: String) {
   def get: T = value
 }
 
+/**
+  * Fundamental constants that are used in sigma's logic and checks
+  */
 object ErgoConstants {
 
   object MaxInputSize extends SizeConstant[Int](1024 * 1024 * 1, 1,
--- a/src/main/scala/org/ergoplatform/ErgoConstants.scala
+++ b/src/main/scala/org/ergoplatform/ErgoConstants.scala
@@ -24,7 +24,7 @@ object ErgoConstants {
     "Max bytearray length") {
   }
 
-  object MaxTokens extends SizeConstant[Byte](4, 6,
+  object MaxTokens extends SizeConstant[Byte](255, 6,
     "Tokens count should not be greater than provided value") {
   }
 
--- a/src/main/scala/org/ergoplatform/ErgoConstants.scala
+++ b/src/main/scala/org/ergoplatform/ErgoConstants.scala
@@ -12,11 +12,11 @@ case class SizeConstant[T: Numeric](value: T, id: Short, description: String) {
   */
 object ErgoConstants {
 
-  object MaxInputSize extends SizeConstant[Int](1024 * 1024 * 1, 1,
+  object MaxInputSize extends SizeConstant[Int](64 * 1024 * 1, 1,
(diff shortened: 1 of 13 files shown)


src/main/scala/org/ergoplatform/ErgoConstants.scala
```diff
@@ -0,0 +1,67 @@
...
+/**
+  * Fundamental constants that are used in sigma's logic and checks
+  */
+object ErgoConstants {
+
+  object MaxInputSize extends SizeConstant[Int](1024 * 1024 * 1, 1,
+    "Input size should not be greater then provided value") {
+  }
+
+  object MaxTreeDepth extends SizeConstant[Int](110, 2,
```
kushti: While 110 is probably OK, a more conservative and maybe safer option is to lower it to e.g. to 64.

src/main/scala/org/ergoplatform/ErgoConstants.scala
```diff
@@ -0,0 +1,67 @@
...
+    "Max tree depth should not be greater then provided value") {
+  }
+
+  object MaxByteArrayLength extends SizeConstant[Int](10000, 3,
+    "Max bytearray length") {
+  }
+
+  object MaxTokens extends SizeConstant[Byte](4, 6,
```
kushti: ErgoBox.MaxTokens is not being using in Ergo and actually outdated. There's alternative "val MaxAssetsPerBox = 255" in ErgoTransaction. Please set MaxTokens to the same value, and then we'll use MaxTokens everywhere instead of MaxAssetsPerBox.

src/main/scala/org/ergoplatform/ErgoConstants.scala
gostkin: in this case a test `big byte array with a lot of operations` from `SpamSpecification` fails, probably we need to change the test then

gostkin: fixed

src/main/scala/org/ergoplatform/ErgoConstants.scala
gostkin: According to whitepaper, "The maximum number of tokens in a box is set to 4", that's what this constant represent. I suppose it will be wrong to change

src/main/scala/org/ergoplatform/ErgoConstants.scala
```diff
@@ -0,0 +1,67 @@
...
+case class SizeConstant[T: Numeric](value: T, id: Short, description: String) {
+  def get: T = value
+}
+
+/**
+  * Fundamental constants that are used in sigma's logic and checks
+  */
+object ErgoConstants {
+
+  object MaxInputSize extends SizeConstant[Int](1024 * 1024 * 1, 1,
```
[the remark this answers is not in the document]
gostkin: that leads to error in ergo tests:
```
[info] - BlockTransactions serialization *** FAILED ***
[info]   InputSizeLimitExceeded was thrown during property evaluation.
[info]     Message: input size 65696 exceeds 65536
[info]     Occurred when passed generated values (
[info]       arg0 = BlockTransactions(id: b80e7cb49f6e0cdd68a655363f6d9f5f2ed5edbb61d7c322f904cbc13114fc3c, headerId: 
```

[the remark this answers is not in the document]
gostkin: by the way, all constants were taken from different places in sigma, so I think there was a reason in its values before

PR #511, opened 2019-05-29
Implement SNumericType.toBytes and toBits (SNumericTypeCoster)

greenhat:


sigma-impl/src/main/scala/sigma/util/Extensions.scala
```diff
@@ -152,6 +152,13 @@ object Extensions {
         throw new ArithmeticException("Int overflow")
       x.toInt
     }
+
+    /** Returns a big-endian representation of this Int in a collection of bytes.
+      * For example, the Int value {@code 0x12131415} would yield the
+      * byte array {@code {0x12, 0x13, 0x14, 0x15}}.
+      * @since 2.0
```
aslesarenko: since 2.1

src/main/scala/sigmastate/eval/CostingRules.scala
```diff
@@ -627,4 +627,23 @@ trait CostingRules extends SigmaLibrary { IR: RuntimeCosting =>
 
   object SigmaDslBuilderCoster extends CostingHandler[SigmaDslBuilder]((obj, m, costedArgs, args) => new SigmaDslBuilderCoster(obj, m, costedArgs, args))
 
+  /** Costing rules for SNumericType methods */
+  class SNumericTypeCoster(obj: RCosted[AnyVal], method: SMethod, costedArgs: Seq[RCosted[_]], args: Seq[Sym]) extends Coster[AnyVal](obj, method, costedArgs, args) {
```
aslesarenko: Looks like it is better to use separate Coster classes and objects for each numeric type SByte, Short, SInt,

sigma-impl/src/main/scala/sigma/util/Extensions.scala
greenhat: Thanks! Fixed.

kushti:
done separately in 6.0

PR #519, opened 2019-06-05
Fixed memory leak and other performance optimizations

aslesarenko:
- memory leak fixed in (resetContext is doing additional cleanups)
- removed dependency on scalan-meta.jar
- removed dependancy on scala-compiler.jar
- various performance optimisations (based on profiling)

### DIFF
--- a/src/main/scala/sigmastate/eval/RuntimeCosting.scala
+++ b/src/main/scala/sigmastate/eval/RuntimeCosting.scala
@@ -159,8 +159,15 @@ trait RuntimeCosting extends CostingRules with DataCosting with Slicing { IR: IR
   }
 
   def costOf(costOp: CostOf, doEval: Boolean): Rep[Int] = {
-    val res = if (doEval) toRep(costOp.eval)
-    else (costOp: Rep[Int])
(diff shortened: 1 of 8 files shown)


src/main/scala/sigmastate/eval/RuntimeCosting.scala
```diff
@@ -159,8 +159,15 @@ trait RuntimeCosting extends CostingRules with DataCosting with Slicing { IR: IR
   }
 
   def costOf(costOp: CostOf, doEval: Boolean): Rep[Int] = {
-    val res = if (doEval) toRep(costOp.eval)
-    else (costOp: Rep[Int])
+    val res = if (doEval) {
+      val c = Const(costOp.eval)
+      // optimized hot-spot: here we avoid rewriting which is done by reifyObject
+      findOrCreateDefinition(c, c.self)
```
catena2w: Please add function comments where it is implemented

aslesarenko: Done

aslesarenko:
merged into #523

PR #516, opened 2019-06-02
Add more tests to CostingSpecification

greenhat:

PR #523, opened 2019-06-10
Simplify cost graphs and speedup costing performance

aslesarenko:
- memory leak fixed in (resetContext is doing additional cleanups)
- removed dependency on scalan-meta.jar
- removed dependancy on scala-compiler.jar
- various performance optimisations (based on profiling)
- costing speedup by using max constants instead of variables (which simplifies cost graph and hence reduce time)
- decreased timeout in SpamSpeciafication
- fixed limit check in ValueSerializer 
- MaxBoxSize set to 4 * 1024
- check MaxBoxSize and MaxPropositionBytes limits in deserialization


src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
```diff
@@ -122,10 +122,12 @@ object ErgoBoxCandidate {
     }
 
     def parseBodyWithIndexedDigests(digestsInTx: Option[Coll[TokenId]], r: SigmaByteReader): ErgoBoxCandidate = {
+      val previousPositionLimit = r.positionLimit
+      r.positionLimit = r.position + ErgoBox.MaxBoxSize
       val value = r.getULong()
       val tree = ErgoTreeSerializer.DefaultSerializer.deserializeErgoTree(r)
       val creationHeight = r.getUInt().toInt
-      val addTokensCount = r.getByte()
```
catena2w: Looks like this breaks binary format?

src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala
```diff
@@ -127,15 +128,18 @@ class ErgoTreeSerializer {
     deserializeErgoTree(r)
   }
 
-  def deserializeErgoTree(r: SigmaByteReader): ErgoTree = {
+  def deserializeErgoTree(r: SigmaByteReader,
+                          maxTreeSizeBytes: Int = SigmaSerializer.MaxPropositionSize): ErgoTree = {
```
catena2w: Is this really a function parameter? Let's either remove it from parameters list or remove the default value

src/test/scala/sigmastate/CostingSpecification.scala
```diff
@@ -197,8 +197,10 @@ class CostingSpecification extends SigmaTestingData {
     cost(s"{ $rootTree.isInsertAllowed }") shouldBe (AccessRootHash + selectField)
     cost(s"{ $rootTree.isUpdateAllowed }") shouldBe (AccessRootHash + selectField)
     cost(s"{ $rootTree.isRemoveAllowed }") shouldBe (AccessRootHash + selectField)
-    cost(s"{ $rootTree.updateDigest($rootTree.digest) == $rootTree }") shouldBe (AccessRootHash + selectField + newAvlTreeCost + comparisonPerKbCost)
-    cost(s"{ $rootTree.updateOperations(1.toByte) == $rootTree }") shouldBe (AccessRootHash + newAvlTreeCost + comparisonPerKbCost + constCost)
+    cost(s"{ $rootTree.updateDigest($rootTree.digest) == $rootTree }") shouldBe
+      (AccessRootHash + selectField + newAvlTreeCost + comparisonCost /* for isConstantSize AvlTree type */)
+    cost(s"{ $rootTree.updateOperations(1.toByte) == $rootTree }") shouldBe
+      (AccessRootHash + newAvlTreeCost + comparisonCost + constCost)
```
catena2w: Let's add tests for changed costing rules, they should be quite easy now.

src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
greenhat: @catena2w I don't think so. We're writing unsigned values, but we were reading signed values. I just fixed reading.

aslesarenko: nope `@inline override def getUByte(): Int = getByte() & 0xFF`

src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala
greenhat: @catena2w Probably not. Easing of testing does not count. Apart from that, decoupling from `SigmaSerializer` constant does not worth it alone. Removing default value will just force this constant leaking out. Let me remove the parameter.

aslesarenko: removed default value.

src/test/scala/sigmastate/CostingSpecification.scala
aslesarenko: only this tests are affected and changed. Others are not affected.
More test will be added in #516

PR #524, opened 2019-06-10
Remove/soften "whenever" requirements to reduce the number of discarded evaluations

greenhat:


src/test/scala/sigmastate/TestingInterpreterSpecification.scala
```diff
@@ -28,7 +29,8 @@ class TestingInterpreterSpecification extends SigmaTestingCommons {
       fakeSelf)
 
   property("Reduction to crypto #1") {
-    forAll() { (h: Int) =>
+    forAll() { i: Int =>
+      val h = i.toAbs
```
aslesarenko: I think it is better to use generator of positive values instead.

PR #548, opened 2019-06-26
Fix exp costing

gostkin:


src/test/scala/sigmastate/utxo/SpamSpecification.scala
```diff
@@ -685,11 +682,6 @@ class SpamSpecification extends SigmaTestingCommons with ObjectGenerators {
 
   property("repeat large loop: exp") {
     // to do: convert to nested
-    /*
-      Passes because costing is not implemented for exp
-        java.lang.NoSuchMethodException: sigmastate.eval.CostingRules$GroupElementCoster.exp(scalan.Base$Exp)
-
-     */
```
aslesarenko: @gostkin look at other tests with repeatScript. The script size should grow with scale until CostLimitException is rised.

PR #555, opened 2019-06-27
Create tests for lazy evaluation of `OR`, `AND`

### LINKED ISSUE #247: Create tests for lazy evaluation of `OR`, `AND` in addition to existing tests
I suppose we should short circuit on first `true` element and do not evaluate the rest of the elements.
Here is the code example that now is failing and should work:
`true || (1/0 == 1)` 

Reported in #246

greenhat:
Close #247

### DIFF
--- a/src/test/scala/special/sigma/SigmaDslTest.scala
+++ b/src/test/scala/special/sigma/SigmaDslTest.scala
@@ -994,30 +994,40 @@ class SigmaDslTest extends PropSpec
   }
 
   property("lazy || and && boolean equivalence") {
-    checkEq(func[Boolean, Boolean]("{ (x: Boolean) => x || (1/0 == 1) }")){ (x: Boolean) =>
-      x || (1 / 0 == 1)
-    }(true)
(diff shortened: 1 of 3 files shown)


src/test/scala/special/sigma/SigmaDslTest.scala
```diff
@@ -973,4 +973,16 @@ class SigmaDslTest extends PropSpec
...
+    checkEq(func[Boolean, Boolean]("{ (x: Boolean) => x || (1/0 == 1) }")){ (x: Boolean) =>
+      x || (1 / 0 == 1)
+    }(true)
+  }
+
+  property("lazy && boolean equivalence") {
+    checkEq(func[Boolean, Boolean]("{ (x: Boolean) => (x && (1/0 == 1)) == false }")){ (x: Boolean) =>
+      (x && (1 / 0 == 1)) == false
+    }(false)
+  }
```
aslesarenko: @greenhat Great! Let's add more eq tests here with 2, 3, 4 nested && and || levels in both left and right arguments.
Around 15 tests that cover all possible situations.

greenhat: Done.

PR #558, opened 2019-06-30
Fold with deep nested folds as zeroes

greenhat:


greenhat:
@aslesarenko the test is on the edge of the timeout, here is my local run:
```
Initalization Cost: 10000; Complexity: 587530
Verify time: 756 millis; SerializedSize: 1979;  Defs: 32772
Timeout: 646 millis
Rejection cause: sigmastate.lang.exceptions.CostLimitException: Estimated execution cost 1095975 exceeds the limit 1000000
```

greenhat:
Consumed by #568 and #572.

PR #559, opened 2019-07-01
V3.0.1 candidate

aslesarenko:


aslesarenko:
Split into #566 and #568

PR #569, opened 2019-07-09
Optimization of ErgoBox deserialization

aslesarenko:
Key changes:
1) ErgoBox deserialization is tuned and optimised 
2) ValidationRules rewritten to avoid allocation as they turn out to be a hot spot
3) ErgoTree propositionBytes are saved in ErgoTree during deserialization, this allow to avoid calling ErgoTreeSerializer.serialize to obtain them later.
4) Other simple performance improvements on the script execution code path.


aslesarenko:
@greenhat Comments like `Hotspot please don't "beautify"!`? :)

greenhat:
@aslesarenko Yeah, something like that. :)

PR #572, opened 2019-07-22
Move spam tests

greenhat:


build.sbt
```diff
@@ -253,3 +253,41 @@ commands += Command.command("ergoItTest") { state =>
...
+  log.info(s"Cleaning $spamPath")
+  s"rm -rf ${spamPath.absolutePath}" !
+
+  log.info(s"Cloning spam tests branch $spamBranch into ${spamPath.absolutePath}")
+  Process(Seq("git", "clone",  "-b", spamBranch, "--single-branch", "git@github.com:greenhat/sigma-spam.git", spamPath.absolutePath),
+    cwd.getAbsoluteFile,
+    envVars: _*) !
+
+  log.info(s"Running spam tests in $spamPath with Sigmastate version $sigmastateVersion")
+  val res = Process(Seq("sbt", task), spamPath, envVars: _*) !
```
aslesarenko: When this `sbt test` process is executed, how current sigma become in dependency? Via SIGMASTATE_VERSION?

greenhat: @aslesarenko Yes, it's set in the `SIGMASTATE_VERSION` env var and is then read in sigma-spam (see https://github.com/greenhat/sigma-spam/blob/7fef65ea7e9af069ca1492eaf0ddf26559bb8fb3/build.sbt#L19 )

PR #590, opened 2019-09-08
add missing SMethod.docInfo for property calls + implement substConst in the compiler 

greenhat:


aslesarenko:
@greenhat Let's merge it according to #595 plan

PR #596, opened 2019-09-18
Exact operations

aslesarenko:
Includes #566 (See description there)
In addition:
- checking overflow of `+`, `-`, `*` operations 
- fixed Pay2SHAddress construction from ErgoTree (added missing test)

This version of Sigmastate-interpreter is semantically equivalent (except overflow checking) with v3.x
series of released versions.


greenhat:
@aslesarenko should it be rebased onto #568 branch?

aslesarenko:
No, I plan to make little change, so that it can be merged into v3.x. Actually remove `exact` operations.
Then we can roll them back in v4.0 as a small patch.

common/src/main/scala/scalan/util/StringUtil.scala
```diff
@@ -0,0 +1,67 @@
+package scalan.util
+
+import spire.syntax.all.cfor
+
+object StringUtil {
+  final def quote(x: Any) = "\"" + x + "\""
+
+  /** Uppercase the first character. */
```
kushti: the comment seems to be wrong

PR #607, opened 2019-09-30
Any value encoder

knizhnik:


src/main/scala/sigmastate/serialization/DataJsonEncoder.scala
```diff
@@ -36,6 +36,15 @@ object DataJsonEncoder {
     ErgoAlgos.encode(bytes).asJson
   }
 
+  def encode(v: AnyValue): Json = {
```
aslesarenko: to be consistent with the method below it should be named `encodeAnyValue`

src/main/scala/sigmastate/serialization/DataJsonEncoder.scala
```diff
@@ -253,4 +262,11 @@ object DataJsonEncoder {
     val (data, _) = decodeWithTpe(json)
     data
   }
+
+  def decodeAnyValue(json: Json): (AnyValue) = {
+    val tpe = SigmaParser.parseType(json.hcursor.downField("type").focus.get.asString.get)
+    val value = json.hcursor.downField("value").focus.get
+    val data = decodeData(value, tpe)
+    TestValue(data,Evaluation.stypeToRType(tpe).asInstanceOf[RType[Any]])
```
aslesarenko:
```suggestion
    TestValue(data, Evaluation.stypeToRType(tpe).asInstanceOf[RType[Any]])
```

knizhnik: Fixed

src/main/scala/sigmastate/serialization/DataJsonEncoder.scala
knizhnik: Renamed

Issue #609, opened 2019-10-02
Proving contract properties (formal verification)


greenhat:
Document - https://github.com/ScorexFoundation/sigmastate-interpreter/pull/610

code - https://github.com/ScorexFoundation/sigmastate-interpreter/pull/617

# Plan:
## Generate ErgoTree from SigmaDsl. 
- [x] explore using macros for ErgoTree generation; 

## Crowdfunding Contract:
- [x] some properties of the Crowdfunding contract are already proven, see [source](http://github.com/ScorexFoundation/sigmastate-interpreter/blob/1f65d90cdab0388acc73bfa43797e39d3d9e32bc/contract-verification/src/main/scala/sigmastate/verification/contract/CrowdFundingContractVerification.scala#L34-L76)
- [ ] find and prove more properties.

### Rewrite ICO contracts:
- [ ] simplify/reduce the proof, the requirement is too complex [here](http://github.com/ScorexFoundation/sigmastate-interpreter/blob/1f65d90cdab0388acc73bfa43797e39d3d9e32bc/contract-verification/src/main/scala/sigmastate/verification/contract/ICOContractVerification.scala#L208-L208) 
- [ ] eliminate the use of `Option.getOrElse`;
### Prove ICO contract properties:
- [x] some properties of the ICO funding contract are already proven, see [source](http://github.com/ScorexFoundation/sigmastate-interpreter/blob/1f65d90cdab0388acc73bfa43797e39d3d9e32bc/contract-verification/src/main/scala/sigmastate/verification/contract/ICOContractVerification.scala#L197-L238)
- [ ] find and prove more properties for ICO funding contract.
- [ ] find and prove properties for ICO issuance contract.
- [ ] find and prove properties for ICO withdrawal contract.

greenhat:
Having spent some time proving the contracts I'm seeing ErgoTree generation as the most "unknown" part of the process. I'd start drafting its implementation ASAP. This could put some additional requirements on the proving (e. g. processing assertions, pre-, post-conditions, encoding collection/other ops, etc.).

aslesarenko:
Consider moving the text above into document and create a PR.
This will be easier to collaborate.

aslesarenko:
I think the roadmap items in the description would be useful.

PR #610, opened 2019-10-02
Plan for contract formal verification

greenhat:
#609

### DIFF
--- a/docs/proving-contract-properties.md
+++ b/docs/proving-contract-properties.md
@@ -1,0 +1,26 @@
+# Goal:
+Verify properties of a contract starting from simple (Crowdfunding) to more complex (ICO).
+
+# Workflow:
+Write the contract in SigmaDls.
+During the compilation:
+* Our macros generate ErgoTree;
+* [Stainless](https://github.com/epfl-lara/stainless) verifies contract properties;
+
+# Plan:
+
+## Crowdfunding Contract:
+- [x] some properties of the Crowdfunding contract are already proven, see [source](http://github.com/ScorexFoundation/sigmastate-interpreter/blob/1f65d90cdab0388acc73bfa43797e39d3d9e32bc/contract-verification/src/main/scala/sigmastate/verification/contract/CrowdFundingContractVerification.scala#L34-L76)
+- [ ] find and prove more properties.
+
+## ICO contract
+### Rewrite ICO contracts:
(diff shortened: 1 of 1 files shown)


docs/proving-contract-properties.md
```diff
@@ -0,0 +1,26 @@
...
+# Workflow:
+Write the contract in SigmaDls.
+During the compilation:
+* Our macros generate ErgoTree;
+* [Stainless](https://github.com/epfl-lara/stainless) verifies contract properties;
+
+# Plan:
+
+## Crowdfunding Contract:
+- [x] some properties of the Crowdfunding contract are already proven, see [source](http://github.com/ScorexFoundation/sigmastate-interpreter/blob/1f65d90cdab0388acc73bfa43797e39d3d9e32bc/contract-verification/src/main/scala/sigmastate/verification/contract/CrowdFundingContractVerification.scala#L34-L76)
```
aslesarenko: This is really nice. I wonder if we can formulate and verify mutual exclusion of contract preconditions.

greenhat: There is only one `require` allowed. So all preconditions should be together.
I'm not sure I get the idea of what do you want to verify here. I've used preconditions for:
1. to set a list of requirements for the input in the contract itself;
2. to "limit" the inputs in proofs (see `prove*` methods);

docs/proving-contract-properties.md
```diff
@@ -0,0 +1,100 @@
...
+
+Such defined ErgoScala allows us to meet the [goals](#goals) by:
+1) writing, executing, testing and debugging Ergo contracts directly in Scala using existing 
+tools, libraries and IDE support
+2) implementing Scala macros to compile ErgoScala fragments (contracts) into [ErgoTree](ref) 
+the intermediate representation of Ergo contracts suitable for storage in Ergo blockchain. (G4)
+3) using [Stainless](https://github.com/epfl-lara/stainless) verification framework to specify pre(post)-conditions for Ergo contracts and formally prove them using auxiliary lemmas. (PureScala is a superset of ErgoScala, 
+thus any ErgoScala fragment is verifiable as long as it is verifiable as PureScala fragment
+in Stainless) (G1, G2, G3)
+4) check the declared pre(post)-conditions using ScalaCheck property testing (G5)
```
greenhat: @aslesarenko How to check a pre-condition itself (e.g. `HEIGHT > 0`)? Do you mean to check on the contract that the post-conditions hold?

aslesarenko: @greenhat Yes, check post-conditions given pre-conditions. The pre-conditions should be checked as part of contract usage (i.e as part of transaction creation)

greenhat: @aslesarenko It's Stainless job to test if post-conditions hold. What's the point to test them with property testing(ScalaCheck)?

PR #631, opened 2020-01-08
Rename ErgoTree.root to template

greenhat:
Following the discussion in https://github.com/ergoplatform/explorer-back/pull/230#discussion_r363848585


aslesarenko:
once EIP5 is implemented `template` will have different meaning, so this rename is not needed.

PR #648, opened 2020-04-03
Checking content length in P2SH, examples fixed

kushti:
This PR contains:

* check that P2SH content bytes (192-bits hash of the corresponding script) are of proper length
* fixing examples for P2SH / P2S addresses in comments

### DIFF
--- a/sigmastate/src/main/scala/org/ergoplatform/ErgoAddress.scala
+++ b/sigmastate/src/main/scala/org/ergoplatform/ErgoAddress.scala
@@ -246,11 +246,11 @@ case class ErgoAddressEncoder(networkPrefix: NetworkPrefix) {
   object IsPay2SHAddress {
     def unapply(exp: SigmaPropValue): Option[Coll[Byte]] = exp match {
       case SigmaAnd(Seq(
-      BoolToSigmaProp(
-      EQ(
-      Slice(_: CalcHash, ConstantNode(0, SInt), ConstantNode(24, SInt)),
-      ByteArrayConstant(scriptHash))),
-      DeserializeContext(Pay2SHAddress.scriptId, SSigmaProp))) => Some(scriptHash)
+            BoolToSigmaProp(
+              EQ(
+                Slice(_: CalcHash, ConstantNode(0, SInt), ConstantNode(24, SInt)),
+                ByteArrayConstant(scriptHash))),
+            DeserializeContext(Pay2SHAddress.scriptId, SSigmaProp))) => Some(scriptHash)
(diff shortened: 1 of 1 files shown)


sigmastate/src/main/scala/org/ergoplatform/ErgoAddress.scala
```diff
@@ -226,24 +230,27 @@ case class ErgoAddressEncoder(networkPrefix: NetworkPrefix) {
...
-            BoolToSigmaProp(
-              EQ(
-                Slice(_: CalcHash, ConstantNode(0, SInt), ConstantNode(24, SInt)),
-                ByteArrayConstant(scriptHash))),
-            DeserializeContext(Pay2SHAddress.scriptId, SSigmaProp))) => Some(scriptHash)
+      BoolToSigmaProp(
+      EQ(
+      Slice(_: CalcHash, ConstantNode(0, SInt), ConstantNode(24, SInt)),
+      ByteArrayConstant(scriptHash))),
+      DeserializeContext(Pay2SHAddress.scriptId, SSigmaProp))) => Some(scriptHash)
```
aslesarenko, line 253: The original code is better formatted, more readable.

kushti: @aslesarenko reverted

sigmastate/src/main/scala/org/ergoplatform/ErgoAddress.scala
```diff
@@ -226,12 +230,15 @@ case class ErgoAddressEncoder(networkPrefix: NetworkPrefix) {
           val p = GroupElementSerializer.parse(r)
           new P2PKAddress(ProveDlog(p), contentBytes)
         case Pay2SHAddress.addressTypePrefix =>
+          if (contentBytes.length != 24) { //192-bits hash used
+            throw new Exception(s"Improper content in P2SH script: $addrStr")
+          }
           new Pay2SHAddress(contentBytes)
```
aslesarenko, line 236: @kushti, @greenhat  I'm not sure we don't have potential problem with HF here. Are we sure this method is not used during contract verification? Because otherwise we are changing  the semantics of scripts (some of them may start failing).

kushti: @aslesarenko addresses are not known on the protocol (ErgoTree) level, so I do not see any problems. Also, 192-bits input was expected in the P2SH script, now the corresponding check added to the address instantiation logic

PR #650, opened 2020-04-14
add ErgoTree.template

greenhat:


sigmastate/src/main/scala/sigmastate/Values.scala
```diff
@@ -954,6 +954,12 @@ object Values {
       _complexity
     }
 
+    /** Serialized tree with constant placeholders */
```
aslesarenko, line 957: Please describe it as property (i.e. what it returns) instead of as method (i.e. what it does).
Give more details, how it is different from `bytes`.

greenhat: Done.

PR #652, opened 2020-04-29
Cannot resolve companion instance for s493 -> IntPlusMonoidCtor(s18)

greenhat:
Ref issue in ergo - https://github.com/ergoplatform/ergo/issues/1095


greenhat:
It fails on new IR only. Temporarily I moved it to be the first in the suite to make sure it fails on CI.

PR #666, opened 2020-06-15
[refer #1132] Change ModifierId representation

knizhnik:


aslesarenko:
LGTM, but please update to the latest version of scorex-util

PR #669, opened 2020-06-25
Cross version testing support

aslesarenko:
This PR introduces support for cross version testing for Sigma operations.


sigmastate/src/main/scala/sigmastate/eval/TreeBuilding.scala
```diff
@@ -380,6 +380,9 @@ trait TreeBuilding extends RuntimeCosting { IR: IRContext =>
       case GM.exp(In(obj), In(arg)) =>
         mkExponentiate(obj.asGroupElement, arg.asBigInt)
 
+      case GM.multiply(In(obj), In(arg)) =>
```
kushti, line 383: How's that is not a consensus-breaking change?

aslesarenko: TreeBuilding is part of front-end compiler and is used to produce ErgoTree from ErgoScript. 
I was pretty sure about it, but checked again, it is never used as part of `verify` and `prove`.
This fixes a bug I found in front-end compiler.

PR #680, opened 2020-08-25
Test vectors for cost estimation

aslesarenko:


sigmastate/src/main/scala/org/ergoplatform/ErgoBox.scala
```diff
@@ -77,6 +77,13 @@ class ErgoBox(
     */
   lazy val bytes: Array[Byte] = ErgoBox.sigmaSerializer.toBytes(this)
 
+  /** Creates a new box by updating some of the additional registers with the given new bindings.
+    * @param newBindings a map of the registers to be updated with new values
+    */
+  def withUpdatedRegisters(newBindings: AdditionalRegisters): ErgoBox = {
```
kushti, line 83: Public API should literally enforce proper usage of it. This method is going in opposite direction, by returning incorrect box (with improper reference). As it is used only once in the testing code, maybe it is better to eliminate the method at all.

sigmastate/src/main/scala/org/ergoplatform/ErgoBox.scala
```diff
@@ -175,7 +184,7 @@ object ErgoBox {
 
   val allZerosModifierId: ModifierId = Array.fill[Byte](32)(0.toByte).toModifierId
 
-  def apply(value: Long,
+  def create(value: Long,
```
kushti, line 187: Similarly to previous comment , consider to rename the method to testBox

sigmastate/src/test/scala/sigmastate/helpers/SigmaTestingCommons.scala
```diff
@@ -224,15 +225,23 @@ trait SigmaTestingCommons extends PropSpec
...
               .withBindings(1.toByte -> Constant[SType](x.asInstanceOf[SType#WrappedType], tpeA))
               .withBindings(bindings: _*)
-          ergoCtx.toSigmaContext(IR, isCost = false)
+          val calcCtx = ergoCtx.toSigmaContext(IR, isCost = false).asInstanceOf[CostingDataContext]
+          val costCtx = calcCtx.copy(isCost = true)
+          (costCtx, calcCtx)
       }
+
+      val estimatedCost = IR.checkCostWithContext(costingCtx, costF, ScriptCostLimit.value, 0L).getOrThrow
+//      println(s"Estimated Cost: $estimatedCost")
```
kushti, line 241: Commented out code

kushti (changes requested):

sigmastate/src/main/scala/org/ergoplatform/ErgoBox.scala
aslesarenko: Why "improper reference"?

sigmastate/src/main/scala/org/ergoplatform/ErgoBox.scala
aslesarenko: I replaced implicit `apply` with explicit `create` factory method to make more clear where the default constructor is used and where is not. 
Other changes is beyond the scope of this PR, which is adding test vectors.
Added TODO refactor.

sigmastate/src/test/scala/sigmastate/helpers/SigmaTestingCommons.scala
aslesarenko: removed

sigmastate/src/main/scala/org/ergoplatform/ErgoBox.scala
kushti: @aslesarenko transaction bytes (means reference field) would be changed after adding anything to a box created by the transaction.

sigmastate/src/main/scala/org/ergoplatform/ErgoBox.scala
kushti: @aslesarenko "create" is not addressing context where this method should be used (testing only)

Also, better not to make ErgoBox case class, who knows if any descendant will appear for which it will be more convenient to use case-class form.

sigmastate/src/main/scala/org/ergoplatform/ErgoBox.scala
aslesarenko: ok, removed.

sigmastate/src/main/scala/org/ergoplatform/ErgoBox.scala
aslesarenko: moved this method to tests

kushti (approved):

PR #685, opened 2020-08-31
signMessage / verifyMessage for signing and verifying arbitrary messages

kushti:
In this PR discussed for a long time idea to use Generalized Schorr Signatures (so non-interactive version of Generalized Schnorr Protocols) has been implemented.

With this PR being merged, a prover (or provers) can verify their identities by using an arbitrary message (i.e. not necessarily a transaction).
See e.g. https://www.ergoforum.org/t/ergotree-as-an-authentication-language/274.


greenhat (approved):

sigmastate/src/test/scala/sigmastate/crypto/SigningSpecification.scala
```diff
@@ -45,6 +47,42 @@ class SigningSpecification extends SigmaTestingCommons {
...
   }
 
+  property("signMessage / verifyMessage roundtrip - simple dlog") {
+    forAll(Gen.alphaNumStr){str =>
+      val msg = str.getBytes("UTF-8")
+      val pi = new ErgoLikeTestProvingInterpreter()
+      val sigmaTree: SigmaBoolean = pi.publicKeys.head
+      val sig = pi.signMessage(sigmaTree, msg, HintsBag.empty).get
+      pi.verifySignature(sigmaTree, msg, sig) shouldBe true
+      pi.verifySignature(sigmaTree, (str + "1").getBytes("UTF-8"), sig) shouldBe false
```
aslesarenko, line 57: Please add negative tests with invalid pk, and sig

sigmastate/src/test/scala/sigmastate/crypto/SigningSpecification.scala
```diff
@@ -45,6 +47,42 @@ class SigningSpecification extends SigmaTestingCommons {
...
+  }
+
+  property("signMessage / verifyMessage roundtrip - complex key") {
+    forAll(Gen.alphaNumStr){str =>
+      val msg = str.getBytes("UTF-8")
+      val pi = new ErgoLikeTestProvingInterpreter()
+      val sigmaTree: SigmaBoolean = CAND(Seq(pi.dlogSecrets.head.publicImage, pi.dhSecrets.head.publicImage))
+      val sig = pi.signMessage(sigmaTree, msg, HintsBag.empty).get
+      pi.verifySignature(sigmaTree, msg, sig) shouldBe true
+      pi.verifySignature(sigmaTree, (str + "1").getBytes("UTF-8"), sig) shouldBe false
```
aslesarenko, line 68: Same as above, more negative checks.

sigmastate/src/test/scala/sigmastate/utxo/DistributedSigSpecification.scala
```diff
@@ -481,4 +486,32 @@ class DistributedSigSpecification extends SigmaTestingCommons {
...
+      .addHint(hintsFromBob.ownCommitments.head)
+
+    val sigBob = proverB.signMessage(sigmaTree, msg, bagB).get
+
+    // Proof generated by Alice without getting Bob's part is not correct
+    verifier.verifySignature(sigmaTree, msg, sigAlice) shouldBe false
+
+    // Compound proof from Bob is correct
+    verifier.verifySignature(sigmaTree, msg, sigBob) shouldBe true
+  }
```
aslesarenko, line 516: Looks like all the scenarios can be generalized to test both signing, and old-style prove/verify. It seams the data flow is almost the same. 
So I suggest to add verifySignature calls into all the existing cases.

aslesarenko (changes requested):

sigmastate/src/test/scala/sigmastate/crypto/SigningSpecification.scala
kushti: Done

sigmastate/src/test/scala/sigmastate/crypto/SigningSpecification.scala
kushti: Done

sigmastate/src/test/scala/sigmastate/utxo/DistributedSigSpecification.scala
kushti: verifySignature is already inside verify(), for transaction bytes as a message.

To bring another (custom) message into all the tests, we need to separate data and execution via an interpreter pattern (e.g. Free monad), but this will reduce readability , and an advantage is unclear.

kushti:
@aslesarenko comments addressed , please make another round!

sigmastate/src/test/scala/sigmastate/utxo/DistributedSigSpecification.scala
aslesarenko: Oh, no Free Monad please.

sigmastate/src/main/scala/sigmastate/SigSerializer.scala
```diff
@@ -142,10 +141,15 @@ object SigSerializer {
...
-    // Verifier step 1: Read the root challenge from the proof.
-      traverseNode(exp, bytes, 0, challengeOpt = None)._1 // get the root hash, then call
+    } else {
+      // Verifier step 1: Read the root challenge from the proof.
+      val (res, finalPos) = traverseNode(exp, bytes, pos = 0, challengeOpt = None) // get the root hash, then call
+      if (finalPos != bytes.length) {
+        throw new Exception("Signature contains extra bytes")
+      }
+      res
+    }
```
aslesarenko, line 153: This is executed as part of Prover.verify() and this introduces a new more stringent requirement on bytes which may break consensus. 
If there is a chance the old nodes will pass extra bytes, but the `res` will otherwise correct. 
I suggest to remove this `if` and leave `// TODO HF: ...` instead.

sigmastate/src/test/scala/sigmastate/crypto/SigningSpecification.scala
aslesarenko: Please finish test cases :)

aslesarenko (changes requested):

sigmastate/src/main/scala/sigmastate/SigSerializer.scala
kushti: reverted, I guess this check could be avoided at all

sigmastate/src/test/scala/sigmastate/crypto/SigningSpecification.scala
kushti: done more tests

kushti:
@aslesarenko comments addressed , please make another round!

aslesarenko (approved):

PR #689, opened 2020-09-18
Avoid ClassCastException in kiama rewritings

aslesarenko:
Most of the kiama rewriting rules were throwing exceptions for some tree nodes. 
This is interpreted by kiama as the rule is not applicable, so the node is preserved and is not rewritten.

This PR changes the rules, so that they have the same effect, but without exception.


aslesarenko:
Subsumed by #690

PR #695, opened 2020-10-16
The Protocol and Requirements to replace AOT with JIT via soft-fork

aslesarenko:


docs/aot-jit-switch.md
```diff
@@ -0,0 +1,184 @@
...
+performed by a v5.0 node. Any such a pair `(i, i+1)` of rules have `Validation Actions`
+which are either the same or equivalent with respect to the [Equivalence Properties of
+Validation Actions](#equivalence-properties-of-validation-actions). This proves
+consistency of validation actions across v4.0 and v5.0 nodes.
+
+5. After SF is activated (`SF Status == active`), both AOT-cost and AOT-verify
+implementations are no longer used in `Validation Action` of v5.0 nodes. The only context
+where v5.0 node needs to use AOT based verification is given by Rule 6, which is to verify 
+a v1 script in a historical mined block before SF is activated. 
+However relying on _Prop 3_ we can replace Rule 6 in a new v5.0.1 release with the
```
greenhat, line 85:
```suggestion
However relying on _Prop3_ we can replace Rule 6 in a new v5.0.1 release with the
```

greenhat (approved):

docs/aot-jit-switch.md
```diff
@@ -0,0 +1,184 @@
...
+Validation of scripts in blocks is defined for each release and depend on _validation
+context_ which includes type of block, soft-fork status and script version. We denote
+blocks being created by miners as `candidate` and those distributed across network as
+`mined`.
+
+Thus, we have 8 different validation contexts multiplied by 2 node versions
+having in total 16 validation rules as summarized in the following table, which 
+specifies the action a node have to take in the given contexts.
+
+Rule#| SF Status| Block Type| Script Version | Release | Validation Action
```
kushti, line 37: Validation Action is not defined.

docs/aot-jit-switch.md
```diff
@@ -0,0 +1,184 @@
...
+12   | active   | candidate | Script v2 | v5.0 | R5.0-JIT-verify 
+||||
+13   | active   | mined     | Script v1 | v4.0 | skip-accept (rely on majority)
+14   | active   | mined     | Script v1 | v5.0 | R5.0-JIT-verify 
+15   | active   | mined     | Script v2 | v4.0 | skip-accept (rely on majority)
+16   | active   | mined     | Script v2 | v5.0 | R5.0-JIT-verify 
+
+Note the following properties of the validation rules.
+
+1. Rules 1-4 specify creation of new candidate blocks _before_ soft-fork is activated.
```
kushti, line 61: Please note that block creation is not a part of the protocol, and miners can do whatever they want here. Even more, possibly some miners are doing custom block assembly right now. 

This it would be reasonable to split consensus and non-consensus procedures, by e.g. moving block assembly rules into an appendix.

docs/aot-jit-switch.md
kushti: Prop 3 is not defined before.

docs/aot-jit-switch.md
```diff
@@ -0,0 +1,184 @@
...
+ _Handle v1 script in mined block when SF is not active._
+ Similar to rules 1 and 2 but for `mined` blocks. 
+ 
+#### Rules 7 and 8
+_Both v4.0 and v5.0 nodes reject v2 scripts in `mined` blocks when SF is NOT active._ 
+Similar to rules 3 and 4.
+
+#### Rules 9 and 10
+These rules allow v1 scripts to enter blockchain even after SF is activated (for backward
+compatibility with Apps).
```
kushti, line 122: Apps not defined

docs/aot-jit-switch.md
```diff
@@ -0,0 +1,184 @@
...
+
+_Prop 4._ JIT-cost is bound by AOT-cost: `forall s:ScriptV1, R5.0-JIT-cost(s) <= R4.0-AOT-cost(s)`
+
+_Prop 5._ ScriptV2 is rejected before SF is active:  
+`forall s:ScriptV2, if not SF is active => R4.0-verify(s) == R5.0-verify(s) == Reject`
+
+### Other Notes
+
+- Since we are going to fix some bugs, the behaviour of v1 and v2 scripts in general not
+required to be precisely equivalent. So while old apps are supported unchanged, new apps
```
kushti, line 180: Sounds contradicting to Prop. 3.

kushti (changes requested):

kushti:
@aslesarenko I think in regards with consensus rules we're on the same page. However, the proposal has bad writing or uncertainties in some places.

docs/aot-jit-switch.md
aslesarenko: Yes, but on the other hand it may be easier to look at the whole set of rules in one table. I will move you comment in the test and make this point more clear.

docs/aot-jit-switch.md
aslesarenko: Changed it into url which refers to the definition.

docs/aot-jit-switch.md
aslesarenko: added definition

docs/aot-jit-switch.md
aslesarenko: changed "Apps" to "applications"

docs/aot-jit-switch.md
aslesarenko: Added the following clarification:

This is because `R5.0-JIT-verify` supports both v1
and v2 scripts so that `R5.0-JIT-verify(Script_v1) == R4.0-AOT-verify(Script_v1)` due to
[Prop 1](#equivalence-properties-of-validation-actions) and `R5.0-JIT-verify(Script_v2)`
may be completely different language. Of cause, we don't want it to be _completely_
different, in particular to ease migration.

aslesarenko:
@kushti The goal of the document is to specify a complete set of rules unambiguously, so that both v4.0 and v5.0 interpreters can implement them. 
If you see concrete changes which can improve the text without complete rewrite, please propose.

kushti (approved):

PR #697, opened 2020-10-28
Ensure v4.0 can support AOT -> JIT switch as v5.0 soft-fork (includes HF changes)

aslesarenko:
Closes #637 
Introduces hard-forking features necessary for AOT -> JIT switch, as a result, this PR should be merged into 
develop only as part of v4.0-RC .

PR implements [validation checks](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/develop/docs/aot-jit-switch.md) performed by v4.0 release (see specification test cases).
The new checks are designed to support future v5.0 soft-fork.

PR #639, opened 2020-01-31
New evaluation and costing

aslesarenko:
> This new implementation of ErgoTree v2 is based on #637,  #697 
The code of this PR can be merged into v3.x/v4.0 branches, BUT NOT ACTIVATED. 
The actual activation should be done in v5.0 release as it is described in https://github.com/ScorexFoundation/sigmastate-interpreter/blob/develop/docs/aot-jit-switch.md.
That said, this PR MUST NOT introduce consensus breaking changes in the existing (AOT based) implementation of the interpreter.

### Motivation
ErgoTree is simple declarative intermediate representation for Ergo contracts.
It is designed to be compact in serialized form and directly executable, i.e. no additional transformation is necessary before it can be efficiently executed. 

This PR implements big-step recursive ErgoTreeEvaluator that works directly with ErgoTree.
Because of this the evaluator is very simple and follows [denotational semantics of ErgoTree](https://ergoplatform.org/docs/ErgoTree.pdf).
Or, the other way around, this implementation of ErgoTreeEvaluator is purely functional with immutable data structures and can be used as definition of ErgoTree's semantics.

### Implementation
ErgoTreeEvaluator takes ErgoTree directly as it is deserialized as part of a transaction.
No additional transformations is performed.
ErgoTree is interpreted directly and all intermediate data are stored in runtime types.
Runtime types are such types as Coll, SigmaProp, AvlTree, BigInt, etc.
It also use immutable Map to keep current DataEnv of computed ValDefs, as a result only the addition operation is used from the map, and the deletion is essentially done by the garbage collection.

### Performance
Since this interpreter directly works with SigmaDsl types (Coll, BigInt, SigmaProp etc), it turns out to be fast.
Since it also does JIT style costing instead of AOT style, it is 5-6x faster then existing implementation.

### Milestone
New evaluation of ErgoTree required changes in how isProven handled.
In particular isProven is not supported in ErgoTree that is passed to prove/verify methods.
This required changes in tests and also in the coverage report.
Unfortunately those changes may break the consensus, because isProven can be serialized and present in ErgoTree.
This prevents ErgoTreeEvaluator to be used in v3.x

Issue #637, opened 2020-01-27
Make sure v5.0 SF will be feasible


aslesarenko:
This tasks should be done in v4.0 in order to make v5.0 via SF possible:
- [x] set size bit to 1 in ErgoTree v2 (version bits == 1) in order to allow lazy deserialization, which will allow to solve #694.
- [x] new validation rule to check version and size Bit in ErgoTree

See also:
1) https://github.com/ScorexFoundation/sigmastate-interpreter/issues/660
2) [motivation, algorithm and analysis](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/master/sigmastate/src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala#L27)

aslesarenko:
Implemented in #697

PR #699, opened 2020-11-06
New direct ErgoTree evaluation and JIT costing

aslesarenko:
> This new implementation of ErgoTree v2 is based on #639  #637,  #697 
The code of this PR can be merged into v3.x/v4.0 branches, BUT NOT ACTIVATED. 
The actual activation should be done in v5.0 release as it is described in https://github.com/ScorexFoundation/sigmastate-interpreter/blob/develop/docs/aot-jit-switch.md.
That said, this PR MUST NOT introduce consensus breaking changes in the existing (AOT based) implementation of the interpreter.
### Roadmap
- [x] all ErgoTree nodes are supported by new ErgoTreeEvaluator
- [x] JIT cost accumulation is implemented for all ErgoTree nodes (fullReduction part)
- [x] new AOT cost estimation is implemented for SigmaBoolean (sigma proposition verification part)
- [x] explicit type checks during ErgoTree evaluation 
- [x] v5.0 JIT-based costing withstands the spam tests
- [x] update EIP-0008: New Transaction Validation with Just-In-Time Costing
    - [x] describe new JIT costing algorithm specification
- [x] new optimized SigSerializer with cost profiling 
    - [x] optimizations and profiling is implemented
    - [x] PR #719 is merged into v4.x
- [x] all `TODO v5.0` in the code are resolved
- [x] cost parameters are finalized based on profiling and Ergo node metrics analysis
    - [x] Jupyter Notebook with Analysis Report is created in https://github.com/ergoplatform/ergo/pull/1330
- [x] this PR is minimized by moving non-breaking changes into v4.x PRs
    - [x] part 1 #720
    - [x] part 2 #726 
- [ ] test vectors for new cost parameters added to tests
- [ ] activation protocol is implemented

### Motivation
ErgoTree is simple declarative intermediate representation for Ergo contracts.
It is designed to be compact in serialized form and directly executable, i.e. no additional transformation is necessary before it can be efficiently executed. 

This PR implements big-step recursive ErgoTreeEvaluator that works directly with ErgoTree.
Because of this the evaluator is very simple and follows [denotational semantics of ErgoTree](https://ergoplatform.org/docs/ErgoTree.pdf).
Or, the other way around, this implementation of ErgoTreeEvaluator is purely functional with immutable data structures and can be used as definition of ErgoTree's semantics.

### Implementation
ErgoTreeEvaluator takes ErgoTree directly as it is deserialized as part of a transaction.
No additional transformations is performed.
ErgoTree is interpreted directly and all intermediate data are stored in runtime types.
Runtime types are such types as Coll, SigmaProp, AvlTree, BigInt, etc.
It also use immutable Map to keep current DataEnv of computed ValDefs, as a result only the addition operation is used from the map, and the deletion is essentially done by the garbage collection.

### Performance
Since this interpreter directly works with SigmaDsl types (Coll, BigInt, SigmaProp etc), it turns out to be fast.
Since it also does JIT style costing instead of AOT style, it is 5-6x faster then existing implementation.

### Milestone
New evaluation of ErgoTree required changes in how isProven handled.
In particular isProven is not supported in ErgoTree that is passed to prove/verify methods.
This required changes in tests and also in the coverage report.
Unfortunately those changes may break the consensus, because isProven can be serialized and present in ErgoTree.
This prevents ErgoTreeEvaluator to be used in v3.x


aslesarenko:
Subsummed by #758

PR #707, opened 2020-12-01
Test vectors for comparison operations

aslesarenko:
- Avoid compile-time warnings by declaring used language features like `import scala.language.implicitConversions`
- added more TODO items found after more tests were added
- simple front-end only optimization to avoid producing block sub-expressions like `{ val idNew = id; idNew }`
- added more ScalaDocs
- added property("MethodCall on numerics")
- pretty printing of ErgoTree when sanity check fails
- optimized generation of test data (see class `Sampled[A]`) 
- new test vectors and test cases for numeric comparison operations (see SigmaDslSpecification.scala)


kushti (changes requested):
This PR is "Test vectors for comparison operations", however, it contains much more than tests, also, no description is provided.

I propose to do following:

1) Split this PR into 3 PRs, one with scala features imports, another with TreeBuilding.processAstGraph optimizations, third with tests
2) Provide descriptions for all the PRs. For TreeBuilding.processAstGraph optimizations, describe why there's no consensus issue here.

aslesarenko:
@kushti While I agree with rationale behind splitting this PR, I suggest to not waste time and to focus on more important parts.  I would better add descriptions in the code itself, rather than PR descriptions.

TreeBuilding is not related to consensus, will also comment the changed code with detailed description.

kushti:
> @kushti While I agree with rationale behind splitting this PR, I suggest to not waste time and to focus on more important parts. I would better add descriptions in the code itself, rather than PR descriptions.
> 
> TreeBuilding is not related to consensus, will also comment the changed code with detailed description.

Proper PR description and title is crucial part of development process. Development progress should be observable without digging in the codebase.

aslesarenko:
@kushti I don't think it is worth the efforts

kushti (approved):

kushti:
@aslesarenko LGTM, but please remove unused imports from SigmaTypeGens and printlns (and unused import) from SigmaTestingCommons

PR #718, opened 2021-03-17
More tests for prove/verify

aslesarenko:
In this PR:
- [x] tests for sign/verify completeness (generateProof/verifySignature)
- [x] test vectors for SigSerializer
- [x] test vectors for FiatShamirTree.toBytes


sigmastate/src/main/scala/sigmastate/SigSerializer.scala
```diff
@@ -137,7 +137,7 @@ object SigSerializer {
           }
 
           // Verifier doesn't need the polynomial anymore -- hence pass in None
```
kushti, line 139: Comment is outdated, "tests" kind of PR contains undocumented changes in logic.

kushti (changes requested):

sigmastate/src/main/scala/sigmastate/SigSerializer.scala
aslesarenko: Please be more specific, so I can document what is missing.

kushti: @aslesarenko code change under this comment is not documented

aslesarenko: @kushti  The decision to use None was a mistake in a first place, which was not found earlier because the tests where not exhaustive.
Now it is found and fixed, so the code looks like it is expected.
I'm not sure which comment to put in the code. 
I can put this comment in the code, but it will look awkward.

kushti: but comment still says "// Verifier doesn't need the polynomial anymore -- hence pass in None"

kushti (approved):

Issue #660, opened 2020-06-08
[planning] Hard-fork Activation Support  


aslesarenko:
### Problem

The hard-fork activation will be implemented according to [EIP-9](https://github.com/ergoplatform/eips/pull/12).
To support activation protocol, sigma code should support both ErgoTree v0 and v1 versions at the same time (even in the same transaction). This means we cannot accumulate changes in a separate branch (as we normally do) and then release them all at once (which would replace v0 version with v1 version implementation).

Support of both versions simultaneously requires a discipline in how the protocol related changes are introduced into the code.

### Solution
Roughly, the idea is that the code changes should be protected by version checking conditions like the following:
```scala
if (currVersion == v1) {
  // new code
} else if (currVersion == v0) {
  // old code
} else {
  // unsupported version
}
```
This will makes the code:
1) support both the old ErgoTree v0 and the new ErgoTree v1 and 
2) switch the version conditionally based on the value of currVersion. 

Existing [v4.0 branch](https://github.com/ScorexFoundation/sigmastate-interpreter/tree/v4.0) has already accumulated many protocol related changes, however those are without version checking conditions.

aslesarenko:
It seems we will have a lot of places with versioned code for 4.0. 

The easiest way I see to implement it is to pass `VersionContext` (or something like this) to all those places using `DynamicVariable`, which is based on `java.lang.InheritableThreadLocal` (since JDK 1.2 for passing UserID, TransactionID kind of things). 
What is required is that `verifier.verify(...)` is called under correctly constructed `VersionContext` (which is part of HF Activation Protocol).

That said, I don't expect problems with versioned implementation of anything in Sigma, as long as `verify` is called under correct `VersionContext`.

It can then be used in a _versioned code fragment_ like the following:
```scala
currentVersionContext match {
  case Version(3, x) | Version(4, 0) => 
    //  code of version 3.x where x is bound to the actual value or version 4.0
    // existing code goes here
  case Version(4, x) =>
   // this case of version 4.x code (can be added later in a soft-fork release) where x > 0
  case _ => 
    // unknown version is interpreted as soft-fork condition 
    // (i.e. this is the old code which is executed in the network after new version was activated)
    // by throwing this exception we signal the validator, to accept the script without validation
    throw new ValidationException(versionedFragmentId)
}
```

[...]

PR #748, opened 2021-11-17
Activation of v5.0

### LINKED ISSUE #747: 5.0 activation
Code for 5.0 activation for old nodes already in the interpreter:

```
if (context.activatedScriptVersion > Interpreter.MaxSupportedScriptVersion) {
        // > 90% has already switched to higher version, accept without verification
        // NOTE: this path should never be taken for validation of candidate blocks
        // in which case Ergo node should always pass Interpreter.MaxSupportedScriptVersion
        // as the value of ErgoLikeContext.activatedScriptVersion.
        // see also ErgoLikeContext ScalaDoc.
        return Success(true -> context.initCost)
      }
```
so old nodes simply skipping scripts check (downgrading to SPV security effectively)
For new (5.0) nodes, switch to JIT costing should be implemented when MaxSupportedScriptVersion increased.

For upcoming soft-forks, only adding new opcodes and types planned, thus existing soft-forking protocol would be enough?

aslesarenko:
Closes #747, #660 
- includes #753 
- introduced  context.currentErgoTreeVersion


sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -266,23 +255,26 @@ trait Interpreter extends ScorexLogging {
     *         after reduction
     */
   def fullReduction(ergoTree: ErgoTree,
-                    context: CTX,
+                    ctx: CTX,
```
kushti, line 258: ScalaDoc not fixed

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -315,11 +307,11 @@ trait Interpreter extends ScorexLogging {
   /** Performs reduction of proposition which contains deserialization operations. */
   private def reductionWithDeserialize(ergoTree: ErgoTree,
                                        prop: SigmaPropValue,
-                                       context: CTX,
-                                       env: ScriptEnv): (ReductionResult, JitReductionResult) = {
+                                       context: CTX, env: ScriptEnv,
```
kushti, line 310: weird formatting

kushti (changes requested):
Please address comments, also 

```
        if (ergoTree.version > Interpreter.MaxSupportedScriptVersion) {
          // We accept the box spending and rely on 90% of all the other nodes.
          // Thus, the old node will stay in sync with the network.
          return Success(true -> context.initCost)
        }
```
why this rule for the next version ? It is planned to be about introducing new and abandoning some old operations, thus we can avoid downgrading to SPV for  old nodes ?

aslesarenko:
> why this rule for the next version ?
the `return Success(true -> context.initCost)` case will work on v5.x node, then v6.0 (Block v4 and above) protocol is activated and we have nothing else to do.

In all the other cases we proceed normally assigning ErgoTree.version to the context.ergoTreeVersion. This is done in fullReduction.

When `ergoTree.version > context.activatedScriptVersion` then this is not yet supported version of ErgoTree.

sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
```diff
@@ -111,6 +111,11 @@ class ErgoLikeContext(val lastBlockUtxoRoot: AvlTreeData,
   */
   val self: ErgoBox = boxesToSpend(selfIndex)
 
+  val currentErgoTreeVersion: Option[Byte] = None
```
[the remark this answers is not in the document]
aslesarenko: fixed.

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
aslesarenko: done

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
aslesarenko: fixed

sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
```diff
@@ -111,6 +111,18 @@ class ErgoLikeContext(val lastBlockUtxoRoot: AvlTreeData,
   */
   val self: ErgoBox = boxesToSpend(selfIndex)
 
+  /** Current version of the ErgoTree executed by the interpreter.
+    * This property is used to implement version dependent operations and passed to
+    * interpreter via [[specia.sigma.Context]].
+    * The value cannot be assigned on [[ErgoLikeContext]] construction and must be
```
kushti, line 117: Why can't the property be passed via constructor. 

Initialization order is error-prone (and doing 2x allocations) , as clients getting the new version of the code will get errors in runtime only.

sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
```diff
@@ -111,6 +111,18 @@ class ErgoLikeContext(val lastBlockUtxoRoot: AvlTreeData,
...
+    * attached using [[withErgoTreeVersion()]] method.
+    * When the value is None, the [[InterpreterException]] is thrown by the interpreter.
+    */
+  val currentErgoTreeVersion: Option[Byte] = None
```
kushti, line 121: Naming not ideal, as "current" does not say much, should it be something like "executedErgoTreeVersion" ?

kushti (changes requested):

sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
aslesarenko: "executed" is only creates an illusion of better description. Which execution? What are the assumptions? 
In such cases is better to read the docs and "current", by being general enough, encourage the reader to consult the docs.

sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
aslesarenko: 1) The overhead of allocation is negligible since it only 1 allocation per script and we don't have even thousands of scripts per second (it would only make sense for millions per second)
For comparison, every ErgoTree operation in v5.0 interpreter creates a new object (because of immutability), so at least 1 allocation per operation. 
2) This context parameter is only necessary inside interpreter and can be safely initialized based on the actually passed ErgoTree. So this is encapsulated into Interpreter and the client don't have to deal with this. (the motivation is similar to evaluationMode we discussed before an you were in favor of encapsulation).
By default it is None and the CostingDataContext, which is created for each ErgoLikeContext and actually used in the reduction have this  `currentErgoTreeVersion: Byte` i.e. it is required.
There is no way to execute the interpreter without setting this parameter.

3) If make it a context parameter it will require a lot of changes everywhere and each client will need to take care of passing the correct version for the given ErgoTree. In particular need to know how to extract that version.

kushti: In Ergo node, there are few instantiations of ErgoLikeContext, and also 

```
class ErgoContext(val stateContext: ErgoStateContext,
                  transactionContext: TransactionContext,
                  inputContext: InputContext,
                  override val costLimit: Long,
                  override val initCost: Long)
  extends ErgoLikeContext(ErgoInterpreter.avlTreeFromDigest(stateContext.previousStateDigest),
    stateContext.sigmaLastHeaders,
    stateContext.sigmaPreHeader,
    transactionContext.dataBoxes,
    transactionContext.boxesToSpend,
    transactionContext.spendingTransaction,
    inputContext.selfIndex,
    inputContext.extension,
    stateContext.validationSettings.sigmaSettings,
    costLimit,
    initCost,
    activatedScriptVersion = (stateContext.blockVersion - 1).toByte // block version N of ErgoProtocol corresponds to version N-1 of ErgoTree (aka script version)
  )
``` 

not sure about other applications.

ErgoContext will implicitly get the iini problems, for example, just with dependency update. And the problem will appear in runtime only. That's not how a library should be developed.

kushti (approved):

PR #753, opened 2021-11-28
Encapsulate evaluationMode selection (towards v5.0)

### LINKED ISSUE #747: 5.0 activation

aslesarenko:
Closes #747 

- [x] Evaluation mode encapsulated
- [x] Soft-fork activation tests updated
- [x] Tests cases for v6.0 activation added

### DIFF
--- a/sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
+++ b/sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
@@ -427,6 +427,7 @@ trait Interpreter extends ScorexLogging {
 
       val (aotReduced, jitReduced) = fullReduction(ergoTree, contextWithCost, env)
 
+      // assert(evalSettings.evaluationMode.isEmpty)
       val evalMode = getEvaluationMode(contextWithCost)
 
       // if necessary perform verification as v4.x (AOT based implementation)
--- a/sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
+++ b/sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
@@ -362,11 +362,11 @@ trait Interpreter extends ScorexLogging {
     *
     * Alternatively, the required evaluation mode can be specified by giving Some(mode)
     * value in `evalSettings.evaluationMode`, in which case the interpreter works as
-    * spedified.
+    * specified.
     */
   def getEvaluationMode(context: CTX): EvaluationMode = {
     evalSettings.evaluationMode.getOrElse {
-      if (context.activatedScriptVersion < Interpreter.MaxSupportedScriptVersion)
+      if (context.activatedScriptVersion < Interpreter.JitActivationVersion)
         AotEvaluationMode
       else
         JitEvaluationMode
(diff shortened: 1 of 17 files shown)


sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -366,6 +356,23 @@ trait Interpreter extends ScorexLogging {
...
+    * By default evaluation mode is determined based on `context.activatedScriptVersion`
+    * so that the interpreter works either as v4.x of v5.0.
+    *
+    * Alternatively, the required evaluation mode can be specified by giving Some(mode)
+    * value in `evalSettings.evaluationMode`, in which case the interpreter works as
+    * spedified.
+    */
+  def getEvaluationMode(context: CTX): EvaluationMode = {
+    evalSettings.evaluationMode.getOrElse {
+      if (context.activatedScriptVersion < Interpreter.MaxSupportedScriptVersion)
```
kushti, line 369: It is better  to introduce a new constant like JitActivationVersion, then there is no need to check MaxSupportedScriptVersion usages before increasing it in future (which can be easily missed)

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -366,6 +356,23 @@ trait Interpreter extends ScorexLogging {
     fullJitCost
   }
 
+  /** Returns evaluation mode used by this interpreter in the given context.
+    * By default evaluation mode is determined based on `context.activatedScriptVersion`
+    * so that the interpreter works either as v4.x of v5.0.
+    *
+    * Alternatively, the required evaluation mode can be specified by giving Some(mode)
+    * value in `evalSettings.evaluationMode`, in which case the interpreter works as
+    * spedified.
```
kushti, line 365: spedified -> specified

aslesarenko: fixed

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
aslesarenko: done

kushti (approved):

PR #756, opened 2021-12-08
Use JitCost type

### LINKED ISSUE #752: Type-tagging JIT costs
"For readability it can be solved by using JitCost tagged type over Long, with the corresponding arithmetic operations over JitCost values. So scaling can be encapsulated to JitCost -> Long and Long -> JitCost operations.
I would prefer to do this JitCost implementation in a separate PR, which can also be tracked by an issue."

aslesarenko:
Closes #752

### DIFF
--- a/sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
+++ b/sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
@@ -230,7 +230,7 @@ object ErgoLikeContext {
 case object MinerPubkey extends NotReadyValueByteArray with ValueCompanion {
   override def opCode: OpCode = OpCodes.MinerPubkeyCode
   /** Cost of calling Context.minerPubkey Scala method. */
-  override val costKind = FixedCost(20)
+  override val costKind = FixedCost(JitCost(20))
   override val opType = SFunc(SContext, SCollection.SByteArray)
   override def companion = this
   protected final override def eval(env: DataEnv)(implicit E: ErgoTreeEvaluator): Any = {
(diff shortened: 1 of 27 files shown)


kushti (changes requested):
Cost descriptions inconsistent now:

PerItemCost(JitCost(1), JitCost(1), 10)
PerItemCost(baseJitCost = 10, perChunkJitCost = 2, chunkSize = 1)
PerItemCost(5, 5, 64/*size of cache line in bytes*/)

FixedCost(20)

please use the same declaration style, also, avoid implicits as from reading code just it is still not possible to figure out what unit is considered.

aslesarenko:
No problem to use the same style, but which one?
 
There are 
- 150 declarations like FixedCost(5) 
- 40 declarations like PerItemCost(20, 3, 5) 
- and only 8 declarations like PerItemCost(baseJitCost = 20, perChunkJitCost = 7, chunkSize = 128)

If I remove (implicit o: Overloaded1) (I assume this is what you are referring to) then I will have to use this style:
- 150 declarations like FixedCost(JitCost(5)) 
- 40 declarations like PerItemCost(JitCost(20), JitCost(3), 5)
- 8 declarations like PerItemCost(baseCost = JitCost(20), perChunkCost = JitCost(7), chunkSize = 128)

Will it make the code more readable?

kushti:
> Will it make the code more readable?

Well, yes, also, allocations not hidden (initial plan was about using tagged typed as far as I remember), so you can do an issue to remove extra allocations when AOT costing will be removed.

aslesarenko:
> allocations not hidden

There is actually no allocations, I checked the generated Java code. 
It is because JitCost is a value class (https://docs.scala-lang.org/overviews/core/value-classes.html)

Tagged types didn't work for JitCost.
I tried to pass Int into a position where tagged type JitCost was expected, no compilation error. 
And the code generated for tagged types is more complex, likely also slower.

So JitCost will likely be without overhead.

aslesarenko:
@kushti comments addressed, please review

kushti (approved):

Issue #659, opened 2020-05-29
Serialize Option[T] in DataSerializer 


aslesarenko:
# Problem
DataSerializer is used to serialize data values for any supported ErgoTree type. It is in particular used to serialize Constant values, where type is first serialized (see TypeSerializer) and then the constant value is serialized using DataSerializer). 
Option[T] is supported type of constant values, i.e. it is possible to create tree node like `Constant(SOption(SInt), Some(10))`, however DataSerializer don't support serialization of Option[T] values and hence such constants are not supported.

# Solution
Add Option[T] case in DataSerializer using [Writer.putOption](https://github.com/ScorexFoundation/scorex-util/blob/3dc334f68ebefbfab6d33b57f2373e80245ab34d/src/main/scala/scorex/util/serialization/VLQWriter.scala#L130) method.

kushti:
done in #990

PR #758, opened 2021-12-17
Release Candidate v5.0

### LINKED ISSUE #464: Example which doesn't work in v4.x
UPD[@aslesarenko]:
Make sure the following example works with JITC.

For example,

path.fold(item, {(a: (Coll[Byte], Coll[Byte])) =>
    a._1.zip(a._2).map({ (c: (Byte, Byte)) => (c._1 ^ c._2).toByte })
})

does not compile, and I'm not sure whether it is okay or not. Please describe fold limitations in the spec.

### LINKED ISSUE #597: Missing GroupElement.multiply costing
UPD by @aslesarenko 
This depends on HF activation infrastructure of  #660 
The example below should fail in ErgoTree v0 and work in ErgoTree v1. 
There is nothing to fix, instead a new test should be added which fails under ET0 and passes under ET1

The following code `{ groupGenerator.multiply(groupGenerator) != groupGenerator }` when gone through the interpreter throws `Unexpected rejection cause: java.lang.NoSuchMethodException: sigmastate.eval.CostingRules$GroupElementCoster.multiply(scalan.Base$Ref), with root cause java.lang.NoSuchMethodException: sigmastate.eval.CostingRules$GroupElementCoster.multiply(scalan.Base$Ref)}` error.

### LINKED ISSUE #603: Fix CONTEXT.selfBoxIndex
Now it always returns `-1`.
Caused by `eq` in http://github.com/ScorexFoundation/sigmastate-interpreter/blob/a598658d63ad822a99ef91ecc79da906ad779ffe/sigmastate/src/main/scala/sigmastate/eval/CostingDataContext.scala#L639-L639 
Switching to `==` fixes it.

To test add:
```
  property("CONTEXT.selfBoxIndex") {
    testEval("CONTEXT.selfBoxIndex == 0")
  }
```
to `TestingInterpreterSpecification.scala`.

### LINKED ISSUE #604: Make sure this example works with JITC
[UPD by @aslesarenko] A test case is needed for the script example below which make sure this script works with the new ErgoTreeEvaluator

A thorough check is needed.
Code:
```
 val headers = CONTEXT.headers
 val ids = headers.map({(h: Header) => h.id })
 val parentIds = headers.map({(h: Header) => h.parentId })
 headers.indices.slice(0, headers.size - 1).forall({ (i: Int) =>
   val parentId = parentIds(i)
   val id = ids(i + 1)
   parentId == id
 })
```
gives error:
```
scalan.Base$StagingException: Error in Evaluation.compile.evaluate(s4571 -> Placeholder(BaseElemLiftable<Int>)), with root cause scalan.Base$StagingException: Don't know how to evaluate(s4571 -> Placeholder(BaseElemLiftable<Int>))}
```

### LINKED ISSUE #608: Validate positive Int during deserialization
Depend on https://github.com/ScorexFoundation/scorex-util/issues/14
Current situation is like this, and needs more consistent use of `getUIntExact`

image

### LINKED ISSUE #640: Fix xorOf
>"A chain of XORs—a XOR b XOR c XOR d (and so on)—is true whenever an odd number of the inputs are true and is false whenever an even number of inputs are true."

https://en.wikipedia.org/wiki/Exclusive_or

Xor(0, 1, 0) => 1 
as (0 ^ 1) ^ 0 = 1

Xor(0, 1, 1)  => 0
as (0 ^ 1) ^ 1 = 0

Current implementation - http://github.com/ScorexFoundation/sigmastate-interpreter/blob/42e55cbfd093252b8005e4607970764dd6610cbe/sigma-impl/src/main/scala/special/sigma/SigmaDslOverArrays.scala#L41-L41

### LINKED ISSUE #662: Implement all TODO JITC

### LINKED ISSUE #665: Implement JITC Specification tests
### Problem
- JIT costing evaluates ErgoTree and performs cost accumulation along the execution trace.
- There is more than 70 operations in ErgoTree and many of them have complex semantics.
- Execution trace of many operations depend on actual values of arguments. For example `||` and `&&` operations don't have short-cutting semantics and don't always evaluate right arguments, which depends on the actual value of the left argument. 
- calculation of cost is part of consensus protocol, because it is used in transaction validation (a transaction can be invalidated because of the cost)
- We need to ensure the the cost function is fixed and never change between releases.

### Solution
In addition to Spam Coverage Report which is used to catch regressions, we can add unit tests which:
1) serve as regression tests
2) test individual operations with internal branching logic (like `||` and `&&`)
3) specify explicitly expected trace of performed operations 
4) cover all alternative branches and specify the expected cost

### LINKED ISSUE #744: Allow return `x` in `flatMap` (e.g. `flatMap { (x: ...) => x }`)
From discord https://discord.com/channels/668903786361651200/840313005064585246/902729968331685988 :
```
I currently have a 

Coll[Coll[(SigmaProp, Coll[Byte])]]

that I am simply trying to flatten into a 

Coll[(SigmaProp, Coll[Byte])]

I tried 

flatMap{ (x : Coll[(SigmaProp, Coll[Byte])]) => x }

 but am running into an error that says: 

Unsupported lambda in flatMap: allowed usage `xs.flatMap(x => x.property)
```

### LINKED ISSUE #749: Introduce a new limit on structural depth of data values
corresponds to todo: "TODO v5.0: introduce a new limit on structural depth of data values" in DataValueComparer

UPDATE (@aslesarenko): 
The goal is to prevent instantiation of complex intermediate values.
Then comparing them will never be an issue.

### LINKED ISSUE #765: Fix Coll.append
The following code shows how the problem with CollectionUtil.concatArrays manifests itself in Coll.append
```scala
      val xs = builder.fromItems(1, 2)
      val pairs = xs.zip(xs)
      val ys = pairs.map(squared(inc)) // this map transforms PairOfCols to CollOverArray

      // due to the problem with concatArrays
      an[ClassCastException] should be thrownBy (ys.append(ys))
```
See the [test](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/e8c069b0f28b3c96018675985c4f03cd64b007cd/library/src/test/scala/special/collections/CollsTests.scala#L255)

### LINKED ISSUE #766: Fix Coll.zip
See description in the [test](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/e8c069b0f28b3c96018675985c4f03cd64b007cd/library/src/test/scala/special/collections/CollsTests.scala#L280)

### LINKED ISSUE #769: Fix SubstConstants
See description in the [code](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/985ce41e2e44d82a44ac57b700c7c735f855428e/sigmastate/src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala#L272)

### LINKED ISSUE #824: Fix equality of PairOfCols
### Background

ErgoTree interpreter v4.x sometimes use optimized representation of collections of pairs, i.e. collections with elements of type (T1, T2), like `Coll[(Coll[Byte], Long)]`.
When a collection contains pairs then it can be represented using PairOfCols class

```scala
class PairOfCols[L, R](val ls: Coll[L], val rs: Coll[R]) extends PairColl[L,R]
```
This representation is much faster for many operations because data is stored unboxed in arrays (cache friendly memory layout), which leads to 2-3x speedups. 

However, some operations like `map` can change representation so that collection of pairs is backed by array of pairs and represented using 
```scala 
class CollOverArray[A](val toArray: Array[A])(implicit tA: RType[A]) extends Coll[A]
```

This for example happens in the following ErgoScript code
```scala
val xs: Coll[(Int, Long)] = ...
val ys = xs.map(in => (in._1, in._2))
```
Here, xs and ys contain the same data, but may have different representation classes.
Assuming xs is represented by PairOfCols, after `map` ys will be represented by CollOverArray.

### The problem

EQ operation of ErgoTree (`==` in ErgoScript) is implemented using `equals` method in collection classes above which (erroneously) treats different representations as not equal, even if the content is the same. 

### Solution
Implementation of EQ should be implemented explicitly using recursive traversal over collection structure instead of relying on `equals` implementation.

aslesarenko:
This PR finalizes v5.0 implementation.

**Summary of changes:**
 - AOT costing of ErgoTree will be replaced by JIT costing
 - introduced new AOT costing of signature verification
 - fixed bug in xorOf global operation
 - fixed bug in CONTEXT.selfBoxIndex operation
 - removed limitations on lambda argument of Coll.flatMap 
 - fixed bug in Coll.append (++ operation)
 - fixed bug in Coll.zip
 - fixed bug in substConstants
 - perfomance optimizations
 - code cleanups

 **Ready for testnet**
  - [x] Implement simple cases of // TODO optimize (in #761)
  - [x] Removed unfinished CreateAvlTree operation
  - [x]  #767
  - [x] Closes #749
  - [x] Closes #608 in #762
  - [x] Closes #640 in #760 and #777
  - [x] Closes #603 in #772
  - [x] Closes #744 in #775 
  - [x] Closes #765 in #771
  - [x] Closes #766 in #771
  - [x] Closes #769 in #773
  - [x] Prepare for #659 to be solved in v6.0
  - [x] #805 
  - [x] Closes #824 

**Ready for mainnet**
  - [x] Implement // TODO cover with tests
  - [x] #779
  - [x] Closes #662 
  - [x] Closes #604 in #780
  - [x] Closes #464 in #788
  - [x] Closes #597 
  - [x] Closes #665 in #785


kushti:
#828 must be resolved before this PR can be accepted

aslesarenko:
@kushti the fix it here #829, however it read the description about why it is better be merged into v5.x (and not here)

sigmastate/src/main/scala/org/ergoplatform/ErgoScriptPredef.scala
```diff
@@ -18,8 +18,7 @@ object ErgoScriptPredef {
   import sigmastate.interpreter.Interpreter._
 
   def compileWithCosting(env: ScriptEnv, code: String, networkPrefix: NetworkPrefix)(implicit IR: IRContext): Value[SType] = {
-    val compiler = new SigmaCompiler(CompilerSettings(
-      networkPrefix, TransformingSigmaBuilder, lowerMethodCalls = true))
+    val compiler = new SigmaCompiler(networkPrefix)
```
kushti, line 21: Unused imports not fixed after this transformation

sigma-impl/src/main/scala/special/sigma/SigmaDslOverArrays.scala
```diff
@@ -12,6 +11,9 @@ import scorex.crypto.hash.{Sha256, Blake2b256}
 import special.SpecialPredef
 import special.collection._
 import scalan.util.Extensions.BigIntegerOps
+import sigmastate.VersionContext
+import sigmastate.VersionContext.JitActivationVersion
```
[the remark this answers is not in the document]
aslesarenko: fixed

sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
```diff
@@ -111,6 +111,18 @@ class ErgoLikeContext(val lastBlockUtxoRoot: AvlTreeData,
   */
   val self: ErgoBox = boxesToSpend(selfIndex)
 
+  /** Current version of the ErgoTree executed by the interpreter.
+    * This property is used to implement version dependent operations and passed to
+    * interpreter via [[specia.sigma.Context]].
```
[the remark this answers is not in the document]
aslesarenko: fixed

sigmastate/src/main/scala/org/ergoplatform/ErgoScriptPredef.scala
aslesarenko: fixed

sigmastate/src/main/scala/sigmastate/lang/Terms.scala
```diff
@@ -140,29 +140,20 @@ object Terms {
 
     protected final override def eval(env: DataEnv)(implicit E: ErgoTreeEvaluator): Any = {
       addCost(Apply.costKind)
-      if (args.isEmpty) {
-        // TODO coverage
-        val fV = func.evalTo[() => Any](env)
-        fV()
-      }
-      else if (args.length == 1) {
+      if (args.length == 1) {
```
kushti, line 143: ScalaDoc missed for both `eval` method and `Apply` class. Also, add comment why new behavior is non-breaking. And comment args limits  etc

sigmastate/src/main/scala/sigmastate/lang/SigmaPredef.scala
```diff
@@ -180,6 +180,16 @@ object SigmaPredef {
           Seq(ArgInfo("", "")))
     )
 
+    val FromBase16Func = PredefinedFunc("fromBase16",
```
kushti, line 183: ScalaDoc missed

kushti (changes requested):

sigmastate/src/main/scala/sigmastate/SigSerializer.scala
```diff
@@ -106,9 +114,12 @@ object SigSerializer extends LazyLogging {
     *
     * @param exp   sigma proposition which defines the structure of bytes from the reader
     * @param proof proof to extract challenges from
+    * @param E     optional evaluator (can be null) which is used for profiling of operations.
```
kushti, line 117: Will it be removed in 5.x? This parameter is pretty confusing.

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -1,46 +1,94 @@
 package sigmastate.interpreter
 
-import java.lang.{Math => JMath}
 import java.util
```
kushti, line 3: unused import

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -1,46 +1,94 @@
...
-import sigmastate.lang.exceptions.{CostLimitException, InterpreterException}
+import sigmastate.interpreter.Interpreter._
+import sigmastate.lang.exceptions.InterpreterException
 import sigmastate.serialization.{SigmaSerializer, ValueSerializer}
-import sigmastate.utils.Helpers
-import sigmastate.utils.Helpers._
 import sigmastate.utxo.DeserializeContext
 import sigmastate.{SType, _}
+import sigmastate.eval.{Evaluation, IRContext, Profiler}
+import scalan.util.BenchmarkUtil
```
kushti, line 22: unused import

sigmastate/src/main/scala/sigmastate/lang/SigmaPredef.scala
aslesarenko: added description in OperationInfo

sigmastate/src/main/scala/sigmastate/lang/SigmaCompiler.scala
```diff
@@ -27,6 +27,10 @@ case class CompilerSettings(
 )
 
 class SigmaCompiler(settings: CompilerSettings) {
```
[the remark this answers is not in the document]
aslesarenko: done. Note this class has been refactored in v5.0.1 PRs, where more ScalaDocs has been written for all public methods.

sigmastate/src/main/scala/sigmastate/serialization/ModQSerializer.scala
```diff
@@ -5,7 +5,7 @@ import sigmastate.lang.Terms._
 import sigmastate.utils.{SigmaByteReader, SigmaByteWriter}
 import sigmastate.{ModQ, SType}
 
-// TODO HF (2h): make sure it is covered with tests
+// TODO v6.0 (2h): make sure it is covered with tests
```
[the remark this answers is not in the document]
aslesarenko: This class doesn't make sense before v6.0. It cannot be removed however, because this serializer was erroneously registered in ValueSerializer's global map. See comment [here](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/dac1109d284b5342859fcd4a0b34b87a502a029d/sigmastate/src/main/scala/sigmastate/serialization/ValueSerializer.scala#L143)

sigmastate/src/main/scala/sigmastate/serialization/ModQArithOpSerializer.scala
```diff
@@ -6,7 +6,7 @@ import sigmastate.utils.SigmaByteWriter.DataInfo
 import sigmastate.utils.{SigmaByteReader, SigmaByteWriter}
 import sigmastate.{ModQArithOpCompanion, SType, ModQArithOp}
 
-// TODO HF (2h): make sure it is covered with tests
+// TODO v6.0 (2h): make sure it is covered with tests
```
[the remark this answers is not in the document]
aslesarenko: see my comment above.

sigmastate/src/main/scala/sigmastate/serialization/TupleSerializer.scala
```diff
@@ -24,7 +25,7 @@ case class TupleSerializer(cons: Seq[Value[SType]] => Value[SType])
 
   override def parse(r: SigmaByteReader): Value[SType] = {
     val size = r.getByte()
-    val values = ValueSerializer.newArray[SValue](size) // assume size > 0 so always create a new array
+    val values = safeNewArray[SValue](size) // assume size > 0 so always create a new array
```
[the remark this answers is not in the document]
aslesarenko: This refers to the fact that if the resulting array is empty, the allocation can be safely avoided because empty arrays are immutable and can be shared. See for example [here](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/e1af1878fcb6ece28114d65b878cdc913c0e7bf3/sigmastate/src/main/scala/sigmastate/serialization/BlockValueSerializer.scala#L32)
This gives speedups observable in a profiler.

But since tuples now always have 2 elements, this array is always non-empty, so the optimization is not applicable.

clarified the comment.

sigmastate/src/main/scala/sigmastate/SigSerializer.scala
aslesarenko: It is only necessary for profiling and setting up cost parameters of signature verification. See estimateCryptoVerifyCost method. Can be removed, but I suggest to keep it.
What exactly not clear, I can add clarifying description.

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
aslesarenko: it is actually used, IntelliJ error. Try to remove and compile

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
aslesarenko: fixed

sigmastate/src/main/scala/sigmastate/SigSerializer.scala
kushti: It is not cleat at all how a serializer could be dependent on an evaluator, as the latter is higher-level entity than the former.

kushti (approved):

sigmastate/src/main/scala/sigmastate/SigSerializer.scala
aslesarenko: @kushti Yes, agree, but the profiling infrastructure is built around evaluator. Will think how it may be refactored.

PR #760, opened 2021-12-21
Finalize v5.0: fix  XorOf operation

### LINKED ISSUE #640: Fix xorOf

aslesarenko:
Closes #640


sigma-impl/src/main/scala/special/sigma/SigmaDslOverArrays.scala
```diff
@@ -40,8 +42,21 @@ class TestSigmaDslBuilder extends SigmaDslBuilder {
 
   @NeverInline
   override def xorOf(conditions: Coll[Boolean]): Boolean = {
-    // This is buggy version used in v4.x interpreter (for ErgoTrees v0, v1)
-    conditions.toArray.distinct.length == 2
+    if (Versions.currentErgoTreeVersion >= JitActivationVersion) {
```
kushti, line 45: Maybe condition can be done for interpreter version? I see no big semantic problems about it. In case of ergo tree version used, we need for roadmap for compiler and tooling update, but I do not see Github issues or other efforts towards that. Also, documentation for developers needed on how xorOf improperly working tor trees v. 0/1. 

I suspect if xorOf used anywhere, unlikely current behavior is intended, thus it could be cleaner to fix interpreter for all the version just.

aslesarenko: Ok, this can be done with the following condition 
`VersionContext.current.activatedVersion >= JitActivationVersion`. 
So the new behavior will only take effect after activation.

aslesarenko: It cannot be done in v5.0-fix-xorOf branch, because test harness is not ready yet. 
So I did it in a separate related PR (see part 2) where the changes are very small.

kushti (approved):

PR #761, opened 2021-12-22
Finalize v5.0 (part1)

aslesarenko:
- [x] implemented some // TODO optimize
- [x] more tests for TypeSerializer
- [x] negative test cases when lambda is not with single argument
- [x] more // TODO cover
- [x] fixed CollectionUtil.joinSeq method + tests
- [x] identified the problem with Coll.append (added regression tests for ET v0, v1)


sigmastate/src/main/scala/sigmastate/utxo/transformers.scala
```diff
@@ -69,7 +70,13 @@ case class Append[IV <: SType](input: Value[SCollection[IV]], col2: Value[SColle
     val inputV = input.evalTo[Coll[IV#WrappedType]](env)
     val col2V = col2.evalTo[Coll[IV#WrappedType]](env)
     addSeqCost(Append.costKind, inputV.length + col2V.length) { () =>
-      inputV.append(col2V)
+      if (E.context.currentErgoTreeVersion >= JitActivationVersion) {
```
kushti, line 73: Maybe condition can be done for interpreter version? I see no big semantic problems about it. In case of ergo tree version used, we need for roadmap for compiler and tooling update, but I do not see Github issues or other efforts towards that.

aslesarenko: If "interpreter version" means release version, then no, the semantics should be connected to ergoTree version, because we never know which release will be executing the ergoTree.

sigmastate/src/main/scala/sigmastate/Values.scala
```diff
@@ -1175,13 +1170,7 @@ object Values {
 
     protected final override def eval(env: DataEnv)(implicit E: ErgoTreeEvaluator): Any = {
       addCost(FuncValue.costKind)
-      if (args.length == 0) {
```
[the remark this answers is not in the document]
aslesarenko: It is not breaking change, because it is a change in the new interpreter. This code shouldn't be there in a first place, so consider it a cleanup. And there are tests that cover this.
Also v4.x only support single argument lambdas.

sigmastate/src/main/scala/sigmastate/serialization/ValueSerializer.scala
```diff
@@ -50,7 +50,6 @@ object ValueSerializer extends SigmaSerializerCompanion[Value[SType]] {
     Relation2Serializer(LE, mkLE[SType]),
     Relation2Serializer(EQ, mkEQ[SType]),
     Relation2Serializer(NEQ, mkNEQ[SType]),
-    CreateAvlTreeSerializer(mkCreateAvlTree),
```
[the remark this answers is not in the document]
aslesarenko: Ops, you are right, Even though the operation is not implemented, and will never be executed, the serializer is unfortunately registered and implemented. So, will rollback this removal.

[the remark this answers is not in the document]
aslesarenko: rolled back the removal

sigmastate/src/main/scala/sigmastate/utxo/transformers.scala
kushti: Your code is executing a tree, you also know behavior of previous versions. You don't know what future versions will do, but that's their business to preserve compatibility (or not).
If you're going with tree version based versioning, see comment above.

sigmastate/src/main/scala/sigmastate/Values.scala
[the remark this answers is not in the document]
kushti: Serializer allows for args.length != 1, so where args.length != 1 prevented before this code being executed?

sigmastate/src/main/scala/sigmastate/utxo/transformers.scala
aslesarenko: I don't understand "You don't know what future versions will do, but that's their business to preserve compatibility (or not)". 
The semantics of each ErgoTree version should be fixed and preserved regardless of which Interpreter implementation executes it. 
Any ErgoTree v0, v1 will still be valid and existing apps can create new transactions with such (not the most recent) versions.
If they opt to not upgrade their code, they will have old semantics and operation behavior. If they need fixed operations they should upgrade. This is their choice.
Appkit already uses the latest activated ergo tree version for new transactions (it takes it from the Node API). 
I don't know about other toolchains, they can use Appkit as reference implementation.

sigmastate/src/main/scala/sigmastate/Values.scala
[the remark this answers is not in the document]
aslesarenko: In v4.x each FuncValue should be processed by this case https://github.com/ScorexFoundation/sigmastate-interpreter/blob/e0905cc8c135fc4a6669f04636dd888aaa11572b/sigmastate/src/main/scala/sigmastate/eval/RuntimeCosting.scala#L1760
And each Apply node by this  https://github.com/ScorexFoundation/sigmastate-interpreter/blob/e0905cc8c135fc4a6669f04636dd888aaa11572b/sigmastate/src/main/scala/sigmastate/eval/RuntimeCosting.scala#L1403

They only support single argument, all others will go to https://github.com/ScorexFoundation/sigmastate-interpreter/blob/e0905cc8c135fc4a6669f04636dd888aaa11572b/sigmastate/src/main/scala/sigmastate/eval/RuntimeCosting.scala#L1857

So, even if ErgoTree can be deserialized, it still can be rejected.

sigmastate/src/main/scala/sigmastate/utxo/transformers.scala
kushti: I see no any reason to discuss even whether to do documentation for protocol changes or not. 

AppKit is not the only tool and will never be so.

Also, why activation rules for XorOf (proposed in https://github.com/ScorexFoundation/sigmastate-interpreter/pull/777) and Append are different?

aslesarenko: You are kidding? They were the same, you proposed to switch from versioning based on ergoTree version (the original variant) to versioning based on activatedVersion. I agree with the reasoning there. In principle they can be applied here as well.
Note, there are other fixes in other operations, should we follow the same versioning logic there as well?

I think I described by logic behind the current implementation, don't want to argue on this.

1) You have the final word, let me know whether it should be changed as you suggest or left as is? I will make the changes. 
2) If documentation is necessary, I can describe changes in LangSpec.md

aslesarenko: Removed changes in Append. The changes will be reintroduced in #771

sigmastate/src/main/scala/sigmastate/eval/Zero.scala
```diff
@@ -19,7 +19,7 @@ case class CZero[T](zero: T) extends Zero[T]
 
 trait ZeroLowPriority {
   implicit def collIsZero[T: Zero: RType]: Zero[Coll[T]] = CZero(Colls.emptyColl[T])
-  implicit def optionIsZero[T: Zero]: Zero[Option[T]] = CZero(Some(Zero.zeroOf[T]))
```
kushti, line 22: Is this change safe? Used in EvalSizeBox.getReg, are implementations getting the same result there for 

```
        val z = Zero.typeToZero(tT)
        val zeroVal = z.zero
        val sized = Sized.typeToSized(tT)
        sized.size(zeroVal)
```

?

sigmastate/src/main/scala/sigmastate/lang/Terms.scala
```diff
@@ -140,21 +140,12 @@ object Terms {
 
     protected final override def eval(env: DataEnv)(implicit E: ErgoTreeEvaluator): Any = {
       addCost(Apply.costKind)
-      if (args.isEmpty) {
-        // TODO coverage
-        val fV = func.evalTo[() => Any](env)
-        fV()
-      }
-      else if (args.length == 1) {
+      if (args.length == 1) {
```
kushti, line 143: The change is not described in PR desc, no comments, where is the same constraint for 4.0 ?

aslesarenko: Already discussed, here is my answer https://github.com/ScorexFoundation/sigmastate-interpreter/pull/761#discussion_r779565072

sigmastate/src/main/scala/sigmastate/eval/Zero.scala
aslesarenko: rolled it back.

kushti (approved):
Finally, LGTM, but please check PR description (outdated now)

PR #763, opened 2021-12-22
Finalize v5.0: fix non-lazy If

aslesarenko:
Old implementation is done in https://github.com/ScorexFoundation/sigmastate-interpreter/blob/e0905cc8c135fc4a6669f04636dd888aaa11572b/sigmastate/src/main/scala/sigmastate/eval/RuntimeCosting.scala#L1722 which is non-lazy.

The new v5.0 implementation is done in https://github.com/ScorexFoundation/sigmastate-interpreter/blob/68365862ac5548a644d91f9b062a38b1bebcfeef/sigmastate/src/main/scala/sigmastate/trees.scala#L1495-L1494 which is lazy for each branch and fixes the problem (See `property("Conditional access (data box register)")`).

In this PR:
- [x] ErgoTreeEvaluator always check global versions to be equal Context versions
- [x] added allowNewToSucceed parameter for ChangedFeature descriptor
- [x] the SigmaDslSpecification test for `if` statement is updated to correctly test difference between v4.x and v5.0 implementations (https://github.com/ScorexFoundation/sigmastate-interpreter/pull/763/commits/dd59bd09096957c87773cdb577440879ce4e8c56)

### DIFF
--- a/sigmastate/src/main/scala/sigmastate/interpreter/ErgoTreeEvaluator.scala
+++ b/sigmastate/src/main/scala/sigmastate/interpreter/ErgoTreeEvaluator.scala
@@ -2,17 +2,17 @@ package sigmastate.interpreter
 
 import org.ergoplatform.ErgoLikeContext
 import org.ergoplatform.SigmaConstants.ScriptCostLimit
-import sigmastate.{FixedCost, JitCost, PerItemCost, SType, TypeBasedCost}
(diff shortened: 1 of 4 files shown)


kushti (changes requested):
Make PR description. Where is a fix for non-lazy if specifically ?

kushti (approved):

PR #768, opened 2021-12-24
Finalize v5.0: Protected methods in Interpreter

aslesarenko:
This PR marks most of the Interpreter methods as `protected`.
As result all public methods accept ErgoTree (and not just Value).
As result many test suits now test behavior of all versions of ErgoTree and as result both old and the new v5.0 interpreter.

- [x] reduceToCrypto become protected
- [x] public interface of Interpreter minimized
- [x] TestEvaluationMode removed


kushti (changes requested):
reduceToCrypto is called in Ergo node, ScriptApiRoute, L. 106: 
          val res = interpreter.reduceToCrypto(req.ctx.asInstanceOf[interpreter.CTX], prop)

aslesarenko:
> reduceToCrypto is called in Ergo node, ScriptApiRoute, L. 106:

removed direct usage reduceToCrypto in https://github.com/ergoplatform/ergo/pull/1579/files

kushti (approved):

PR #771, opened 2021-12-28
Finalize v5.0: fix Zip and Append

aslesarenko:


common/src/main/scala/sigmastate/VersionContext.scala
```diff
@@ -20,6 +20,10 @@ case class VersionContext(activatedVersion: Byte, ergoTreeVersion: Byte) {
 
   /** @return true, if JIT version of Ergo protocol was activated. */
   def isJitActivated: Boolean = activatedVersion >= JitActivationVersion
+
+  /** @return true, if ErgoTree version >= JitActivationVersion. */
+  def isEvaluateErgoTreeUsingJIT: Boolean =
```
kushti, line 25: Sounds like a tree of a version where isEvaluateErgoTreeUsingJIT is not evaluated used JIT costing, which is not true.

library-impl/src/main/scala/special/collection/CollsOverArrays.scala
```diff
@@ -59,7 +60,12 @@ class CollOverArray[@specialized A](val toArray: Array[A])(implicit tA: RType[A]
   @NeverInline
   def append(other: Coll[A]): Coll[A] = {
     if (toArray.length <= 0) return other
-    val result = CollectionUtil.concatArrays(toArray, other.toArray)
+    val result = if (VersionContext.current.isEvaluateErgoTreeUsingJIT) {
```
kushti, line 63: explicit ErgoTree version comparison would be a better description here. Maybe because a comment is missed.

And as for other PRs not clear why versioning here is tree based while for other operations it is interpreter-based. Also, documentation is still missed.

kushti (changes requested):

common/src/main/scala/sigmastate/VersionContext.scala
kushti: "isEvaluateErgoTreeUsingJIT" sounds like whether a tree should be evaluated using JIT or not, but in reality the function checks that tree version is above threshold (and it is not related to JIT evaluation in general)

library-impl/src/main/scala/special/collection/CollsOverArrays.scala
kushti: I am not enforcing. Both options are possible, but for tree-based versioning documentation should be done for application developer for possible versions, and how to choose a version. Here you pass a choice to a developer thus must document how to make it (otherwise, the work is simply not complete thus does not make sense).
Interpreter-based activation is implicit for a developer, so requires only for some list for interpreter developers, to track changes.

common/src/main/scala/sigmastate/VersionContext.scala
aslesarenko: I would like to keep the method and use it where necessary.
So, if mentioning JIT is misleading here I can rename it to `isErgoTreeVersionGreaterV1`. 
@kushti Is it ok?

kushti: @aslesarenko yes, sounds precise now

aslesarenko: Done

library-impl/src/main/scala/special/collection/CollsOverArrays.scala
aslesarenko: Ok, changed to activation-based versioning. 
1) since this is semantic bug fix and, as you mentioned already, people usually expect the operation to have correct semantics
2) after activation it will simplify usability, as dealing with versions will not be required.

The tests are updated accordingly

kushti (approved):
LGTM, but please note:

1) for interpreter, version, isActivatedVersionJit sounds ok , and even more precise than isActivatedVersionGreaterV1
2) Please check the following commented out test in CollsTest  
```
// TODO the following test fails because equality of Seq is not deep, and nested arrays are shallow compared
//    forAll(tokensGen, minSuccess) { c1 =>
//      println(s"c1=${c1.x.toArray.toSeq.map { case (id, v) => (id.toSeq, v) }}")
//      val tokens = c1.x
//      assert(tokens.toArray.toSeq == tokensArr.toSeq)
//    }
```

maybe it is not actual anymore, if actual, better to do GH issue, as commented out code , as well as TODO in the testing codebase can be forgotten for like forever

aslesarenko:
> * for interpreter, version, isActivatedVersionJit sounds ok , and even more precise than isActivatedVersionGreaterV1
> 
> * Please check the following commented out test in CollsTest

1) I would stick with isActivatedVersionGreaterV1 for consistency.
2) This is obsolete check, removed.

kushti:
@aslesarenko named constants are used for improving readability, consistency is not needed there. For interpreter, isJitActivated says more than  isActivatedVersionGreaterV1

aslesarenko:
Ok, will change in the next PR

PR #774, opened 2022-01-01
InputLike interface with boxId

pragmaxim:
Regarding this comment https://github.com/ergoplatform/ergo/pull/1549#discussion_r777116646

We could have functions accepting `InputLike` when they are only interested in `boxId`


sigmastate/src/main/scala/org/ergoplatform/Input.scala
```diff
@@ -9,12 +9,19 @@ import sigmastate.interpreter.{ContextExtension, ProverResult}
 import sigmastate.serialization.SigmaSerializer
 import sigmastate.utils.{Helpers, SigmaByteReader, SigmaByteWriter}
 
+sealed trait InputLike {
```
aslesarenko, line 12: Please describe in ScalaDocs the role this trait plays.

aslesarenko (changes requested):

PR #776, opened 2022-01-03
Removed commons-io

satsen:
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/develop/docs/notes.md

Note: I made sure to only use JDK 7 and Scala 2.11 methods


satsen:
I also tried to remove `cats` since it's only used 3 times and in one class only (JsonCodecs), but the strange Scala syntax is just beyond me. After hours I made up something but I don't even know if it will function the same, so if anybody can get rid of `cats`, that would be great.

aslesarenko:
@satsen, please fix compilation

satsen:
@aslesarenko That is code that I did not touch. My Scala knowledge is not enough to fix errors I did not cause, perhaps github uses wrong scala, i don't know.

aslesarenko:
You removed dependency and some other code doesn't compile even though you don't touch it.

PR #789, opened 2022-04-02
V5.0 removing arrays hash code

### LINKED ISSUE #197: Revise and optimize hashCode for Digest32 and other hashes
We have some places where we compute `Arrays.hashCode(d)`, for some hash d.
This is not necessary, because d already have random bytes which we can use to build Int.
`Ints.fromByteArray(d)` should be much more efficient.
Moreover, this hashCode has actually better randomness and lower collision rate.

jozanek:
Closes #197


jozanek:
@aslesarenko is it safe to assume that digest and challenge are always non-empty?

sigmastate/src/main/scala/sigmastate/interpreter/ProverResult.scala
```diff
@@ -13,6 +13,7 @@ import sigmastate.utils.{SigmaByteReader, SigmaByteWriter}
   * @param extension - user-defined variables to be put into context
   */
 class ProverResult(val proof: Array[Byte], val extension: ContextExtension) {
+  // Usage of Ints.fromByteArray isn't allowed, as `proof` can be empty array
```
aslesarenko, line 16: not only that, but the proof prefix bytes are not random, so applying this optimization doesn't make sense. Please remove the comment.

sigmastate/src/main/scala/sigmastate/utils/Helpers.scala
```diff
@@ -100,13 +100,6 @@ object Helpers {
     case arr: Array[Boolean] => util.Arrays.hashCode(arr)
   }
 
-  def optionArrayEquals[A](maybeA1: Option[Array[A]], maybeA2: Option[Array[A]]): Boolean =
-    (maybeA1, maybeA2) match {
-      case (None, None) => true
-      case (Some(a1), Some(a2)) => deepHashCode(a1) == deepHashCode(a2)
-      case _ => false
-    }
```
aslesarenko, line 108: Please rollback this removal.

common/src/main/scala/scalan/util/CollectionUtil.scala
```diff
@@ -354,18 +354,6 @@ object CollectionUtil {
...
-      var result: Int = 1
-      for (x <- xs) {
-        val elementHash = x match {
-          case arr: Array[_] => CollectionUtil.deepHashCode(arr)
-          case _ => Objects.hashCode(x)
-        }
-        result = 31 * result + elementHash;
-      }
-      result
-    }
```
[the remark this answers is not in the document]
jozanek: Wasn't used anywhere.

sigmastate/src/main/scala/org/ergoplatform/validation/RuleStatus.scala
```diff
@@ -45,6 +45,7 @@ case class  ReplacedRule(newRuleId: Short) extends RuleStatus {
 case class ChangedRule(newValue: Array[Byte]) extends RuleStatus {
   val statusCode: Byte = RuleStatus.ChangedRuleCode
 
+  // Usage of Ints.fromByteArray isn't allowed, as `newValue` can be empty array
```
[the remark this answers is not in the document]
jozanek: Removed in new version.

sigmastate/src/main/scala/sigmastate/interpreter/ProverResult.scala
jozanek: Removed in new version. The comment was put there to explain why we use Ints.fromByteArray everywhere in the codebase except these 2 usages.

sigmastate/src/main/scala/sigmastate/utils/Helpers.scala
jozanek: Removed in new version.

sigmastate/src/main/scala/sigmastate/UncheckedTree.scala
```diff
@@ -24,7 +26,7 @@ trait UncheckedConjecture extends UncheckedSigmaTree with ProofTreeConjecture {
   })
 
   override def hashCode(): Int =
-    31 * Arrays.hashCode(challenge) + children.hashCode()
+    31 * Ints.fromByteArray(challenge) + children.hashCode()
```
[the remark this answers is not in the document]
jozanek: Added new (mixed) way of getting hash code.

sigmastate/src/main/scala/sigmastate/UncheckedTree.scala
```diff
@@ -13,6 +15,9 @@ case object NoProof extends UncheckedTree
 
 sealed trait UncheckedSigmaTree extends UncheckedTree {
   val challenge: Array[Byte]
+  lazy val challengeOptimizedHash =
+    if (challenge.size > 4) Ints.fromByteArray(challenge)
+    else Arrays.hashCode(challenge)
```
aslesarenko, line 20: Why strictly > 4? Also using lazy may be problematic, because challenge is declared as val and may be overriden. This makes initialization order tricky. 
Please change to def.

sigmastate/src/main/scala/sigmastate/UncheckedTree.scala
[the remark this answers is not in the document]
aslesarenko: please see new comment

sigmastate/src/main/scala/sigmastate/UncheckedTree.scala
jozanek: It comes from guava documentation: throws IllegalArgumentException - if bytes has fewer than 4 elements

aslesarenko: what about 4 bytes?

sigmastate/src/main/scala/sigmastate/UncheckedTree.scala
```diff
@@ -102,7 +102,7 @@ case class COrUncheckedNode(override val challenge: Challenge,
 
 case class CThresholdUncheckedNode(override val challenge: Challenge,
                                    override val children: Seq[UncheckedSigmaTree],
-                                   k: Integer,
+                                   k: Int,
```
aslesarenko, line 105: Why this change?

sigmastate/src/main/scala/sigmastate/UncheckedTree.scala
jozanek: Tested with 4 bytes, guava doesn't throw exception. See updated code in new version.

sigmastate/src/main/scala/sigmastate/UncheckedTree.scala
jozanek: As discussed above, we should avoid using boxed numerics.

aslesarenko: ah, right, Int will be unboxed, good catch.

aslesarenko (changes requested):

sigmastate/src/test/scala/sigmastate/UncheckedTreeSpecification.scala
```diff
@@ -0,0 +1,91 @@
...
+      cThresholdUncheckedNode(bigChallenge)
+    )
+
+    forAll(bigSizeExamples) { uncheckedTree =>
+      uncheckedTree.challengeOptimizedHash shouldBe Ints.fromByteArray(uncheckedTree.challenge)
+    }
+
+    // following hashes are just manual calculations from definition
+    val schnorrHash = 31 * (31 * 2433880 + 1) - 185272974
+    uncheckedSchnorr(emptyChallenge).hashCode() shouldBe schnorrHash
```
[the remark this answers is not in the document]
jozanek: Added property test to latest version.

sigmastate/src/test/scala/sigmastate/UncheckedTreeSpecification.scala
```diff
@@ -0,0 +1,104 @@
...
+      cOrUncheckedNode(size4Challenge),
+      cOrUncheckedNode(bigChallenge),
+      cThresholdUncheckedNode(emptyChallenge),
+      cThresholdUncheckedNode(size3Challenge),
+      cThresholdUncheckedNode(size4Challenge),
+      cThresholdUncheckedNode(bigChallenge)
+    )
+
+    forAll(equalHashCodes) { uncheckedTree =>
+      uncheckedTree.hashCode shouldBe uncheckedTree.hashCode
```
aslesarenko, line 101: what are you testing here?

jozanek: Basically determinism on hashCode function. Changed in newer version.

aslesarenko (approved):

jozanek:
@aslesarenko any help I can offer to push it into master?

PR #790, opened 2022-04-04
Finalize v5.0: fix flatMap (part 2)

aslesarenko:
Removed constraint on lambda of flatMap (any lambda will be supported in v5.0).

### DIFF
--- a/sigmastate/src/main/scala/sigmastate/types.scala
+++ b/sigmastate/src/main/scala/sigmastate/types.scala
@@ -1688,9 +1688,6 @@ object SCollection extends STypeCompanion with MethodByNameUnapply {
     */
   def flatMap_eval[A, B](mc: MethodCall, xs: Coll[A], f: A => Coll[B])
                         (implicit E: ErgoTreeEvaluator): Coll[B] = {
-    if (!VersionContext.current.isJitActivated) {
-      checkValidFlatmap(mc)
-    }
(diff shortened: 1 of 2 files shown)


sigmastate/src/main/scala/sigmastate/types.scala
```diff
@@ -1688,9 +1688,6 @@ object SCollection extends STypeCompanion with MethodByNameUnapply {
     */
   def flatMap_eval[A, B](mc: MethodCall, xs: Coll[A], f: A => Coll[B])
                         (implicit E: ErgoTreeEvaluator): Coll[B] = {
-    if (!VersionContext.current.isJitActivated) {
-      checkValidFlatmap(mc)
```
kushti, line 1692: Why check for pre-JIT interpretation removed? How this change is non-breaking?

aslesarenko: This code (flatMap_eval) belong to new v5.0 interpreter, it is not used in v4.x.
Will be used after v5.0 activation.

In the previous version of v5.0 I implemented the same constraints on flatMap as in v4.x.
The intention was to make ergoTree.version based versioning.
But, as in the other previous cases, tree based versioning was replaced with activation-based.
And, as the next step, the limitations where removed, because they don't make sense with JITC.

PR #795, opened 2022-05-06
[v5.0.1] Remove v4.x implementation from Interpreter

aslesarenko:
In this PR the Interpreter is cleaned up from v4.x AOT costing and old reduction operations. Thus, only v5.0 implementation remains. 

The tests are adjusted to new cost values, but should remain exactly the same.

Note, the old AOT code will be cleaned up in separate PR.


sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -625,12 +480,11 @@ object Interpreter {
 
   /** Result of fullReduction to sigma tree with costing. */
   case class FullReductionResult(
```
kushti, line 482: This class looks like excessive meaningless wrapping now

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
```diff
@@ -1,7 +1,6 @@
 package sigmastate.interpreter
```
kushti, line 2: A lot of unused imports now

kushti (approved):
Why expected costs changed in tests? Please explain

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
aslesarenko: Will be removed in subsequent PRs

aslesarenko:
> Why expected costs changed in tests? Please explain

Explained this in ScalaDoc here https://github.com/ScorexFoundation/sigmastate-interpreter/blob/74d49fb171ff8236510f283d5e22c703c716985f/sigmastate/src/test/scala/special/sigma/SigmaDslSpecification.scala#L49

sigmastate/src/main/scala/sigmastate/interpreter/Interpreter.scala
aslesarenko: Cleaned up.

PR #802, opened 2022-05-15
[v5.x] Refactor tests

aslesarenko:
In this PR:
- [x] test harness updated to reflect removal of v4.x interpreter
- [x] more test cases added to cover possible MethodCalls in ErgoTree for some operations
- [x] more unused code removed

PR #803, opened 2022-05-16
[v5.0.1] Lower MethodCalls in TreeBuilding

### LINKED ISSUE #799: Lower MethodCalls in TreeBuilding
In particular more context properties can be recognized in the following code. 
As result the corresponding specialized operations would be used instead of MethodCalls in the resulting tree.

```scala
  object IsContextProperty {
    def unapply(d: Def[_]): Option[SValue] = d match {
      case ContextM.HEIGHT(_) => Some(Height)
      case ContextM.INPUTS(_) => Some(Inputs)
      case ContextM.OUTPUTS(_) => Some(Outputs)
      case ContextM.SELF(_) => Some(Self)
      case _ => None
    }
  }
```

jozanek:
Closes #799 
- [ ] parameterize expected expressions in tests using lowerMethodCallsInTests (see other tests for examples)
- [ ] add case for LastBlockUtxoRootHash to unlowerMethodCalls method
- [ ] fix tests

### DIFF
--- a/sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
+++ b/sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
@@ -299,6 +299,7 @@ case object Outputs extends LazyCollection[SBox.type] with FixedCostValueCompani
   }
 }
 
+// TODO: Should be copied to trees.scala?
 /** When interpreted evaluates to a AvlTreeConstant built from Context.lastBlockUtxoRoot */
 case object LastBlockUtxoRootHash extends NotReadyValueAvlTree with ValueCompanion {
   override def companion = this
--- a/sigmastate/src/main/scala/sigmastate/lang/SigmaCompiler.scala
+++ b/sigmastate/src/main/scala/sigmastate/lang/SigmaCompiler.scala
@@ -4,8 +4,7 @@ import fastparse.core.Parsed
 import fastparse.core.Parsed.Success
 import org.bitbucket.inkytonik.kiama.rewriting.Rewriter.{everywherebu, rewrite, rule}
 import org.ergoplatform.ErgoAddressEncoder.NetworkPrefix
-import org.ergoplatform.Global
-import org.ergoplatform.LastBlockUtxoRootHash
+import org.ergoplatform.{Context, Global, LastBlockUtxoRootHash}
 import sigmastate.Values.{SValue, Value}
 import sigmastate.eval.IRContext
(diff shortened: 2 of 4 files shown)


aslesarenko (commented):
@jozanek Please rebase on top of #802 
This will be merged into v5.0.1-RC branch after #802.
Also see subtasks in description.

sigmastate/src/main/scala/org/ergoplatform/ErgoLikeContext.scala
```diff
@@ -299,6 +299,7 @@ case object Outputs extends LazyCollection[SBox.type] with FixedCostValueCompani
   }
 }
 
+// TODO: Should be copied to trees.scala?
```
aslesarenko, line 302: No, refactoring will be done later after AOT cleanup.

sigmastate/src/main/scala/sigmastate/lang/SigmaCompiler.scala
```diff
@@ -124,6 +125,10 @@ class SigmaCompiler(settings: CompilerSettings) {
         MethodCall(xs,
           SCollection.GetOrElseMethod.withConcreteTypes(Map(tIV -> xs.tpe.elemType)),
           Vector(index, default), Map())
+      case LastBlockUtxoRootHash =>
+        // Dummy implementation to satisfy compiler, current one is commented below:
+        MethodCall(Global, SGroupElement.MultiplyMethod, Vector(), Map())
+        // MethodCall.typed[Value[SAvlTree.type]]( ValUse(1, SContext), SContext.lastBlockUtxoRootHashMethod, Vector(), Map())
```
aslesarenko, line 131: use this `MethodCall(Context, SContext.lastBlockUtxoRootHashMethod, Vector(), Map())`

sigmastate/src/test/scala/special/sigma/SigmaDslSpecification.scala
```diff
@@ -5144,26 +5144,47 @@ class SigmaDslSpecification extends SigmaDslTesting
...
+                Vector(),
+                Map()
+              ),
+              SAvlTree.getMethodByName("isUpdateAllowed"),
+              Vector(),
+              Map()
+            )
+          )),
+        preGeneratedSamples = Some(samples))
+    }
```
aslesarenko, line 5187: Look how I did parameterization without unnecessary duplication. Only the expected tree depends on lowerMethodCallsInTests.

aslesarenko (changes requested):

PR #807, opened 2022-05-24
[v5.0.1] Update dependencies towards Scala 2.13

### LINKED ISSUE #806: Cross-compilation to Scala 2.13
We need to cross-compile to both Scala 2.11 and Scala 2.13 (and optionally 2.12). Scala 2.11 is necessary for Android and RoboVM (the later for iOS). 
Sigma unfortunately have diverse dependencies which need to be updated.

Due to Scala 2.11 requirements, there is no simple recipe. Updating to the latest version is often not possible. 
This means that before starting migration, a migration plan should be prepared with the analysis of the situation for each individual dependency.
The migration should be done in v5.0.1 branches (i.e. it will be post v5.0 releases).

aslesarenko:
Closes #806 
This PR updates dependencies to simplify adding Scala 2.13 cross compilation.


aslesarenko:
Subsumed in #815

PR #808, opened 2022-05-28
Update Dep: 2.13 changes

kii-dot:


common/src/main/scala/scalan/util/CollectionUtil.scala
```diff
@@ -222,8 +221,8 @@ object CollectionUtil {
       res
     }
 
-    def filterCast[B:ClassTag](implicit cbf: CanBuildFrom[Source[A], B, Source[B]]): Source[B] = {
-      val b = cbf()
+    def filterCast[B:ClassTag](implicit cbf: Factory[B, Source[B]]): Source[B] = {
+      val b = cbf.newBuilder
```
aslesarenko, line 225: scala.collection.compat.BuildFrom should be used. I added the missing dependency in #807 and migrated this file. Please rollback changes and use as example in other places.

common/src/test/scala/sigmastate/VersionTesting.scala
```diff
@@ -10,7 +10,7 @@ trait VersionTesting {
     * In the branch for v6.0 the new version 3 should be added so that the tests run for both.
     */
   protected val activatedVersions: Seq[Byte] =
-    (0 to VersionContext.MaxSupportedScriptVersion).map(_.toByte).toArray
+    (0 to VersionContext.MaxSupportedScriptVersion).map(_.toByte).toSeq
```
aslesarenko, line 13: the right way is to use `.toArray[Byte]` as it just cannot infer the type for some reason.

core/src/main/scala/scalan/Base.scala
```diff
@@ -467,7 +467,7 @@ abstract class Base { scalan: Scalan =>
       * and leave non-Ref items unchanged. */
     final def apply(xs: Seq[Any])(implicit o: Overloaded1): Seq[Any] = {
       val len = xs.length
-      if (len == 0) mutable.WrappedArray.empty
+      if (len == 0) Seq[Any]()
```
aslesarenko, line 470: mutable.WrappedArray should be replaces with scala.collection.compat.immutable.ArraySeq. 
See recent commits in #807 where I replaces some of them.

Also, keep in mind that Sigma is almost entirely is part of hotspot code. 
Thus performance is critical and every allocation counts (basically every line is executed millions of times).
While migrating, introducing new allocations is not acceptable.
Please read https://github.com/ScorexFoundation/sigmastate-interpreter/blob/develop/docs/perf-style-guide.md for details of why Seq() is bad, and also for other performance patterns used in Sigma.

core/src/main/scala/scalan/compilation/GraphVizExport.scala
```diff
@@ -131,7 +131,7 @@ trait GraphVizExport extends Base { self: Scalan =>
     case l: Long => s"${l}l"
     case arr: Array[_] => s"Array(${arr.toSeq.map(formatConst).mkString(", ")})"
     case seq: Seq[_] =>
-      s"${seq.stringPrefix}(${seq.map(formatConst).mkString(", ")})"
+      s"${seq.toString()}(${seq.map(formatConst).mkString(", ")})"
```
aslesarenko, line 134:
```suggestion
      s"Seq(${seq.map(formatConst).mkString(", ")})"
```

core/src/main/scala/scalan/staged/ProgramGraphs.scala
```diff
@@ -31,7 +31,7 @@ trait ProgramGraphs extends AstGraphs { self: Scalan =>
 
     override def boundVars = Nil
     override def isIdentity: Boolean = false
-    override def freeVars = mutable.WrappedArray.empty[Sym]
+    override def freeVars = Seq[Sym]()
```
aslesarenko, line 34: See above comment about Seq()

sigmastate/src/test/scala/sigmastate/helpers/TestingHelpers.scala
```diff
@@ -112,13 +112,13 @@ object TestingHelpers {
     * in our test cases
     */
   def createTransaction(outputCandidates: IndexedSeq[ErgoBoxCandidate]): ErgoLikeTransaction = {
-    new ErgoLikeTransaction(WrappedArray.empty, WrappedArray.empty, outputCandidates)
+    new ErgoLikeTransaction(IndexedSeq.empty, IndexedSeq.empty, outputCandidates)
```
aslesarenko, line 115: See above comment on how WrappedArray should be replaced.

aslesarenko (changes requested):

core/src/main/scala/scalan/Base.scala
k3ithl1m: Should i change all occurence of Array to ArraySeq?

aslesarenko:
Subsumed in #815

PR #816, opened 2022-06-30
Fix registers of Unit type

aslesarenko:
In this PR added missing case for Sized[Unit]


aslesarenko:
Won't be merged in v4.x and otherwise not necessary

PR #817, opened 2022-07-05
[v5.0] Support for registers with Unit value

aslesarenko:
In this PR:
- UnitConstant implemented as an interface to ConstantNode (the old UnitConstant had no serializer and couldn't be used in ErgoTree), however ContantNode((), SUnit) can be added in ErgoTree, but its processing will fails in v4.x interpreter because of the missing case in stypeToElem
- Tests cases added to check that spending boxes with Unit registers is not supported in v4.x, but will be supported after v5.0 is activated.


kushti (changes requested):
This PR contains not just tests, change misleading title and description. Describe changes done in RuntimeCosting. Actually, it seems changes there are consensus-breaking. Also, why `UnitConstantCode` is not used anymore? Where do you have tests for Unit serialization / deserialization?

kushti:
This PR is introducing consensus-breaking changes, nodes (running on 5.0 candidate branch `v5.0-finalize) `before this
would fail with `exception check failed on sigmastate.lang.exceptions.CosterException: Don't know how to convert SType SUnit to Elem (root cause: sigmastate.lang.exceptions.CosterException: Don't know how to convert SType SUnit to Elem)`
on example script from a test in this PR (`{ SELF.R4[Unit].isDefined }`)

At the moment of writing this comment, that would mean hard-fork in 5.0 testnet.

sigmastate/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
```diff
@@ -122,6 +134,55 @@ class BasicOpsSpecification extends SigmaTestingCommons
...
+        test("R1", env, ext,
+          script = "", /* means cannot be compiled
+                         the corresponding script is { SELF.R4[Unit].isDefined } */
+          ExtractRegisterAs[SUnit.type](Self, reg1)(SUnit).isDefined.toSigmaProp,
+          additionalRegistersOpt = Some(Map(
+            reg1 -> UnitConstant.instance
+          ))
+        )
+
+        // TODO frontend: implement missing Unit support in compiler
```
kushti, line 149: For many reasons, better to have a Github issue for such a task

sigmastate/src/main/scala/sigmastate/Values.scala
```diff
@@ -449,14 +449,13 @@ object Values {
...
-  case class UnitConstant() extends EvaluatedValue[SUnit.type] {
-    override def tpe = SUnit
-    val value = ()
-    override def companion: ValueCompanion = UnitConstant
-  }
-  object UnitConstant extends ValueCompanion {
-    override def opCode = UnitConstantCode
-    override def costKind = Constant.costKind
+  object UnitConstant  {
+    val instance = apply()
```
kushti, line 453: ScalaDoc missed

kushti (changes requested):

sigmastate/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
aslesarenko: created issue https://github.com/ScorexFoundation/sigmastate-interpreter/issues/820

sigmastate/src/main/scala/sigmastate/Values.scala
aslesarenko: done

kushti (approved):

PR #826, opened 2022-08-23
Fix RoboVM incompatibility

### LINKED ISSUE #825: RoboVM incompatibility
Latest appkit snapshot is not working on iOS any more because of Java 8 method used here:

https://github.com/ScorexFoundation/sigmastate-interpreter/blob/fada073b82a16a928c457693b888da4c0310aca6/library-impl/src/main/scala/special/collection/Helpers.scala#L30

aslesarenko:
Closes #825

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,8 @@ env:
 jobs:
   build:
     name: Test and publish a snapshot
+    env:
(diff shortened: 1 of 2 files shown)


MrStahlfelge:
Reducing transaction works on iOS again now, thank you.
When we have a new release, we need to update the dependency in ergoplatform/ergo-appkit#189

MrStahlfelge (approved):

PR #831, opened 2022-10-10
V4.x pretty ergo tree

jozanek:


jozanek:
@aslesarenko how can we make this merged into master?

PR #835, opened 2022-10-22
Release Candidate v5.0.1

### LINKED ISSUE #836: 5.0.0 interpreter introducing breaking changes in deserialization of a Tuple instance stored in register
5.0.0 serialization can't deserialize Tuple value stored in register as Tuple is not EvaluatedValue anymore. This is causing disagreement between 5.0.0 and previous versions

### LINKED ISSUE #837: Breaking serialization changes in 5.0.0 caused by switching from getUInt.toInt to getUIntExact
In 5.0.0 in many places `.getUInt.toInt` was replaced with `.getUIntExact`, but this change could be breaking, the following example shows it:

```

case object ByPassSerializer extends ValueSerializer[IntConstant] {
  override def opDesc = ???

  override def serialize(obj: IntConstant, w: SigmaByteWriter): Unit = {
    w.putUInt(Int.MaxValue.toLong + 1)
  }

  override def parse(r: SigmaByteReader): IntConstant = {
    val i = r.getUInt.toInt
    IntConstant(i)
  }
}

object SerHack extends App {
  val w = SigmaSerializer.startWriter()
  ByPassSerializer.serialize(IntConstant(1), w)
  val r = SigmaSerializer.startReader(w.toBytes)
  val res = ByPassSerializer.parse(r)
  println(res)
  // prints IntConstant(-2147483648)
}
```

, while serializer with  `.getUIntExact` throws an exception.

Around most of places negative value would cause an exception anyway (e.g. `new Array(negValue)` will do it), however, I found a place where 4.x will not throw an exception unlike 5.0.0, and also there are some places to check whether negative value will throw an exception in 4.x:

* `deserializeConstants` in `ErgoTreeSerializer` :

4.x is using

```
      val nConsts = r.getUInt().toInt
      if (nConsts > 0) {
          // HOTSPOT:: allocate new array only if it is not empty
          val res = ValueSerializer.newArray[Constant[SType]](nConsts)
          cfor(0)(_ < nConsts, _ + 1) { i =>
            res(i) = constantSerializer.deserialize(r)
          }
          res
        }
        else
          Constant.EmptySeq
```

(formatting is not good btw) 

so it seems negative value will be just ignored. In 5.0.0 `val nConsts = r.getUIntExact` will throw an exception

I guess 5.0.0 change should revert the change, with a comment to be made on why negative value is OK. 

* `ValUseSerializer` - show that negative id 4.x could get in `val id = r.getUInt().toInt` will always fail validation after deserialization
* `FuncValueSerializer` - show the same for `val id = r.getUInt().toInt`

aslesarenko:
Fixes #836 
Fixes #837 
This PR contains fix for Tuple in register deserialization and other potentially breaking changes in deserializers.

### DIFF
--- a/sigmastate/src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
+++ b/sigmastate/src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
@@ -187,9 +187,11 @@ object ErgoBoxCandidate {
       r.positionLimit = r.position + ErgoBox.MaxBoxSize
       val value = r.getULong()                  // READ
       val tree = DefaultSerializer.deserializeErgoTree(r, SigmaSerializer.MaxPropositionSize)  // READ
-      val creationHeight = r.getUInt().toInt    // READ
-      // Note, when creationHeight < 0 as a result of Int overflow nothing happens here
-      // and ErgoBoxCandidate with negative creation height is created
+      val creationHeight = r.getUIntExact       // READ
+      // NO-FORK: ^ in v5.x getUIntExact may throw Int overflow exception
+      // in v4.x r.getUInt().toInt is used and may return negative Int instead of the overflow
+      // and ErgoBoxCandidate with negative creation height is created, which is then invalidated
+      // during transaction validation. See validation rule # 122 in the Ergo node (ValidationRules.scala)
       val nTokens = r.getUByte()                // READ
       val tokenIds = safeNewArray[Array[Byte]](nTokens)
       val tokenAmounts = safeNewArray[Long](nTokens)
--- a/sigmastate/src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
+++ b/sigmastate/src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
@@ -187,7 +187,9 @@ object ErgoBoxCandidate {
       r.positionLimit = r.position + ErgoBox.MaxBoxSize
       val value = r.getULong()                  // READ
       val tree = DefaultSerializer.deserializeErgoTree(r, SigmaSerializer.MaxPropositionSize)  // READ
-      val creationHeight = r.getUIntExact       // READ
+      val creationHeight = r.getUInt().toInt    // READ
+      // Note, when creationHeight < 0 as a result of Int overflow nothing happens here
+      // and ErgoBoxCandidate with negative creation height is created
       val nTokens = r.getUByte()                // READ
       val tokenIds = safeNewArray[Array[Byte]](nTokens)
       val tokenAmounts = safeNewArray[Long](nTokens)
@@ -195,6 +197,9 @@ object ErgoBoxCandidate {
         val nDigests = digestsInTx.length
         cfor(0)(_ < nTokens, _ + 1) { i =>
           val digestIndex = r.getUIntExact    // READ
+          // NO-FORK: in v5.x getUIntExact throws Int overflow exception
+          // in v4.x r.getUInt().toInt is used and may return negative Int in which case
+          // the error below is thrown
           if (digestIndex < 0 || digestIndex >= nDigests)
             sys.error(s"failed to find token id with index $digestIndex")
           val amount = r.getULong()           // READ
--- a/sigmastate/src/main/scala/org/ergoplatform/validation/SigmaValidationSettingsSerializer.scala
+++ b/sigmastate/src/main/scala/org/ergoplatform/validation/SigmaValidationSettingsSerializer.scala
@@ -19,7 +19,9 @@ object SigmaValidationSettingsSerializer extends SigmaSerializer[SigmaValidation
   }
 
   override def parse(r: SigmaByteReader): SigmaValidationSettings = {
-    val nRules = r.getUIntExact
+    val nRules = r.getUInt().toInt
+    // Note, when nRules < 0 as a result of Int overflow, the loop is empty
+    // deserialization will likely fail elsewhere
     val parsed = (0 until nRules).map { _ =>
       val ruleId = r.getUShort().toShortExact
       val status = RuleStatusSerializer.parse(r)
--- a/sigmastate/src/main/scala/org/ergoplatform/validation/SigmaValidationSettingsSerializer.scala
+++ b/sigmastate/src/main/scala/org/ergoplatform/validation/SigmaValidationSettingsSerializer.scala
@@ -4,6 +4,7 @@ import sigmastate.serialization.SigmaSerializer
 import sigmastate.utils.{SigmaByteReader, SigmaByteWriter}
 import scalan.util.Extensions.{IntOps,LongOps}
 
+// TODO v5.x: remove unused class and related json encoders
 /** The rules are serialized ordered by ruleId.
   * This serializer preserves roundtrip identity `deserialize(serialize(_)) = identity`
   * however it may not preserve `serialize(deserialize(_)) = identity` */
@@ -21,7 +22,6 @@ object SigmaValidationSettingsSerializer extends SigmaSerializer[SigmaValidation
   override def parse(r: SigmaByteReader): SigmaValidationSettings = {
     val nRules = r.getUInt().toInt
     // Note, when nRules < 0 as a result of Int overflow, the loop is empty
-    // deserialization will likely fail elsewhere
     val parsed = (0 until nRules).map { _ =>
       val ruleId = r.getUShort().toShortExact
       val status = RuleStatusSerializer.parse(r)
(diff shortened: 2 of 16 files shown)


sigmastate/src/main/scala/org/ergoplatform/validation/SigmaValidationSettingsSerializer.scala
```diff
@@ -19,7 +19,9 @@ object SigmaValidationSettingsSerializer extends SigmaSerializer[SigmaValidation
   }
 
   override def parse(r: SigmaByteReader): SigmaValidationSettings = {
-    val nRules = r.getUIntExact
+    val nRules = r.getUInt().toInt
+    // Note, when nRules < 0 as a result of Int overflow, the loop is empty
+    // deserialization will likely fail elsewhere
```
kushti, line 24: Such comment is confusing, "likely fail elsewhere" does not really says anything, please specify how a transaction where a script has negative id encoded will fail. It it will fail, no need to revert getUIntExact

sigmastate/src/main/scala/sigmastate/AvlTreeData.scala
```diff
@@ -92,8 +92,10 @@ object AvlTreeData {
     override def parse(r: SigmaByteReader): AvlTreeData = {
       val digest = r.getBytes(DigestSize)
       val tf = AvlTreeFlags(r.getByte())
-      val keyLength = r.getUIntExact
-      val valueLengthOpt = r.getOption(r.getUIntExact)
+      val keyLength = r.getUInt().toInt
+      val valueLengthOpt = r.getOption(r.getUInt().toInt)
+      // Note, when keyLength and valueLengthOpt < 0 as a result of Int overflow,
+      // the deserializer succeeds with invalid AvlTreeData
```
kushti, line 98: Will invalid AvlTreeData fail validation after? If so , why not to keep  `getUIntExact` ?

sigmastate/src/main/scala/sigmastate/serialization/ValDefSerializer.scala
```diff
@@ -30,7 +30,9 @@ case class ValDefSerializer(override val opDesc: ValueCompanion) extends ValueSe
   }
 
   override def parse(r: SigmaByteReader): Value[SType] = {
-    val id = r.getUIntExact
+    val id = r.getUInt().toInt
+    // Note, when id < 0 as a result of Int overflow, the r.valDefTypeStore(id) won't throw
+    // but this will likely fail elsewhere
```
kushti, line 35: It will fail on  require(id >= 0, "id must be >= 0") in class ValDef

sigmastate/src/main/scala/sigmastate/serialization/ValUseSerializer.scala
```diff
@@ -12,7 +12,9 @@ case class ValUseSerializer(cons: (Int, SType) => Value[SType]) extends ValueSer
   }
 
   override def parse(r: SigmaByteReader): Value[SType] = {
-    val id = r.getUIntExact
+    val id = r.getUInt().toInt
+    // Note, when id < 0 as a result of Int overflow, the r.valDefTypeStore(id) won't throw
+    // but this will likely fail elsewhere
```
kushti, line 17: Such comment is confusing, "likely fail elsewhere" does not really says anything, please specify how a transaction where a script has negative id encoded will fail. It it will fail, no need to revert getUIntExact

sigmastate/src/main/scala/sigmastate/serialization/FuncValueSerializer.scala
```diff
@@ -28,9 +28,14 @@ case class FuncValueSerializer(cons: (IndexedSeq[(Int, SType)], Value[SType]) =>
...
     val argsSize = r.getUIntExact
+    // NO-FORK: in v5.x getUIntExact may throw Int overflow exception
+    // in v4.x r.getUInt().toInt is used and may return negative Int instead of the overflow
+    // in which case the array allocation will throw NegativeArraySizeException
     val args = safeNewArray[(Int, SType)](argsSize)
     cfor(0)(_ < argsSize, _ + 1) { i =>
-      val id = r.getUIntExact
+      val id = r.getUInt().toInt
+      // Note, when id < 0 as a result of Int overflow, the r.valDefTypeStore(id) won't throw
+      // but this will likely fail elsewhere
```
kushti, line 38: Such comment is confusing, "likely fail elsewhere" does not really says anything, please specify how a transaction where a script has negative id encoded will fail. It it will fail, no need to revert getUIntExact

sigmastate/src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala
```diff
@@ -207,7 +207,12 @@ class ErgoTreeSerializer {
     val header = r.getByte()
     CheckHeaderSizeBit(header)
     val sizeOpt = if (ErgoTree.hasSize(header)) {
-      Some(r.getUIntExact)
+      val size = r.getUInt().toInt
+      // Note, when size < 0 as a result of Int overflow nothing happens here and later
+      // when deserialization proceeds normally as sizeOpt is not used on this pass.
+      // However, when ValidationException is thrown in deserializeErgoTree this negative
+      // size value will lead to undefined behavior
```
kushti, line 214: What does undefined behavior means? Why can't it be defined by observing and reasoning on usages of the method ?

kushti (changes requested):

sigmastate/src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
```diff
@@ -187,14 +187,19 @@ object ErgoBoxCandidate {
...
+      // Note, when creationHeight < 0 as a result of Int overflow nothing happens here
+      // and ErgoBoxCandidate with negative creation height is created
       val nTokens = r.getUByte()                // READ
       val tokenIds = safeNewArray[Array[Byte]](nTokens)
       val tokenAmounts = safeNewArray[Long](nTokens)
       if (digestsInTx != null) {
         val nDigests = digestsInTx.length
         cfor(0)(_ < nTokens, _ + 1) { i =>
           val digestIndex = r.getUIntExact    // READ
+          // NO-FORK: in v5.x getUIntExact throws Int overflow exception
```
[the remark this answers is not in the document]
aslesarenko: Well, for completeness. I will leave it as history trace.

sigmastate/src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala
```diff
@@ -187,14 +187,19 @@ object ErgoBoxCandidate {
       r.positionLimit = r.position + ErgoBox.MaxBoxSize
       val value = r.getULong()                  // READ
       val tree = DefaultSerializer.deserializeErgoTree(r, SigmaSerializer.MaxPropositionSize)  // READ
-      val creationHeight = r.getUIntExact       // READ
+      val creationHeight = r.getUInt().toInt    // READ
+      // Note, when creationHeight < 0 as a result of Int overflow nothing happens here
+      // and ErgoBoxCandidate with negative creation height is created
```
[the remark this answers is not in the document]
aslesarenko: fixed

sigmastate/src/main/scala/org/ergoplatform/validation/SigmaValidationSettingsSerializer.scala
aslesarenko: Hm, it turns out this serializer is only used by json encoders and not used in Ergo node.
I suggest to leave it as it is (removed confusing "// deserialization will likely fail elsewhere" part)
Added TODO v5.x

sigmastate/src/main/scala/sigmastate/AvlTreeData.scala
aslesarenko: Some AvlTree operations (remove_eval, update_eval, contains_eval) won't throw. Added comment in this code and also added test.

sigmastate/src/main/scala/sigmastate/serialization/ValDefSerializer.scala
aslesarenko: good catch, fixed

sigmastate/src/main/scala/sigmastate/serialization/ValUseSerializer.scala
aslesarenko: Added comment.

sigmastate/src/main/scala/sigmastate/serialization/FuncValueSerializer.scala
aslesarenko: clarified in comment

sigmastate/src/main/scala/sigmastate/serialization/ErgoTreeSerializer.scala
aslesarenko: added more comment in code

kushti (approved):

PR #839, opened 2022-11-03
ErgoBoxCandidate.tokens does not account for same token multiple times in array

### LINKED ISSUE #838: ErgoBoxCandidate.tokens does not account for same token multiple times in array
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/251784a9f7c1b325c4859fe256c9fe3862fffe4e/sigmastate/src/main/scala/org/ergoplatform/ErgoBoxCandidate.scala#L105

This fails when a token is added into the array multiple times.

Causing ergoplatform/ergo-appkit#199

MrStahlfelge:
closes #838


kushti:
@MrStahlfelge @aslesarenko from initial investigation, `tokens` used in the protocol client in wallet part only, so there should be no consensus-critical issues with this fix

aslesarenko (approved):

kushti (approved):

PR #842, opened 2022-11-15
Export modules individually

anon-br:
This PR modifies package.json in order to allow individual modules exporting while ignores all `internal-*` files.

Modules can be imported as follows:

```js
import { Value, Values } from "sigmastate-js/core";
import { ErgoTrees } from "sigmastate-js/ergotree";
```

This is good because allows bundlers like webpack to "[tree-shake](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking)" the code, when possible, and extract only what's needed for production.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,10 @@
   "files": [
     "interpreter/js/target/scala-2.13/interpreter-fastopt/"
   ],
+  "exports": {
+    "./internal-*": null,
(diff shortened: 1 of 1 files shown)


aslesarenko:
@anon-br, but those internal files are `required` in the modules, so they need to be uploaded.

aslesarenko (approved):

PR #848, opened 2022-12-08
LangSpec.md updated + more ScalaDocs

aslesarenko:


docs/LangSpec.md
```diff
@@ -426,7 +435,7 @@ The following syntax is supported to access registers on box objects:
 ```
 box.R3[Int].get          // access R3 register, check that its value of type Int and return it
 box.R3[Int].isDefined    // check that value of R3 is defined and has type Int
-box.R3[Int].getOrElse(d) // access R3 value if defined, otherwise return `d`
+box.R3[Int].getOrElse(d) // access R3 register, check that its value of type Int, return it if defined, otherwise return `d`
```
kushti, line 438: What if it is defined but not of type Int?

kushti (changes requested):
What about other changes from ? https://github.com/ScorexFoundation/sigmastate-interpreter/milestone/9?closed=1 

See no changes in regards with Unit, flatmap , zip/append , xor,

aslesarenko:
> What about other changes from ? https://github.com/ScorexFoundation/sigmastate-interpreter/milestone/9?closed=1 
> See no changes in regards with Unit, flatmap , zip/append , xor,

All the changes are invisible for developers, because all of them fix bugs so that after activation the operations work as expected (and documented).
E.g. `zip` used to produce invalid result for collections of different length, and `append` failed with exception in some situation.

kushti:
@aslesarenko see no .propBytes property of SigmaProp type

kushti (approved):

PR #853, opened 2023-01-16
[v5.x] Clean up parsers + more ScalaDocs

aslesarenko:
In this PR ErgoScript parser implementation was cleaned-up and documented.


aslesarenko:
Merged into v5.0.8

PR #856, opened 2023-01-25
[v5.x] Remove dependencies, part1

aslesarenko:
In this PR:
- code simplified and unnecessary classes removed
- java dependencies removed commons-io, commons-math, junit-interface
- dependence on kiama removed (minimally required code copied to sigma)
- dependency on Logger replaced with a simple method (see #859)


sigmastate/src/main/scala/sigmastate/interpreter/ProverInterpreter.scala
```diff
@@ -556,7 +555,7 @@ trait ProverInterpreter extends Interpreter with ProverUtils {
     case ut: UnprovenTree => ut
 
     case t: ProofTree =>
-      log.warn("Wrong input in prove(): ", t);
+      logMessage(s"Wrong input in prove(): $t");
```
kushti, line 558: Why unconditionally dumping to console instead of using flexible logger (with logging levels, different output options, configurable etc). Does not look like a good change.

kushti (changes requested):

common/src/main/scala/scalan/util/Extensions.scala
```diff
@@ -216,4 +215,11 @@ object Extensions {
         None
     }
   }
+
+  implicit final class Ensuring[A](private val self: A) extends AnyVal {
```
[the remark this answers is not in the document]
aslesarenko: fixed

sigmastate/src/main/scala/sigmastate/interpreter/ProverInterpreter.scala
aslesarenko: This is because there is no good dependency for logging which support Scala 2.11-2.13 and also Scala.js. 
Also sigma should be non-opinionated library so that it can be reused in different context (including the very constrained JS environment).
So the simplest way to abstract is just call some method and let the library users to override the default behavior.

kushti: In this case better to extract logging functions into a separate interface likely

aslesarenko: This is out of this PR's scope. Created an issue https://github.com/ScorexFoundation/sigmastate-interpreter/issues/859

kushti: the change is also out of PR scope seemingly, not corresponding to PR description. Consider to move it to corresponding PR

kushti (approved):

sigmastate/src/main/scala/sigmastate/interpreter/ProverInterpreter.scala
aslesarenko: Updated description

PR #857, opened 2023-01-27
Implement ContractTemplate data structures according to EIP5

### LINKED ISSUE #851: ContractTemplate data structure according to EIP5
**Goal:**
To implement representation of templates, serialization and generation of ErgoTree from template by substituting parameters.

**Non-goal:**
Implement compiler from source code to template representation.

- [ ] define ContractTemplate classes, serializers and JSON encoders according to EIP
- [ ] cover with unit tests
- [ ] implement applyTemplate(p1, p2, p3) -> ErgoTree
- [ ] cover `applyTemplate` method with tests

darkdrag00nv2:
Resolves #851 

Added support for ContractTemplate data structures according to EIP-5

As part of this work, I've done the following:
-  Moved `SigmaParser` to a new module - #864 
- Moved `DataJsonEncoder` and `JsonCodecs` to the `sdk` module - #867 
- Implemented the data structures, serialization, json conversion & `applyTemplate` for ContractTemplate - this PR


sigmastate/src/main/scala/sigmastate/serialization/ContractTemplateSerializer.scala
```diff
@@ -0,0 +1,144 @@
...
+        res
+      } else {
+        Parameter.EmptySeq
+      }
+    }
+
+    // TODO: Here we need to create the list of constants (IndexedSeq[Constant[T]]) and populate it in the constant
+    // store as r.constantStore so that the expressionTree which contains ConstantPlaceholder's can be deserialized.
+    // Unfortunately, the value part of the constant is not always known here and can be None here.
+//    constTypes.indices.map(i => DeserializationSigmaBuilder.mkConstant(null, constTypes(i)))
```
darkdrag00nv2, line 125: @aslesarenko Need your input here. 

In order to deserialize the `expressionTree` of the ContractTemplate, we need to have all the constants defined in the `constantStore`. This is because the serialized `expressionTree` can have `ConstantPlaceholder` in the tree. The deserializer for the placeholder looks up the constant from the constantStore during the deserialization process - [code link](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/dab64a3c8b4568f3e84adbb48842047c9c0f1f7f/sigmastate/src/main/scala/sigmastate/serialization/ConstantPlaceholderSerializer.scala#L21)

For some of the parameters, the value in `constValues` will be `None`. Hence, it is not possible to populate constants in the constantStore since we don't know the value of them.

Basically, we need to be able to do something like `DeserializationSigmaBuilder.mkConstant(null, constTypes(i))` i.e. set the value as `null` but I don't think that's possible.

.bsp/sbt.json
```diff
@@ -0,0 +1 @@
+{"name":"sbt","version":"1.4.5","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["/usr/lib/jvm/java-11-openjdk-amd64/bin/java","-Xms100m","-Xmx100m","-classpath","/home/darkdrag00n/.local/share/JetBrains/IdeaIC2022.3/Scala/launcher/sbt-launch.jar","xsbt.boot.Boot","-bsp"]}
```
aslesarenko, line 1: Do we really need this file committed?

sigmastate/src/main/scala/sigmastate/Values.scala
```diff
@@ -1522,4 +1523,121 @@ object Values {
...
+  object ContractTemplate {
+    def apply(name: String,
+              description: String,
+              constTypes: IndexedSeq[SType],
+              constValues: Option[IndexedSeq[Option[SType#WrappedType]]],
+              parameters: IndexedSeq[Parameter],
+              expressionTree: SigmaPropValue): ContractTemplate = {
+      new ContractTemplate(None, name, description, constTypes, constValues, parameters, Right(expressionTree))
+    }
+  }
```
aslesarenko, line 1642: Please rebase on top of https://github.com/ScorexFoundation/sigmastate-interpreter/compare/sdk-separate-interpreters
and move all the new code into `sdk` module.

sigmastate/src/main/scala/sigmastate/Values.scala
```diff
@@ -1522,4 +1523,121 @@ object Values {
...
+    constTypes: IndexedSeq[SType],
+    constValues: Option[IndexedSeq[Option[SType#WrappedType]]],
+    parameters: IndexedSeq[Parameter],
+    expressionTree: Either[UnparsedContractTemplate, SigmaPropValue],
+    private val templateBytes: Array[Byte],
+ ) {
+
+    require(constValues.isEmpty || constValues.get.size == constTypes.size,
+      s"constValues must be empty or of same length as constTypes. Got ${constValues.get.size}, expected ${constTypes.size}")
+    require(parameters.size <= constTypes.size, "number of parameters must be <= number of constants")
```
aslesarenko, line 1559: can also add requirement on parameters(i).placeholder.

sigmastate/src/main/scala/sigmastate/Values.scala
```diff
@@ -1522,4 +1523,121 @@ object Values {
...
+
+  object Parameter {
+    /** Immutable empty IndexedSeq, can be used to save allocations in many places. */
+    val EmptySeq: IndexedSeq[Parameter] = Array.empty[Parameter]
+  }
+
+
+  /** This is alternative representation of ContractTemplate when it cannot be parsed
+    * due to `error`. */
+  case class UnparsedContractTemplate(bytes: mutable.WrappedArray[Byte], error: ValidationException)
```
aslesarenko, line 1544: I don't think we need to represent unparsable templates. This feature in ErgoTree was added because deserialization there should always succeed to implement soft-fork logic. Here it is not necessary.

sigmastate/src/main/scala/sigmastate/Values.scala
```diff
@@ -1522,4 +1523,121 @@ object Values {
...
+      s"constValues must be empty or of same length as constTypes. Got ${constValues.get.size}, expected ${constTypes.size}")
+    require(parameters.size <= constTypes.size, "number of parameters must be <= number of constants")
+
+    def this(treeVersion: Option[Byte],
+             name: String,
+             description: String,
+             constTypes: IndexedSeq[SType],
+             constValues: Option[IndexedSeq[Option[SType#WrappedType]]],
+             parameters: IndexedSeq[Parameter],
+             expressionTree: Either[UnparsedContractTemplate, SigmaPropValue]) =
```
aslesarenko, line 1567: Either is not necessary.

sigmastate/src/main/scala/sigmastate/Values.scala
```diff
@@ -1522,4 +1523,121 @@ object Values {
...
+
+    /** Serialized bytes of this template. */
+    final def bytes: Array[Byte] = {
+      if (_bytes == null) {
+        _bytes = ContractTemplateSerializer.DefaultSerializer.serializeContractTemplate(this)
+      }
+      _bytes
+    }
+
+    def applyTemplate(paramValues: IndexedSeq[Constant[SType]]): ErgoTree = {
```
aslesarenko, line 1587: applyTemplate should take Map[String, Constant] as argument, so parameters a substituted by name.

sigmastate/src/main/scala/sigmastate/Values.scala
```diff
@@ -1522,4 +1523,121 @@ object Values {
...
+
+  object ContractTemplate {
+    def apply(name: String,
+              description: String,
+              constTypes: IndexedSeq[SType],
+              constValues: Option[IndexedSeq[Option[SType#WrappedType]]],
+              parameters: IndexedSeq[Parameter],
+              expressionTree: SigmaPropValue): ContractTemplate = {
+      new ContractTemplate(None, name, description, constTypes, constValues, parameters, Right(expressionTree))
+    }
```
aslesarenko, line 1641: This probably not necessary once you remove Either.

sigmastate/src/main/scala/sigmastate/serialization/ContractTemplateJsonEncoder.scala
```diff
@@ -0,0 +1,71 @@
...
+      }
+    }
+    val name = json.hcursor.downField("name").focus.get.asString.get
+    val description = json.hcursor.downField("description").focus.get.asString.get
+
+    val tpe = SigmaParser.parseType(json.hcursor.downField("type").focus.get.asString.get)
+
+//    ContractTemplate(treeVersionOpt, name, description)
+  }
+}
```
aslesarenko, line 71: please use JsonCodecs as examples (which is based on circe).
You can use some helpers from ther at well.
But don't add this new code there, instead define it in ContractTemplate companion object.

sigmastate/src/main/scala/sigmastate/serialization/ContractTemplateSerializer.scala
```diff
@@ -0,0 +1,144 @@
...
+      if (nParameters > 0) {
+        val res = safeNewArray[Parameter](nParameters)
+        cfor(0)(_ < nParameters, _ + 1) { i =>
+          res(i) = ParameterSerializer.deserialize(r)
+        }
+        res
+      } else {
+        Parameter.EmptySeq
+      }
+    }
```
aslesarenko, line 120: You need to inherit SigmaSerializer (see SigmaBoolean as an example of serializer implementation).
Put serializers in companion object of the corresponding class.
Template serializer should have just 2 methods as in SigmaBoolean, there is no need to design it after ErgoTree serializers, because they are too complex because of optimization and various use cases. 
templates instead should be simple (write all or read all or throw an SerializerException)

sigmastate/src/main/scala/sigmastate/serialization/ContractTemplateSerializer.scala
```diff
@@ -0,0 +1,144 @@
...
+
+    ContractTemplate(
+      treeVersion, name, description,
+      constTypes, constValues, parameters,
+      Right(expressionTree.toSigmaProp), templateBytes)
+  }
+}
+
+object ContractTemplateSerializer {
+  val DefaultSerializer = new ContractTemplateSerializer
```
aslesarenko, line 143: use SigmaBoolean as an example, not ErgoTree.

sigmastate/src/main/scala/sigmastate/serialization/ParameterSerializer.scala
```diff
@@ -0,0 +1,29 @@
...
+import sigmastate.utils.{SigmaByteReader, SigmaByteWriter}
+
+import java.nio.charset.StandardCharsets
+
+object ParameterSerializer {
+  import sigmastate.Operations.ConstantPlaceholderInfo._
+
+  def serialize[T <: SType](parameter: Parameter, w: SigmaByteWriter): Unit = {
+    w.putUInt(parameter.name.length)
+    w.putBytes(parameter.name.getBytes(StandardCharsets.UTF_8))
```
aslesarenko, line 14: parameter.name.length is not always the number of bytes.

sigmastate/src/test/scala/sigmastate/serialization/ContractTemplateSerializerSpecification.scala
```diff
@@ -0,0 +1,71 @@
...
+  private def extractConstants(prop: SigmaPropValue)(implicit IR: IRContext): (IndexedSeq[Values.Constant[SType]], SigmaPropValue) = {
+    val env = Map[String, Any]()
+    val res = compiler.compileTyped(env, prop)
+    checkCompilerResult(res)
+    val calcF = res.compiledGraph
+    val constantsStore = new ConstantStore()
+    val outExpr = IR.buildTree(calcF, Some(constantsStore))
+    val constants = constantsStore.getAll
+    (constants, outExpr)
+  }
```
aslesarenko, line 38: You should never need the compiler for template operations. Compiler is only used for text => ErgoTree transformation and we don't have to do with `text` in templates.

aslesarenko (changes requested):

sigmastate/src/main/scala/sigmastate/Values.scala
darkdrag00nv2: @aslesarenko Need some help on this. I need some of the code from `sc` module in the `sdk` module. Example is `JsonCodecs`, `DataJsonEncoder`, `SigmaParser` etc.

I tried adding the dependency for `sc` in the build.sbt but it fails as `sc` is not a `crossProject`. Can you please advise on how to proceed here?

1. I can make `sc` a `crossProject`
2. I can move the needed logic outside to some other module - `corelib` or `interpreter` which are already a dependency for `sdk`.

Thanks!

aslesarenko: sdk module shouldn't depend on sc.
I think you can move JsonCodecs, DataJsonEncoder to sdk module.
But, it is better we keep SigmaParser out of sdk.
Do you really need it for eip5?

darkdrag00nv2: I need SigmaParser to decode SType. 

Basically this:
``` scala
implicit val sTypeDecoder: Decoder[SType] = Decoder.instance({ implicit cursor =>
      fromTry(Try.apply(SigmaParser.parseType(cursor.value.asString.get)))
    })
```

Also note that, `DataJsonEncoder` depends on `SigmaParser` for the same -- to decode the SType. See [here](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/f6db2bbf3f4ddcc07d18d23bc01f557c23290264/sc/src/main/scala/sigmastate/serialization/DataJsonEncoder.scala#L266-L283).

I thought about copying the `SigmaParser.parseType` functionality into `sdk` module but just copying [these 9-10 lines](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/f6db2bbf3f4ddcc07d18d23bc01f557c23290264/sc/src/main/scala/sigmastate/lang/SigmaParser.scala#L104-L112) isn't enough as it depends on the definition of `Type` which is present in `Types.scala` and the dependencies go on and on.

May be we should move the `SigmaParser` functionality into a shared module?

.bsp/sbt.json
darkdrag00nv2: Nope, accidentally pushed it. Removed it now.

sigmastate/src/main/scala/sigmastate/Values.scala
darkdrag00nv2: Sure, removed that part from the code.

sigmastate/src/main/scala/sigmastate/serialization/ContractTemplateJsonEncoder.scala
darkdrag00nv2: Done.

sigmastate/src/main/scala/sigmastate/serialization/ContractTemplateSerializer.scala
darkdrag00nv2: Done, added a `defaultOf` method.

sigmastate/src/main/scala/sigmastate/Values.scala
darkdrag00nv2: Fixed by #864 and #867

sigmastate/src/main/scala/sigmastate/Values.scala
darkdrag00nv2: Sure, added a few more restrictions in the `validate` method.

sigmastate/src/main/scala/sigmastate/Values.scala
darkdrag00nv2: Removed

sigmastate/src/main/scala/sigmastate/Values.scala
darkdrag00nv2: Yeah, updated the code.

sigmastate/src/main/scala/sigmastate/Values.scala
darkdrag00nv2: Still good to keep if we want to allow creating a ContractTemplate with version as None.

sigmastate/src/main/scala/sigmastate/serialization/ContractTemplateSerializer.scala
darkdrag00nv2: Sure, thanks for the pointer. It helped. Updated the code.

sigmastate/src/main/scala/sigmastate/serialization/ContractTemplateSerializer.scala
darkdrag00nv2: Done

sigmastate/src/main/scala/sigmastate/serialization/ParameterSerializer.scala
darkdrag00nv2: yeah, thanks. I didn't know. Have updated the code to convert to bytes based on UTF_8 encoding.

sigmastate/src/test/scala/sigmastate/serialization/ContractTemplateSerializerSpecification.scala
darkdrag00nv2: Yeah, I just wanted to use the compiler to extract the constants from the Tree. Not needed though and I've updated the code.

darkdrag00nv2:
This is mostly ready for review except that we should merge #867 first and I need to fix the Scala 2.11 compilation in the test.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+      // Populate constants in constantStore so that the expressionTree can be deserialized.
+      val constants = constTypes.indices.map(i => {
+        val t = constTypes(i)
+        DeserializationSigmaBuilder.mkConstant(defaultOf(t), t)
+      })
+      constants.foreach(c => r.constantStore.put(c)(DeserializationSigmaBuilder))
+
+      val _ = r.getUInt().toInt
+      val expressionTree = ValueSerializer.deserialize(r)
+      CheckDeserializedScriptIsSigmaProp(expressionTree)
```
aslesarenko, line 307: I suggest to not use ValidationRules. They are part of soft-forkability logic and better not to be used outside of that. Use simple type comparison directly here and throw serializer exception.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+
+    implicit val encoder: Encoder[ContractTemplate] = Encoder.instance({ ct =>
+      val expressionTreeWriter = SigmaSerializer.startWriter()
+      ValueSerializer.serialize(ct.expressionTree, expressionTreeWriter)
+
+      Json.obj(
+        "treeVersion" -> ct.treeVersion.asJson,
+        "name" -> Json.fromString(ct.name),
+        "description" -> Json.fromString(ct.description),
+        "constTypes" -> ct.constTypes.asJson,
```
aslesarenko, line 334: If sTypeEncoder is used here, than it is wrong. In EIP5 the string should be hex coming from TypeSerializer.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+            ValueSerializer.deserialize(r).toSigmaProp)
+        case _ => Left(DecodingFailure("Failed to decode contract template", cursor.history))
+      }
+    })
+  }
+
+  /** Synthetic default value for the type.
+    * Used for deserializing contract templates.
+    */
+  private def defaultOf[T <: SType](tpe: T): T#WrappedType = {
```
aslesarenko, line 392: There was this Zero typeclass which I removed. Looks like it can be brought back https://github.com/ScorexFoundation/sigmastate-interpreter/commit/a064a0f71f8a4f40e41496c64280bd7e7a63670c

aslesarenko (changes requested):

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+    require(paramIndices.size == parameters.size, "multiple parameters point to the same placeholder")
+
+    // Validate that any constValues[i] = None has a parameter.
+    if (constValues.isEmpty) {
+      require(parameters.size == constTypes.size,
+        "all the constants must be provided via parameter since constValues == None")
+    } else {
+      cfor(0)(_ < constTypes.size, _ + 1) { i =>
+        require(constValues.get(i).isDefined || paramIndices.contains(i),
+          s"placeholder ${i} does not have a default value and absent from parameter as well")
```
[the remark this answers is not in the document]
darkdrag00nv2: Change for EIP5 - https://github.com/ergoplatform/eips/pull/90

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    ErgoTree(
+      ErgoTree.ConstantSegregationHeader,
+      constants,
+      this.expressionTree
+    )
+  }
+
+  /** The default equality of case class is overridden to exclude `complexity`. */
```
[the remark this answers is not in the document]
darkdrag00nv2: Yeah, removed. Was a copy paste from `ErgoTree` class.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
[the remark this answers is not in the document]
darkdrag00nv2: Renamed in the PR as well.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+  private def validate(): Unit = {
+    require(constValues.isEmpty || constValues.get.size == constTypes.size,
+      s"constValues must be empty or of same length as constTypes. Got ${constValues.get.size}, expected ${constTypes.size}")
+    require(parameters.size <= constTypes.size, "number of parameters must be <= number of constants")
+
+    // Validate that no parameter is duplicated i.e. points to the same position & also to a valid constant.
+    // Also validate that no two parameters exist with the same name.
+    val paramNames = mutable.Set[String]()
+    val paramIndices = this.parameters.map(p => {
+      require(p.placeholder >= 0 && p.placeholder < constTypes.size,
```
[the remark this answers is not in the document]
darkdrag00nv2: Done

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+    private def serializeString(s: String, w: SigmaByteWriter): Unit = {
+      val bytes = s.getBytes(StandardCharsets.UTF_8)
+      w.putUInt(bytes.length)
+      w.putBytes(bytes)
+    }
+
+    private def parseString(r: SigmaByteReader): String = {
+      val length = r.getUInt().toInt
+      new String(r.getBytes(length), StandardCharsets.UTF_8)
+    }
```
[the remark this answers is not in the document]
darkdrag00nv2: Done

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+  val EmptySeq: IndexedSeq[Parameter] = Array.empty[Parameter]
+
+  /** HOTSPOT: don't beautify this code */
+  object serializer extends SigmaSerializer[Parameter, Parameter] {
+    override def serialize(data: Parameter, w: SigmaByteWriter): Unit = {
+      import sigmastate.Operations.ConstantPlaceholderInfo._
+
+      serializeString(data.name, w)
+      serializeString(data.description, w)
+      w.putUInt(data.placeholder, indexArg)
```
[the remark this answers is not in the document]
darkdrag00nv2: Done

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+        require(paramValue.tpe == constTypes(i),
+          s"parameter type mismatch, expected ${constTypes(i)}, got ${paramValue.tpe}")
+        constants(i) = StdSigmaBuilder.mkConstant(paramValue.value, constTypes(i))
+      } else {
+        constants(i) = StdSigmaBuilder.mkConstant(constValues.get(i).get, constTypes(i))
+      }
+    }
+
+    ErgoTree(
+      ErgoTree.ConstantSegregationHeader,
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, updated.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+      }
+    }
+  }
+
+  def applyTemplate(paramValues: Map[String, Values.Constant[SType]]): Values.ErgoTree = {
+    val nConsts = constTypes.size
+    val requiredParameterNames =
+      this.parameters
+        .filter(p => constValues.isEmpty || constValues.get(p.placeholder).isEmpty)
+        .map(p => p.name)
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
darkdrag00nv2: Done.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
darkdrag00nv2: Done, updated to use TypeSerializer.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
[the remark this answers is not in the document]
aslesarenko: Unfortunately it is not that simple. See sizeBit of the header and how it depends on version. Also search for headerWithVersion method.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+import sigmastate.lang.DeserializationSigmaBuilder
+import sigmastate.basics.CryptoConstants
+import org.ergoplatform.ErgoBox
+import scorex.util.ModifierId
+import sigmastate.utils.Helpers
+import scorex.crypto.authds.ADDigest
+import scorex.crypto.hash.Blake2b256
+
+
+case class Parameter(
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
darkdrag00nv2: Note that not all cases were handled in the original class. So I added additional code to handle them.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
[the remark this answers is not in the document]
darkdrag00nv2: Thanks. I've updated the test and then the implementation to take the sizeBit into account.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,476 @@
...
+        if (nConstants > 0) {
+          val res = safeNewArray[Option[SType#WrappedType]](nConstants)
+          cfor(0)(_ < nConstants, _ + 1) { i =>
+            res(i) = r.getOption((() => DataSerializer.deserialize(constTypes(i), r))())
+          }
+          res
+        } else {
+          Array.empty[Option[SType#WrappedType]]
+        }
+      })())
```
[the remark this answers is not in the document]
darkdrag00nv2: Can you help here please? New to Scala and not sure how to reduce the parens. They all seem to be required.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
darkdrag00nv2: For testing this code path, I was wondering if it would be nice to randomly generate `ErgoTree` in tests. Mostly to make sure that the `Zero` typeclass is properly tested. Do we use a fuzzer somewhere else in the code? If yes, please share.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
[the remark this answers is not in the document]
aslesarenko: committed the fix

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
aslesarenko: No, we don't have fuzzer. ErgoTree is too complex for fuzzer probably.

sdk/shared/src/test/scala/org/ergoplatform/sdk/ContractTemplateSpecification.scala
```diff
@@ -0,0 +1,344 @@
...
+        Constant.EmptySeq,
+        expressionTrees(2)
+      )
+    )
+
+    templates.indices.foreach(i =>
+      templates(i).applyTemplate(Some(ergoTreeVersionInTests), templateValues(i)) shouldEqual expectedErgoTree(i)
+    )
+  }
+
```
aslesarenko, line 242: One problem with the tests below. They are not systematic. There two ways to solve it. 1) create template generator 2) manually create many templates from simples to more complex increasing complexity in one parameter on each step.
If do it manually, it would be easier to reuse common sub-components. It is safe becase the data structure is ummutable so parts can be shared by many instances.

aslesarenko (changes requested):
Looks good overall, but many small changes required here and there.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,508 @@
...
+        .filter(p => constValues.get(p.constantIndex).isEmpty)
+        .map(p => p.name)
+    requiredParameterNames.foreach(name => require(
+      paramValues.contains(name),
+      s"value for parameter $name was not provided while it does not have a default value."))
+
+    val paramIndices = this.parameters.map(p => p.constantIndex).toSet
+    val constants = safeNewArray[Constant[SType]](nConsts)
+    cfor(0)(_ < nConsts, _ + 1) { i =>
+      if (paramIndices.contains(i) && paramValues.contains(parameters(i).name)) {
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, good catch. Fixed and added a new test.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,508 @@
...
+
+private trait Zero[T] {
+  def zero: T
+}
+
+private case class CZero[T](zero: T) extends Zero[T]
+
+private trait ZeroLowPriority {
+  implicit def collIsZero[T: Zero: RType]: Zero[Coll[T]] = CZero(Colls.emptyColl[T])
+  implicit def optionIsZero[T: Zero]: Zero[Option[T]] = CZero(Some(Zero.zeroOf[T]))
```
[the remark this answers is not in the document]
darkdrag00nv2: Done, updated.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,508 @@
...
+  implicit def pairIsZero[A: Zero, B: Zero]: Zero[(A,B)] = CZero(Zero[A].zero, Zero[B].zero)
+  implicit def funcIsZero[A: Zero, B: Zero]: Zero[A =>B] = CZero((_ : A) => { Zero[B].zero })
```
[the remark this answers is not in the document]
darkdrag00nv2: I think so. I am using it in handling `case ft: FuncType[a, b] => funcIsZero(typeToZero(ft.tDom), typeToZero(ft.tRange))` in `typeToZero`.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,508 @@
...
+  implicit val BigIntIsZero: Zero[BigInt] = CZero(CBigInt(BigInteger.ZERO))
+  implicit val GroupElementIsZero: Zero[GroupElement] = CZero(CGroupElement(CryptoConstants.dlogGroup.identity))
+  implicit val AvlTreeIsZero: Zero[AvlTree] = CZero({
+    val avlProver = new BatchAVLProver[Digest32, Blake2b256.type](keyLength = 32, None)
+    val digest = avlProver.digest
+    val treeData = new AvlTreeData(digest, AvlTreeFlags.AllOperationsAllowed, 32, None)
+    CAvlTree(treeData)
+  })
+  implicit val sigmaPropIsZero: Zero[SigmaProp] = CZero(CSigmaProp(TrivialProp.FalseProp))
+  implicit val AnyIsZero: Zero[Any] = CZero(0)
```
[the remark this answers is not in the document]
darkdrag00nv2: Are you suggesting we don't need to handle `case AnyType => Zero[Any]`? In other words, there won't be any `SAny` in the ErgoTree?

sdk/shared/src/test/scala/org/ergoplatform/sdk/ContractTemplateSpecification.scala
```diff
@@ -0,0 +1,344 @@
...
+    res shouldBe obj
+  }
+  
+  private def serializationRoundTrip(template: ContractTemplate): Assertion = {
+    val w = SigmaSerializer.startWriter()
+    ContractTemplate.serializer.serialize(template, w)
+    val bytes = w.toBytes
+    val r = SigmaSerializer.startReader(bytes)
+    val res2 = ContractTemplate.serializer.parse(r)
+    res2 shouldEqual template
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

sdk/shared/src/test/scala/org/ergoplatform/sdk/ContractTemplateSpecification.scala
```diff
@@ -0,0 +1,344 @@
...
+          createParameter("p1", 0),
+          createParameter("p2", 1),
+          createParameter("p3", 2)),
+        EQ(Plus(ConstantPlaceholder(0, SType.typeByte),
+          ConstantPlaceholder(1, SType.typeByte)),
+          ConstantPlaceholder(2, SType.typeByte)).toSigmaProp
+      ),
+      t =>
+        t.isInstanceOf[IllegalArgumentException] &&
+        t.getMessage.contains("constValues must be empty or of same length as constTypes. Got 2, expected 3"))
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,508 @@
...
+            ValueSerializer.deserialize(r).toSigmaProp)
+        case _ => Left(DecodingFailure("Failed to decode contract template", cursor.history))
+      }
+    })
+  }
+}
+
+private trait Zero[T] {
+  def zero: T
+}
```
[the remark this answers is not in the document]
darkdrag00nv2: Done.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,508 @@
...
+    outputs = Colls.emptyColl,
+    height = IntIsZero.zero,
+    selfBox = CostingBox(syntheticBox),
+    selfIndex = IntIsZero.zero,
+    lastBlockUtxoRootHash = AvlTreeIsZero.zero,
+    _minerPubKey = Colls.emptyColl[Byte],
+    vars = Colls.emptyColl,
+    activatedScriptVersion = ByteIsZero.zero,
+    currentErgoTreeVersion = ByteIsZero.zero
+  )
```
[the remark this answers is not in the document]
darkdrag00nv2: No, I am defining these `synthetic` values for reuse while defining Zero for complicated constructs such as `Box`, `Context`.

For example:
`syntheticBox` is used in defining `Zero[Box]` as well as `Zero[Context]`.

sdk/shared/src/test/scala/org/ergoplatform/sdk/ContractTemplateSpecification.scala
darkdrag00nv2: Done, went ahead with the contract template generator approach.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
darkdrag00nv2: ok, I guess the template generator is sufficient.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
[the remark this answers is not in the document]
aslesarenko: Ok, in this case Zero for A is not required.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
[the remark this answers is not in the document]
aslesarenko: What is the difference between syntheticBox and Zero[Box].zero? You can just use the instance of Zero[Box] everywhere you need it.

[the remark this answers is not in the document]
darkdrag00nv2: I see what you're saying. Updated the code now.

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
```diff
@@ -0,0 +1,508 @@
...
+          }).asJson),
+        "parameters" -> ct.parameters.asJson,
+        "expressionTree" -> expressionTreeWriter.toBytes.asJson
+      )
+    })
+
+    /** Json decoder for ContractTemplate */
+    implicit val decoder: Decoder[ContractTemplate] = Decoder.instance({ implicit cursor =>
+      val constTypesResult = cursor.downField("constTypes").as[IndexedSeq[SType]]
+      val expressionTreeBytesResult = cursor.downField("expressionTree").as[Array[Byte]]
```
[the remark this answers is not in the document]
darkdrag00nv2: I don't think so. The test for round trip for json would have failed otherwise.

sdk/shared/src/main/scala/org/ergoplatform/sdk/utils/Zero.scala
```diff
@@ -0,0 +1,144 @@
...
+      extensionRoot = Colls.emptyColl[Byte],
+      minerPk = GroupElementIsZero.zero,
+      powOnetimePk = GroupElementIsZero.zero,
+      powNonce = Colls.emptyColl[Byte],
+      powDistance = BigIntIsZero.zero,
+      votes = Colls.emptyColl[Byte]
+    )
+  })
+
+  def typeToZero[T](t: RType[T]): Zero[T] = (t match {
```
darkdrag00nv2, line 120: After looking more closely, I am wondering if I should update `typeToZero` method to take `SType` instead of `RType`.

This way I'll be able to avoid converting from `SType` to `RType` before calling this function.

` DeserializationSigmaBuilder.mkConstant(Zero.zeroOf(Zero.typeToZero(Evaluation.stypeToRType(t))), t)
`

sdk/shared/src/main/scala/org/ergoplatform/sdk/ContractTemplate.scala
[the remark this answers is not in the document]
darkdrag00nv2: Sure, updated it.

sdk/shared/src/main/scala/org/ergoplatform/sdk/utils/Zero.scala
aslesarenko: Yes, make sense, I wonder what can be the signature?

aslesarenko (commented):
Looks like almost done. To fix the tests please cherry-pick this commit https://github.com/ScorexFoundation/sigmastate-interpreter/commit/9791ceaebdb54b55047239b03428444f1101d848
  (which is related to https://github.com/scala/bug/issues/2034)

darkdrag00nv2:
Thanks for sharing the fix. Do you suggest cherry-picking in this PR or #871 which I specifically opened for this?

aslesarenko:
> Do you suggest cherry-picking in this PR or #871 which I specifically opened for this?

I thought about cherry-picking here, hopefully it will be enough to fix the CI tests.

darkdrag00nv2:
@aslesarenko Cherry-picked here. Had to resolve a small merge conflict.

darkdrag00nv2:
@aslesarenko This should be ready to merge now.

aslesarenko (approved):

kushti (approved):
Approved, but better to document Zero.scala (trait Zero , CZero, ZeroLowPriority etc)

aslesarenko:
@darkdrag00nv2, please address the comment from kushti.

darkdrag00nv2:
@aslesarenko @kushti Added Scaladoc for Zero.scala. Thanks!

PR #863, opened 2023-02-14
[v5.x] Abstract from Bouncycastle (restructure, part 2)

aslesarenko:
In this PR bouncy castle dependency is abstracted behind CryptoFacade.
This is a necessary step towards Scala.js support where the same CryptoFacade is implemented using existing JS libraries.


sigma-api/src/main/scala/special/sigma/SigmaDsl.scala
```diff
@@ -200,6 +199,7 @@ trait BigInt {
 @scalan.Liftable
 @WithMethodCallRecognizers
 trait GroupElement {
+  /** Checks if this is the point at infinity. */
```
kushti, line 202: Actually, for a group there is a notion of identity element, not a "point at infinity" (the latter is from elliptic curves terminology). Thus it would be better not to continue to mix terms, rename `isInfinity` to `isIdentity` and `negate` to `inverse`

sigmastate/src/main/scala/sigmastate/basics/BcDlogGroup.scala
```diff
@@ -1,28 +1,26 @@
...
+import sigmastate.crypto.{CryptoContext, CryptoFacade}
 
-import scala.collection.{Seq, mutable}
-import scala.util.Try
+import scala.collection.mutable
 
-abstract class BcDlogGroup[ElemType <: ECPoint](val x9params: X9ECParameters) extends DlogGroup[ElemType] {
 
-  lazy val curve = x9params.getCurve
+/** Base class for all dlog groups (with bouncycastle-like interface).
```
kushti, line 11: Not all dlog groups, but EC-based groups where DLOG problem is hard

sigmastate/src/main/scala/sigmastate/basics/BcDlogGroup.scala
```diff
@@ -1,28 +1,26 @@
...
+  * @param ctx context which abstracts basic operations with curve and elements.
+  */
+abstract class BcDlogGroup(val ctx: CryptoContext) extends DlogGroup {
+  /** Modulus of the finite field of the underlying curve. */
```
kushti, line 15: Characteristic of a finite field

sigmastate/src/main/scala/sigmastate/basics/BcDlogGroup.scala
```diff
@@ -1,28 +1,26 @@
...
+  lazy val p: BigInteger = ctx.getModulus
 
-  //modulus of the field
-  lazy val p: BigInteger = curve.getField.getCharacteristic
+  /** Order of the group (parameter N of the elliptic curve). */
```
kushti, line 18: Notation is different from standard to standard, please provide a link to a standard "N" is taken from

sigmastate/src/main/scala/sigmastate/crypto/CryptoFacade.scala
```diff
@@ -0,0 +1,86 @@
...
+  def multiplyPoints(p1: Ecp, p2: Ecp): Ecp = Platform.multiplyPoints(p1, p2)
+
+  /** Return simplified string representation of the point (used only for debugging) */
+  def showPoint(p: Ecp): String = Platform.showPoint(p)
+
+  /** Returns the value of bit 0 in BigInteger representation of this point. */
+  def testBitZeroOfFieldElem(p: ECFieldElem): Boolean = Platform.testBitZeroOfFieldElem(p)
+
+  /** Returns byte representation of the given field element. */
+  def getEncodedOfFieldElem(p: ECFieldElem): Array[Byte] = Platform.getEncodedOfFieldElem(p)
```
kushti, line 49: encodeFieldElem ? getEncodedOfFieldElem sounds like result should be of Encoded[] type

sigmastate/src/main/scala/sigmastate/crypto/CryptoFacade.scala
```diff
@@ -0,0 +1,86 @@
...
+    * @param p2 second point
+    * @return group multiplication (p1 * p2)
+    */
+  def multiplyPoints(p1: Ecp, p2: Ecp): Ecp = Platform.multiplyPoints(p1, p2)
+
+  /** Return simplified string representation of the point (used only for debugging) */
+  def showPoint(p: Ecp): String = Platform.showPoint(p)
+
+  /** Returns the value of bit 0 in BigInteger representation of this point. */
+  def testBitZeroOfFieldElem(p: ECFieldElem): Boolean = Platform.testBitZeroOfFieldElem(p)
```
kushti, line 46: Why implementation details are exposed here? I guess the method should return sign encoded as a boolean, then name accordingly and describe encoding in ScalaDoc

kushti (changes requested):

sigmastate/src/main/scala/sigmastate/basics/BcDlogGroup.scala
aslesarenko: @kushti could you please suggest an exact edit.

sigmastate/src/main/scala/sigmastate/basics/BcDlogGroup.scala
aslesarenko: fixed

sigma-api/src/main/scala/special/sigma/SigmaDsl.scala
aslesarenko: Created a new [issue](https://github.com/ScorexFoundation/sigmastate-interpreter/issues/865) so this can be done on top of the latest `develop` branch.
Note, the comment about `negate`.

sigmastate/src/main/scala/sigmastate/basics/BcDlogGroup.scala
aslesarenko: fixed

sigmastate/src/main/scala/sigmastate/crypto/CryptoFacade.scala
aslesarenko: The logic behind this name is that it uses `getEncoded` method of ECFieldElement. So, if you are not insisting of the rename let's leave it like this.

sigmastate/src/main/scala/sigmastate/crypto/CryptoFacade.scala
aslesarenko: Fixed the description.

Here is the implementation in bouncycastle:
```
    public boolean testBitZero()
    {
        return toBigInteger().testBit(0);
    }
```
I'm not sure the encoding should be described on the CryptoFacade level.

For renaming, I can either add TODO or create an issue to be done later on top of `develop`.

sigmastate/src/main/scala/sigmastate/crypto/BigIntegers.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  private val MAX_ITERATIONS = 1000
+
+  /** Create the given number of random bits.
+    * @param bitLength the number of random bits to create.
+    * @param random a source of randomness.
+    * @return a byte array containing random bits.
+    */
+  @throws[IllegalArgumentException]
+  def createRandom(bitLength: Int, random: SecureRandom) = {
```
kushti, line 22: result type missed

sigma-api/src/main/scala/special/sigma/SigmaDsl.scala
kushti: Why not to remove isInfinity in this PR? The problem is that a group does not have any "point at infinity", and now on top of name which is not proper you're doing comment which does not make any sense

sigmastate/src/main/scala/sigmastate/basics/BcDlogGroup.scala
kushti: @aslesarenko getModulus => getCharacteristic or getFieldCharacteristic

sigmastate/src/main/scala/sigmastate/crypto/CryptoFacade.scala
kushti: Why are you exposing in the name what is called inside ? The name should say what the method is doing, not how

sigmastate/src/main/scala/sigmastate/crypto/CryptoFacade.scala
kushti: What you are actually doing that is checking sign of a coordinate of a EC point (when it is encoded with affine coordinates). Why not to make method named correspondingly, hiding BC implementation details from the facade (which what good design should do) ?

sigmastate/src/main/scala/sigmastate/crypto/CryptoContext.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import java.math.BigInteger
+
+/** A context for cryptographic operations. */
+abstract class CryptoContext {
+  /** The order of the underlying finite field. */
+  def getModulus: BigInteger
+
+  /** The order of the underlying group. */
+  def getOrder: BigInteger
```
kushti, line 11: Why getOrder ? In Scala get* is not popular, and idiomatic way is do just `def order` (especially considering that likely it will be rewritten with a constant in implementation, as group order is a constant indeed). 

The same for all other get* methods in this class

sigmastate/src/main/scala/sigmastate/crypto/CryptoContext.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  /** The generator of the underlying group. */
+  def validatePoint(x: BigInteger, y: BigInteger): Ecp
+
+  /** The point at infinity. */
+  def getInfinity(): Ecp
```
kushti, line 17: identity

sigmastate/src/main/scala/sigmastate/crypto/CryptoContext.scala
```diff
@@ -0,0 +1,25 @@
...
+
+/** A context for cryptographic operations. */
+abstract class CryptoContext {
+  /** The order of the underlying finite field. */
+  def getModulus: BigInteger
+
+  /** The order of the underlying group. */
+  def getOrder: BigInteger
+
+  /** The generator of the underlying group. */
```
kushti, line 13: ScalaDoc is wrong here

sigmastate/src/main/scala/sigmastate/crypto/Platform.scala
```diff
@@ -0,0 +1,106 @@
...
+     * BC treats EC as additive group while we treat that as multiplicative group.
+     * Therefore, exponentiate point is multiply.
+     */
+    Ecp(p.value.multiply(n))
+  }
+
+  /** Check if a point is infinity. */
+  def isInfinityPoint(p: Ecp): Boolean = p.value.isInfinity
+
+  /** Negate a point. */
```
kushti, line 94: The comment above says "BC treats EC as additive group while we treat that as multiplicative group.", then negate should be removed to inverse (as in an additive group only inversion could be called a  negation)

sigmastate/src/main/scala/sigmastate/interpreter/CryptoConstants.scala
```diff
@@ -2,17 +2,21 @@ package sigmastate.interpreter
...
-import org.bouncycastle.math.ec.custom.sec.SecP256K1Point
-import sigmastate.basics.{BcDlogGroup, SecP256K1}
-
+/** Constants used in crypto operations implementation. */
 object CryptoConstants {
-  type EcPointType = SecP256K1Point
+  /** Type of group elements used in the signature scheme. */
+  type EcPointType = Ecp
 
+  /** Length of encoded group elements in bytes. */
```
kushti, line 13: encoded group elements => encoded group element

sigmastate/src/main/scala/sigmastate/serialization/GroupElementSerializer.scala
```diff
@@ -21,12 +22,12 @@ object GroupElementSerializer extends SigmaSerializer[EcPointType, EcPointType]
   private lazy val identityPointEncoding = Array.fill(encodingSize)(0: Byte)
 
   override def serialize(point: EcPointType, w: SigmaByteWriter): Unit = {
-    val bytes = if (point.isInfinity) {
+    val bytes = if (CryptoFacade.isInfinityPoint(point)) {
```
kushti, line 25: Better import CryptoFacade functions in the beginning of the method to have short names after, readability of the new version is poor

sigma-api/src/main/scala/special/sigma/SigmaDsl.scala
aslesarenko: Why it is named isInfinity in BC then? 
Anyway, renamed the method and removed ScalaDoc. 
After you critique, I'm reluctant to write anything there.

sigmastate/src/main/scala/sigmastate/basics/BcDlogGroup.scala
aslesarenko: renamed

sigmastate/src/main/scala/sigmastate/crypto/CryptoFacade.scala
aslesarenko: renamed

sigmastate/src/main/scala/sigmastate/crypto/CryptoFacade.scala
aslesarenko: @kushti Please suggest the name of your liking, I will use it.

sigmastate/src/main/scala/sigmastate/crypto/BigIntegers.scala
aslesarenko: fixed

sigmastate/src/main/scala/sigmastate/crypto/CryptoContext.scala
aslesarenko: fixed

sigmastate/src/main/scala/sigmastate/crypto/CryptoContext.scala
aslesarenko: renamed

sigmastate/src/main/scala/sigmastate/crypto/Platform.scala
aslesarenko: I got it, but we cannot fully rename `negate` as you wish, because it is used in ErgoScript. So it is better to leave consistently `negate` everywhere instead of having both `negate` and `inverse`, which will be even more confusing.

sigmastate/src/main/scala/sigmastate/interpreter/CryptoConstants.scala
aslesarenko: fixed

sigmastate/src/main/scala/sigmastate/serialization/GroupElementSerializer.scala
aslesarenko: It is matter of taste. This way is more explicit. Anyway, this is not significant and I would better leave it as is. (I already expect enough troubles with merging)

sigma-api/src/main/scala/special/sigma/SigmaDsl.scala
kushti: In an elliptic curve implementation you mean ? 

Well, an elliptic curve has a notion of infinity point, as well as addition, negation etc. A group does not have a notion of infinity point. A group has an identity element (which is instantiated with an infinity point when you instantiate a group on top of EF). Check group theory materials, e.g. https://crypto.stanford.edu/pbc/notes/group/group.html . You can make a ScalaDoc based on that.

sigmastate/src/main/scala/sigmastate/basics/BcDlogGroup.scala
kushti: ScalaDoc not updated

sigmastate/src/main/scala/sigmastate/crypto/CryptoFacade.scala
kushti: `signOf(p)`

kushti:
Rework https://github.com/ScorexFoundation/sigmastate-interpreter/issues/865 according to recent changes. Make an issue about fixing "negate".

This is pretty important issue, as currently you're exposing a group description which does not sound. Again , a group defined via multiplicative notation (https://crypto.stanford.edu/pbc/notes/group/group.html) does not have "negate" operation.

sigma-api/src/main/scala/special/sigma/SigmaDsl.scala
aslesarenko: Please write the ScalaDoc is you think make sense. I will go ahead with other tasks.

sigmastate/src/main/scala/sigmastate/basics/BcDlogGroup.scala
aslesarenko: fixed

sigmastate/src/main/scala/sigmastate/crypto/CryptoFacade.scala
aslesarenko: renamed

aslesarenko:
> This is pretty important issue, as currently you're exposing a group description which does not sound. Again , a group defined via multiplicative notation (https://crypto.stanford.edu/pbc/notes/group/group.html) does not have "negate" operation.

Updated the issue. I don't understand the reason you call this group multiplicative, while its implementation is based on BC which is additive. What is the point? Why multiplicative?

sigma-api/src/main/scala/special/sigma/SigmaDsl.scala
kushti: ` /** Checks if the provided element is an identity element. */` simply

kushti:
> > This is pretty important issue, as currently you're exposing a group description which does not sound. Again , a group defined via multiplicative notation (https://crypto.stanford.edu/pbc/notes/group/group.html) does not have "negate" operation.
> 
> Updated the issue. I don't understand the reason you call this group multiplicative, while its implementation is based on BC which is additive. What is the point? Why multiplicative?

Papers and textbooks in cryptography are using multiplicative notation for groups, and the main hard problems is called discrete log problem then (not multiplicative inverse problem). Thus multiplicative notation exposed allows for ErgoTree / ErgoScript code to be close to paper description of protocols. See the ZeroJoin paper https://eprint.iacr.org/2020/560.pdf as an example. Also, Sigma protocols code was inspired by SCAPI framework architecture, which is doing similar wrapping , see their documentation on DLOG group interface https://scapi.readthedocs.io/en/latest/primitives/dlog.html

kushti (approved):

sigma-api/src/main/scala/special/sigma/SigmaDsl.scala
aslesarenko: fixed, thanks

PR #864, opened 2023-02-15
Extract SigmaParser and deps into a new module

darkdrag00nv2:
Ground work for #851 

Extracted `SigmaParser` and its dependencies to a new module named `parsers` and added support for cross-compilation on it.

This was needed because in order to decode a `ContractTemplate` for EIP-5, we need to decode the `SType` from the JSON. Support for doing that is available in `SigmaParser` which needs to be available from the `sdk` module. This helps avoid the `sdk` module depending on the `sc` module.


darkdrag00nv2:
@aslesarenko This mostly works except one problem. The `SigmaParserTest` which I've moved to the `parsers` module has some test case that use the `SigmaCompiler`. [See here](https://github.com/ScorexFoundation/sigmastate-interpreter/blob/f6db2bbf3f4ddcc07d18d23bc01f557c23290264/sc/src/test/scala/sigmastate/lang/SigmaParserTest.scala#L40). The `parsers` module can't depend on the `sc` module as the later already depends on the former.

I can see two solutions:
1. Keep `SigmaParserTest` inside the `sc` module itself. Probably not ideal
2. Move the test cases that do not need the `SigmaCompiler` to the `parsers` module and move the remaining to `SigmaCompilerTest`

Thoughts?

aslesarenko:
>  Move the test cases that do not need the `SigmaCompiler` to the `parsers` module and move the remaining to `SigmaCompilerTest`

@darkdrag00nv2 This sounds good.

parsers/shared/src/test/scala/sigmastate/lang/SigmaParserTest.scala
```diff
@@ -561,24 +551,6 @@ class SigmaParserTest extends AnyPropSpec with ScalaCheckPropertyChecks with Mat
...
 
-  property("negative tests") {
-    fail("(10", 1, 4)
-    fail("10)", 1, 3)
-    fail("X)", 1, 2)
-    fail("(X", 1, 3)
-    fail("{ X", 1, 4)
-    fail("{ val X", 1, 8)
-    fail("\"str", 1, 5)
-  }
```
aslesarenko, line 572: Looks like moving tests out of here is not motivated enough. You can re-implement the `fail` method without relying on SigmaCompiler, but using SigmaParser directly. If you look at the `parse` method, you will see it just delegates to SigmaParser.

aslesarenko (changes requested):

parsers/shared/src/test/scala/sigmastate/lang/SigmaParserTest.scala
darkdrag00nv2: Yeah, right. Thanks for catching, I've updated the PR.

darkdrag00nv2:
`+test` still works.

aslesarenko (approved):

PR #867, opened 2023-02-23
Move DataJsonEncoder and JsonCodecs to the sdk module

darkdrag00nv2:
Ground work part 2 for #851 

Moved `DataJsonEncoder` and `JsonCodecs` to the `sdk` module. This is done because we need these to be available from the `sdk` module for EIP-5 implementation and we don't want the `sdk` module to depend on the `sc` module.


darkdrag00nv2:
`+test` ran successfully.

aslesarenko (approved):

PR #872, opened 2023-04-06
autodoc

glasgowm148:
Added autodoc generated documentation for each file in the repo.


aslesarenko:
If the files can be regenerated, then the instruction and the generation script (if any) should be enough. 
Do we really need to add all this files to the repo? 
How to ensure they are in sync with the original files?
Ideally we just have sbt command that do generation on demand.

PR #874, opened 2023-04-17
Add missing CryptoFacade methods

aslesarenko:
In this PR:
- [x] add CryptoContext.curve
- [x] add CryptoFacade.encodePoint
- [x] add CryptoFacade.getCurve

### DIFF
--- a/interpreter/src/main/scala/sigmastate/crypto/CryptoContext.scala
+++ b/interpreter/src/main/scala/sigmastate/crypto/CryptoContext.scala
@@ -4,6 +4,9 @@ import java.math.BigInteger
 
 /** A context for cryptographic operations. */
 abstract class CryptoContext {
+  /** The underlying elliptic curve. */
+  def curve: Curve
+
   /** The characteristics of the underlying finite field. */
   def fieldCharacteristic: BigInteger
 
--- a/interpreter/src/main/scala/sigmastate/crypto/CryptoFacade.scala
+++ b/interpreter/src/main/scala/sigmastate/crypto/CryptoFacade.scala
@@ -48,6 +48,19 @@ object CryptoFacade {
   /** Returns byte representation of the given field element. */
   def encodeFieldElem(p: ECFieldElem): Array[Byte] = Platform.encodeFieldElem(p)
 
+  /** Returns byte representation of the given point.
+    *
+    * @param p          point to encode
+    * @param compressed if true, generates a compressed point encoding
+    */
+  def encodePoint(p: Ecp, compressed: Boolean): Array[Byte] = Platform.encodePoint(p, compressed)
+
+  /** Returns a [[Curve]] instance describing the elliptic curve of the point p
+    *
+    * @param p the elliptic curve point
+    */
+  def getCurve(p: Ecp): Curve = Platform.getCurve(p)
+
   /** Returns the x-coordinate.
     *
     * Caution: depending on the curve's coordinate system, this may not be the same value as in an
--- a/interpreter/src/main/scala/sigmastate/crypto/Platform.scala
+++ b/interpreter/src/main/scala/sigmastate/crypto/Platform.scala
@@ -1,12 +1,17 @@
 package sigmastate.crypto
 
 import org.bouncycastle.crypto.ec.CustomNamedCurves
-import org.bouncycastle.math.ec.{ECFieldElement, ECPoint}
+import org.bouncycastle.math.ec.{ECPoint, ECFieldElement, ECCurve}
 
 import java.math.BigInteger
 
 /** JVM specific implementation of crypto methods*/
 object Platform {
+  /** Returns a [[Curve]] instance describing the elliptic curve of the point p
+    * @param p the elliptic curve point
+    */
+  def getCurve(p: Ecp): Curve = Curve(p.value.getCurve)
+
   /** Returns the x-coordinate.
     *
     * Caution: depending on the curve's coordinate system, this may not be the same value as in an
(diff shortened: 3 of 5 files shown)


interpreter/src/main/scala/sigmastate/crypto/Platform.scala
```diff
@@ -1,12 +1,17 @@
...
 
 import org.bouncycastle.crypto.ec.CustomNamedCurves
-import org.bouncycastle.math.ec.{ECFieldElement, ECPoint}
+import org.bouncycastle.math.ec.{ECPoint, ECFieldElement, ECCurve}
 
 import java.math.BigInteger
 
 /** JVM specific implementation of crypto methods*/
 object Platform {
+  /** Returns a [[Curve]] instance describing the elliptic curve of the point p
```
kushti, line 10: No need to replicate in the comments what is already in the code, namely, "Returns a [[Curve]] instance".

"Description of elliptic curve of point p which belongs to the curve" is enough

interpreter/src/main/scala/sigmastate/crypto/Platform.scala
```diff
@@ -46,6 +51,12 @@ object Platform {
   /** Returns byte representation of the given field element. */
   def encodeFieldElem(p: ECFieldElem): Array[Byte] = p.value.getEncoded
 
+  /** Returns byte representation of the given point.
```
kushti, line 54: "Returns" is already in the code.  "byte representation of the given point."  is enough

interpreter/src/main/scala/sigmastate/crypto/Platform.scala
```diff
@@ -94,6 +105,9 @@ object Platform {
   /** Negate a point. */
   def negatePoint(p: Ecp): Ecp = Ecp(p.value.negate())
 
+  /** Wrapper for curve type. */
```
kushti, line 108: But why the wrapper is needed at all ?

interpreter/src/main/scala/sigmastate/crypto/CryptoFacade.scala
```diff
@@ -48,6 +48,19 @@ object CryptoFacade {
   /** Returns byte representation of the given field element. */
   def encodeFieldElem(p: ECFieldElem): Array[Byte] = Platform.encodeFieldElem(p)
 
+  /** Returns byte representation of the given point.
```
kushti, line 51: See comment below

kushti (changes requested):

interpreter/src/main/scala/sigmastate/crypto/CryptoContext.scala
```diff
@@ -4,6 +4,9 @@ import java.math.BigInteger
 
 /** A context for cryptographic operations. */
 abstract class CryptoContext {
+  /** The underlying elliptic curve. */
```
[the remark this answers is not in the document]
aslesarenko: There is the underlying curve, look other methods and their implementations. Fixed description of the class itself and the method. Still "underlying" word make sense.

interpreter/src/main/scala/sigmastate/crypto/package.scala
```diff
@@ -1,8 +1,12 @@
 package sigmastate
 
 package object crypto {
+  /** Instance of Elliptic Curve. */
```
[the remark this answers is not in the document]
aslesarenko: To make it consistent with the other types here, added the word "descriptor".

interpreter/src/main/scala/sigmastate/crypto/Platform.scala
aslesarenko: Fixed, but this is a matter of taste. "Returns ..." is ok accoding to the style guide.

interpreter/src/main/scala/sigmastate/crypto/Platform.scala
aslesarenko: Fixed, no problem, but does it really worth to spend efforts on this? I don't think we should be so strict in our reviews. Do you want me to follow the same strict standard, I assume yes?

interpreter/src/main/scala/sigmastate/crypto/Platform.scala
aslesarenko: To hide underlying nature of the abstraction and avoid leakage of the concrete ECCurve class

interpreter/src/main/scala/sigmastate/crypto/CryptoFacade.scala
aslesarenko: fixed

interpreter/src/main/scala/sigmastate/crypto/CryptoFacade.scala
```diff
@@ -48,6 +48,19 @@ object CryptoFacade {
...
+    *
+    * @param p          point to encode
+    * @param compressed if true, generates a compressed point encoding
+    */
+  def encodePoint(p: Ecp, compressed: Boolean): Array[Byte] = Platform.encodePoint(p, compressed)
+
+  /** Returns a [[Curve]] instance describing the elliptic curve of the point p
```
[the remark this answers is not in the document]
aslesarenko: fixed

interpreter/src/main/scala/sigmastate/crypto/Platform.scala
kushti: It is not about "Returns...", rather, "a [[Curve]] instance"

interpreter/src/main/scala/sigmastate/crypto/Platform.scala
kushti: What standard are you following during review? Any link ?

aslesarenko: This is figuratively speaking "standard". You establish the review rules here, and I try to follow them while reviewing Ergo PRs.

kushti: And vice versa. Why not to standardize them then? It would be good for sure to avoid comments about "black names" etc

PR #878, opened 2023-04-24
Support Scala.js and further reorg (part1)

aslesarenko:
In this PR:
- [x] Scala.js compilation configured for common, core-lib and interpreter modules
- [x] consensus code further minimized
- [x] more code moved to compiler `sc` module
- [x] unused code removed

### DIFF
--- a/common/jvm/src/test/scala/scalan/util/FileUtil.scala
+++ b/common/jvm/src/test/scala/scalan/util/FileUtil.scala
@@ -1,13 +1,14 @@
 package scalan.util
 
+import scalan.util.CollectionUtil.AnyOps
+import scalan.util.StringUtil.cleanFileName
+
 import java.io._
 import java.net.{JarURLConnection, URL}
 import java.nio.file._
 import java.nio.file.attribute.BasicFileAttributes
 import scala.Console
 import scala.collection.JavaConverters._
-import scalan.util.StringUtil.{StringUtilExtensions, cleanFileName}
-import scalan.util.CollectionUtil.AnyOps
 
 object FileUtil {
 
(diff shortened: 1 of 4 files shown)


common/jvm/src/test/scala/scalan/util/FileUtil.scala
```diff
@@ -6,10 +6,36 @@ import java.nio.file._
 import java.nio.file.attribute.BasicFileAttributes
 import scala.Console
 import scala.collection.JavaConverters._
-import scalan.util.StringUtil.StringUtilExtensions
+import scalan.util.StringUtil.{StringUtilExtensions, cleanFileName}
```
kushti, line 9: StringUtilExtensions import not used

import scala.Console    as well

interpreter/shared/src/main/scala/sigmastate/sigmastate.scala
```diff
@@ -0,0 +1,49 @@
+import sigmastate.Values._
+import sigmastate.lang.CheckingSigmaBuilder
+
+package object sigmastate {
+  import CheckingSigmaBuilder._
+
+  /**
+    * SInt addition
```
kushti, line 8: Why SInt? What does this comment mean ?

kushti (approved):

common/jvm/src/test/scala/scalan/util/FileUtil.scala
aslesarenko: fixed (except Console, which is used)

interpreter/shared/src/main/scala/sigmastate/sigmastate.scala
aslesarenko: fixed

PR #883, opened 2023-05-07
Support Scala.js and further reorg (part4)

aslesarenko:
In this PR:
- [x] ScalablyTyped is configured to import FleetSDK
- [x] Implemented conversion between Fleet transaction JS object and ErgoLikeTransaction that can be reduced and signied
- [x] fixed equality of TokenId 
- [x] implemented JS version of CryptoFacade.encodePoint
- [x] ScalaDocs for all touched code
- [x] refactoring and cleanup


sdk/js/src/main/scala/org/ergoplatform/sdk/js/ProverBuilder.scala
```diff
@@ -0,0 +1,52 @@
...
+    )
+    this
+  }
+
+  def withEip3Secret(index: Int): ProverBuilder = {
+    _builder.withEip3Secret(index)
+    this
+  }
+
+  def withDHTData(g: String, h: String, u: String, v: String, x: js.BigInt): ProverBuilder = {
```
kushti, line 32: withDHTSecret is the better name

sdk/js/src/main/scala/org/ergoplatform/sdk/js/Prover.scala
```diff
@@ -0,0 +1,57 @@
...
+import org.ergoplatform.sdk
+import typings.fleetSdkCommon.boxesMod.Box
+import typings.fleetSdkCommon.{commonMod, inputsMod, tokenMod, transactionsMod}
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSExportTopLevel
+
+/** Equivalent of [[sdk.Prover]] available from JS. */
+@JSExportTopLevel("Prover")
+class Prover(_prover: sdk.Prover) extends js.Object {
```
kushti, line 12: too generic name, better to have more descriptive

interpreter/shared/src/main/scala/org/ergoplatform/ErgoBox.scala
```diff
@@ -103,10 +103,13 @@ object ErgoBox {
     val size: Short = 32
   }
 
-  type TokenId = Digest32
+  /** Token id is tagged collection of bytes. */
+  type TokenId = Digest32Coll
```
kushti, line 107: is this change really needed? would break a lot of things upstream

If needed, make alt constructor with old arguments at least

kushti (changes requested):

interpreter/js/src/main/scala/sigmastate/crypto/Platform.scala
```diff
@@ -71,7 +71,10 @@ object Platform {
     * @param p point to encode
     * @param compressed if true, generates a compressed point encoding
     */
-  def encodePoint(p: Ecp, compressed: Boolean): Array[Byte] = ???
+  def encodePoint(p: Ecp, compressed: Boolean): Array[Byte] = {
+    val hex = p.point.toHex(compressed)
```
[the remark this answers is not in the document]
aslesarenko: This method wraps ECPoint.getEncoded, where the logic you mention is not implemented.
This logic which you describe is implemented in GroupElement.getEncoded.

ECPoint.getEncoded is not used in sigma directly, but it is used in ergo-wallet for Extended keys.

However, there is indeed difference between toHex and ECPoint.getEncoded, for infinite point.
Fixed this and added tests.

For a reference, here is now toHex is implemented https://github.com/anon-br/sigmajs-crypto-facade/blob/master/src/noble-secp256k1.ts#L491

And there is ECPoint implementation
```java
    public byte[] getEncoded(boolean compressed)
    {
        if (this.isInfinity())
        {
            return new byte[1];
        }

        ECPoint normed = normalize();

        byte[] X = normed.getXCoord().getEncoded();

        if (compressed)
        {
            byte[] PO = new byte[X.length + 1];
            PO[0] = (byte)(normed.getCompressionYTilde() ? 0x03 : 0x02);
            System.arraycopy(X, 0, PO, 1, X.length);
            return PO;
        }

        byte[] Y = normed.getYCoord().getEncoded();

        byte[] PO = new byte[X.length + Y.length + 1];
        PO[0] = 0x04;
        System.arraycopy(X, 0, PO, 1, X.length);
        System.arraycopy(Y, 0, PO, X.length + 1, Y.length);
        return PO;
    }
```

sdk/js/src/main/scala/org/ergoplatform/sdk/js/Prover.scala
aslesarenko: It is generic prover, nothing special about it, so the name is generic. Please suggest.

interpreter/shared/src/main/scala/org/ergoplatform/ErgoBox.scala
kushti: Alt constructor of ErgoBox. You can modify Ergo node and Appkit, but what about Kiosk and other tools and apps?

interpreter/js/src/main/scala/sigmastate/crypto/Platform.scala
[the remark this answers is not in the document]
kushti: Then the method with its description looks confusing, as from the name and the description it seemingly corresponds to GroupElementSerializer.serialize . Please update description and method name, however, like getASN1Encoding

interpreter/shared/src/main/scala/org/ergoplatform/ErgoBox.scala
aslesarenko: Kiosk is using Appkit and the other apps probably too. 
Sigma has always been discouraged to be used directly (see second paragraph in readme https://github.com/ScorexFoundation/sigmastate-interpreter#ergoscript-compiler-and-ergotree-interpreter).

This is not the first breaking change and also not the most significant and has not been any complains so far.
The most significant was switch to v5.0 which was smooth because Sigma changes was properly hidden behind Appkit's abstractions.
So applications which use direct dependencies on Sigma will have to be updated.

Once refactoring is done (hopefully soon), we can do stabilization of sigma APIs in sigma `sdk` module.
My plan is to make `sdk` module a stable abstraction layer for Scala applications so that using Appkit will not be necessary from Scala projects.
Appkit itself will become a thin wrapper over sigma sdk, so most of the logic will migrate from Appkit to Sigma SDK. (AppkitProvingInterpreter is one example of it, and Extended keys is another).

sdk/js/src/main/scala/org/ergoplatform/sdk/js/Prover.scala
aslesarenko: done

interpreter/js/src/main/scala/sigmastate/crypto/Platform.scala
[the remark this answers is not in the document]
aslesarenko: done

PR #885, opened 2023-06-13
Cross-compile ErgoScript compiler to JS

aslesarenko:
In this PR:
- [x] `graph-ir` module cross-compiled to JVM and JS
- [ ] `sc` module cross-compiled to JVM and JS


interpreter/shared/src/main/scala/sigmastate/eval/CostingDataContext.scala
```diff
@@ -25,12 +25,13 @@ import scalan.{Nullable, RType}
 import scorex.crypto.hash.{Blake2b256, Digest32, Sha256}
 import sigmastate.Values.ErgoTree.EmptyConstants
 import sigmastate.basics.DLogProtocol.ProveDlog
-import sigmastate.basics.{ProveDHTuple, CryptoConstants}
+import sigmastate.basics.{CryptoConstants, ProveDHTuple}
 import sigmastate.crypto.{CryptoFacade, Ecp}
 import sigmastate.lang.TransformingSigmaBuilder
```
kushti, line 30: import sigmastate.lang.TransformingSigmaBuilder and 
import scalan.Nullable

are not used now

kushti (approved):

interpreter/shared/src/main/scala/sigmastate/eval/CostingDataContext.scala
aslesarenko: fixed

PR #887, opened 2023-07-01
fix exports

mgpai22:
Package was not working for me. Modified exports so should work now.

### DIFF
--- a/sigma-js/package.json
+++ b/sigma-js/package.json
@@ -9,7 +9,7 @@
   ],
   "exports": {
     "./internal-*": null,
-    "./*": "./dist/*.js"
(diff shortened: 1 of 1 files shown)


anon-br (commented):
This exporting strategy is intentional due to the huge size of the JS output file (~11 MB), so there is a plan to have multiple exports to make it as tree-shakeable as possible, for now you can import as `sigmastate-js/main`.

In the future, you can import only the modules you need, such as `sigmastate-js/interpreter`, `sigmastate-js/compiler` and so on.

mgpai22:
Got it, thanks

PR #896, opened 2023-07-28
Refactoring and improvments

### LINKED ISSUE #757: Remove ScriptCostLimit
ScriptCostLimit in SigmaConstants used in tests and profiler only, so not really a hard-coded constant. Replace it with some other value, e.g. val costLimitForTests somewhere

### LINKED ISSUE #862: Outdated changelogs and other specs
Some changelogs seem to be abandoned:

https://github.com/ScorexFoundation/sigmastate-interpreter/blob/develop/docs/releasenotes.md 
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/develop/docs/soft-fork-log.md

Does this spec correspond to reality? 

https://github.com/ScorexFoundation/sigmastate-interpreter/blob/develop/docs/sigma-front-end.md

aslesarenko:
In this PR:
- [x] closes #757 
- [x] closes #862 
- [x] simplified sigma protocol abstractions
- [x] fixed equality of Challenge type and all the case classes where it is used
- [x] avoided duplicated code in computing response to challenge 
- [x] some cleanups and avoided warnings


interpreter/shared/src/main/scala/sigmastate/trees.scala
```diff
@@ -38,10 +35,9 @@ trait SigmaConjecture extends SigmaBoolean {
 }
 
 /**
-  * Basic trait for leafs of crypto-trees, such as ProveDlog and ProveDiffieHellman instances
+  * Basic trait for leafs of crypto-trees, such as [[sigmastate.basics.DLogProtocol.ProveDlog]] and [[sigmastate.basics.ProveDHTuple]] instances
   */
-trait SigmaProofOfKnowledgeLeaf[SP <: SigmaProtocol[SP], S <: SigmaProtocolPrivateInput[SP, _]]
-  extends SigmaBoolean with SigmaProtocolCommonInput[SP]
+trait SigmaLeaf extends SigmaBoolean
```
kushti, line 40: why not to eliminate SigmaLeaf, what it is encoding?

kushti (approved):

PR #900, opened 2023-08-05
Reduction and signing of Fleet transactions

aslesarenko:
In this PR more methods exported to JS:
- [x] implemented SigmaProver.reduce which takes Fleet's  UnsignedTransaction
- [x] implemented SigmaProver.signReduced which can return Fleet's SignedTransaction
- [x] ReducedTransaction exported to JS and can be serialised and deserialized
- [x] added more tests 
- [x] some fixes for JS and cleanups


sdk/shared/src/main/scala/org/ergoplatform/sdk/multisig/Signer.scala
```diff
@@ -0,0 +1,17 @@
+package org.ergoplatform.sdk.multisig
+
+import org.ergoplatform.P2PKAddress
+import org.ergoplatform.sdk.SigmaProver
+
+case class Signer(prover: SigmaProver) {
```
kushti, line 6: The file should be removed from this PR I suppose

sdk/shared/src/main/scala/org/ergoplatform/sdk/BlockchainParameters.scala
```diff
@@ -0,0 +1,52 @@
...
+  * @param tokenAccessCost cost of a token contained in a transaction, in computation unit
+  * @param inputCost cost of a transaction input, in computation unit
+  * @param dataInputCost cost of a transaction data input, in computation unit
+  * @param outputCost cost of a transaction output, in computation unit
+  * @param maxBlockCost computation units limit per block
+  * @param softForkStartingHeight height when voting for a soft-fork had been started
+  * @param softForkVotesCollected votes for soft-fork collected in previous epochs
+  * @param blockVersion Protocol version activated on the network
+  */
+case class BlockchainParameters(
```
kushti, line 18: Why not ErgoParameters ? Parameters are not related to a generic blockchain, rather, the Ergo one (I guess it was named ErgoLikeParameters to support Ergo-like blockchains as well)

interpreter/shared/src/main/scala/org/ergoplatform/validation/SigmaValidationSettingsSerializer.scala
```diff
@@ -27,10 +28,12 @@ object SigmaValidationSettingsSerializer extends SigmaSerializer[SigmaValidation
       val status = RuleStatusSerializer.parse(r)
       ruleId -> status
     }
-    val initVs = ValidationRules.currentSettings
-    val res = parsed
-      .filter(pair => initVs.get(pair._1).isDefined)
-      .foldLeft(initVs) { (vs, rule) => vs.updated(rule._1, rule._2) }
+    val map = parsed.map { case (id, status) =>
+      val (rule, _) = ValidationRules.currentSettings.get(id)
+          .getOrElse(throw SerializerException(s"Rule with id $id is not registered"))
```
kushti, line 33: Looks like consensus-breaking change: before, settings not in the currentSettings were ignored just, now they are failing deserialization

kushti (changes requested):

sdk/shared/src/main/scala/org/ergoplatform/sdk/multisig/Signer.scala
aslesarenko: yes, removed

sdk/shared/src/main/scala/org/ergoplatform/sdk/BlockchainParameters.scala
aslesarenko: The new name clarifies that it is `blockchain` parameters (those stored in blockchain), not some other parameters.
The package `org.ergoplatform.sdk` already suggests it is related to Ergo blockchain, so no need to add Ergo once again.

`ErgoLike` prefix is pretty useless (ErgoLikeContext doesn't tell anything about what kind of context it is, InterpreterContext is much better).

I was thinking to rename all `ErgoLike` names at some point as part of further refactoring and code simplification

interpreter/shared/src/main/scala/org/ergoplatform/validation/SigmaValidationSettingsSerializer.scala
aslesarenko: This serialiser is only use in JsonCodecs to serialise Context data. Not part of the consensus.

sdk/shared/src/main/scala/org/ergoplatform/sdk/BlockchainParameters.scala
kushti: So for other Ergo-like blockchains (e.g. sidechains) you will have different packages?

kushti (approved):

sdk/shared/src/main/scala/org/ergoplatform/sdk/BlockchainParameters.scala
aslesarenko: Yes, probably. Everything in org.ergoplatform is part of reference implementation. 
It will depend on how other projects will be architected, as a layer on top of core ergo packages, or as a repo fork with the necessary changes.

PR #934, opened 2023-11-09
replacing-arrays-hash-code 

### LINKED ISSUE #197: Revise and optimize hashCode for Digest32 and other hashes

0xOptimistic:
closes #197 

searched for all array.hashcode, replaced it with Ints.fromByteArray


mhsnmk:
@kushti @aslesarenko can you review this.?

aslesarenko:
Unfortunately such replacements are not possible in general. Also, there is pending PR #789, which is on hold due to low priority. So this PR is duplicate.

kushti:
@mhsnmk why Ints.fromByteArray everywhere?

PR #937, opened 2023-11-18
[WIP] Contract template compiler

### LINKED ISSUE #852: Implement ContractTemplate compiler
The EIP-5 is created to standardize contract template format, which can be supported by ErgoTree compilers.

For ErgoScript compiler the syntax of contracts can be extended 

```scala
/** Description.
  * @param p1 
  * @param p2
  * @param p3
  * @return 
  */
@contract def MyContract(p1: Int, p2: GroupElement, p3: Coll[Byte]) = {
   // contract logic using p1 - p3
}
```

1) Some constants have names in the contract declaration, those names can be used to access the corresponding constant 
2) The mapping between constants and names is known during ErgoTree compilation and can be saved in the resulting ContractTemplate structure.
3) This name-index mapping can be made available and used during transaction construction by SDK libraries supporting EIP-5

ross-weir:
closes #852 

- [x] lots of polishing
- [x] harden some parsers to account for whitespace & new lines
- [x] replace with `SigmaParser` parsers where possible
- [x] tests/docs


ross-weir:
@ergomorphic i have a few todos left but this could have an initial quick looking over - contract template compilation is working

parsers/shared/src/main/scala/sigmastate/lang/ContractTemplateParser.scala
```diff
@@ -0,0 +1,12 @@
...
+import fastparse._
+import fastparse.NoWhitespace._
+
+case class ParsedContractTemplate(docs: ContractDoc, signature: ContractSignature)
+
+object ContractTemplateParser {
+  def parse(source: String): Parsed[ParsedContractTemplate] = fastparse.parse(source, parse(_))
+
+  def parse[_: P]: P[ParsedContractTemplate] = P(ContractDocParser.parse ~ "\n" ~ ContractSignatureParser.parse ~ " ".rep.? ~ "=" ~ " ".rep.? ~ &("{")).map(s => ParsedContractTemplate(s._1, s._2))
+}
```
aslesarenko, line 12: I suggest to rename the main object to ContractParser. Will it be used for template or somewhere else it is separate question. And also I think we don't need many object, just put all the code in ContractParser class/object

parsers/shared/src/main/scala/sigmastate/lang/ContractDocParser.scala
```diff
@@ -0,0 +1,124 @@
+package sigmastate.lang
+
+import fastparse._
+import fastparse.NoWhitespace._
+
+/**
+ * Represents a docstring line.
+ */
+private case class DocumentationToken(kind: DocumentationToken.Kind, name: Option[String], body: Option[String])
```
aslesarenko, line 9: Don't think parts of parsed contract AST should be private.

parsers/shared/src/main/scala/sigmastate/lang/ContractDocParser.scala
```diff
@@ -0,0 +1,124 @@
...
+        case Seq() => Seq()
+      }
+    }
+
+    ContractDoc(descriptionText, extractParamDocs(paramTokens))
+  }
+}
+
+/** Parser for docstrings of contract templates. */
+object ContractDocParser {
```
aslesarenko, line 96: I don't think we need a separate object. Does it make sense as an independent parser? All the parsers below can just be moved to ContractParser.

parsers/shared/src/main/scala/sigmastate/lang/ContractSignatureParser.scala
```diff
@@ -0,0 +1,27 @@
...
+case class ContractParam(name: String, tpe: SType, defaultValue: Option[SType#WrappedType])
+
+case class ContractSignature(name: String, params: Seq[ContractParam])
+
+object ContractSignatureParser {
+  import SigmaParser._
+
+  def parse(source: String): Parsed[ContractSignature] = {
+    fastparse.parse(source, parse(_))
+  }
```
aslesarenko, line 16: Same here, all the parsers can move to ContractParser and no need to ContractSignatureParser as independent facade. Or am I missing something?

sc/shared/src/main/scala/sigmastate/lang/SigmaTemplateCompiler.scala
```diff
@@ -0,0 +1,51 @@
...
+import sigmastate.eval.CompiletimeIRContext
+import org.ergoplatform.sdk.Parameter
+
+/** Compiler which compiles Ergo contract templates into a [[ContractTemplate]]. */
+class SigmaTemplateCompiler {
+  def compile(source: String): ContractTemplate = {
+    ContractTemplateParser.parse(source) match {
+      case Parsed.Success(template, index) =>
+        val body = source.slice(index, source.length)
+        val compiler = new SigmaCompiler(NetworkType.Mainnet.networkPrefix.toByte)
```
aslesarenko, line 17: this SigmaTemplateCompiler should be parameterized with network type

aslesarenko (changes requested):

parsers/shared/src/main/scala/sigmastate/lang/ContractDocParser.scala
```diff
@@ -0,0 +1,124 @@
...
+  import DocumentationToken._
+
+  def parse(source: String): Parsed[ContractDoc] = {
+    fastparse.parse(source, parse(_))
+  }
+
+  def parse[_: P]: P[ContractDoc] = P(" ".rep.? ~ "/*" ~ docLine.rep ~ " ".rep.? ~ "*/").map(ContractDoc.apply)
```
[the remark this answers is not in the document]
ross-weir: Changed to `WL` in most places, there was some sections in the docstring parsing that I'm not using it as `WL` consumes doc related chars which is undesirable in that case

parsers/shared/src/main/scala/sigmastate/lang/ContractDocParser.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  private def linePrefix[_: P] = P(" ".rep.? ~ "*" ~ " ".rep.? ~ !"/")
+
+  private def word[_: P] = CharsWhile(c => c != ' ')
```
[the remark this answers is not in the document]
ross-weir: Could you please suggest a replacement, I wasn't able to find one that looked suitable here

sc/shared/src/main/scala/sigmastate/lang/SigmaTemplateCompiler.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Compiles the provided contract source code into a [[ContractTemplate]].
+   * @param source The ErgoScript contract source code.
+   * @return The parsed contract template.
+   */
+  def compile(source: String): ContractTemplate = {
+    ContractParser.parse(source) match {
+      case Parsed.Success(template, index) =>
+        val body = source.slice(index, source.length)
```
aslesarenko, line 23: This looks like a hack. I suggest to make the `body` to be part of ParsedContractTemplate and all the data parsed at once. Then you can create an overloaded `compile` method and pass the parsed body directly to finish compilation.

sc/shared/src/test/scala/sigmastate/lang/SigmaTemplateCompilerTest.scala
```diff
@@ -0,0 +1,55 @@
...
+        |* @param p2 description of the 2nd parameter
+        |* which is pretty complex and on many
+        |* lines to describe functions
+        |* @param param3 the final parameter
+        |* @return
+        |*/
+        |@contract def contractName(p1: Int = 5, p2: String = "default string", param3: Long) = {
+        |  sigmaProp(true)
+        |}""".stripMargin
+    val compiler = SigmaTemplateCompiler(0)
```
aslesarenko, line 30: Use ErgoAddressEncoder.MainnetNetworkPrefix constant here and at similar places. No magic literals.

sc/shared/src/test/scala/sigmastate/lang/SigmaTemplateCompilerTest.scala
```diff
@@ -0,0 +1,55 @@
...
+        |* @param p1 describe p1
+        |* @param p2 description of the 2nd parameter
+        |* which is pretty complex and on many
+        |* lines to describe functions
+        |* @param param3 the final parameter
+        |* @return
+        |*/
+        |@contract def contractName(p1: Int = 5, p2: String = "default string", param3: Long) = {
+        |  sigmaProp(true)
+        |}""".stripMargin
```
aslesarenko, line 29: does it support 
```scala
@contract def contractName(p1: Int = 5, p2: String = "default string", param3: Long) = sigmaProp(true)
```
syntax without braces? Please add the test case.

aslesarenko (changes requested):

sc/shared/src/test/scala/sigmastate/lang/SigmaTemplateCompilerTest.scala
ross-weir: done

sc/shared/src/test/scala/sigmastate/lang/SigmaTemplateCompilerTest.scala
ross-weir: done

sc/shared/src/main/scala/sigmastate/lang/SigmaTemplateCompiler.scala
ross-weir: I _think_ I've done what you're asking for here

parsers/shared/src/main/scala/sigmastate/lang/ContractParser.scala
```diff
@@ -117,8 +117,9 @@ case class ContractSignature(name: String, params: Seq[ContractParam])
  * Includes the parsed docstring preceding the contract as well as the contract signature.
  * @param docs Docstring parsed from the contract.
  * @param signature Signature of the contract.
+ * @param body The code of the contract.
  */
-case class ParsedContractTemplate(docs: ContractDoc, signature: ContractSignature)
+case class ParsedContractTemplate(docs: ContractDoc, signature: ContractSignature, body: String)
```
aslesarenko, line 122: This is better, but still ParsedContractTemplate is only partially parsed.
I think maybe better to parse it completely using SigmaCompiler.parse method.
And then create a new `compileParsed` method in SigmaCompiler, to pass `body` there.

aslesarenko (changes requested):
One last change, otherwise looks good.

aslesarenko:
Also, please fix the tests.

parsers/shared/src/main/scala/sigmastate/lang/ContractParser.scala
ross-weir: Done

ross-weir:
> Also, please fix the tests.

I'm guessing it might have been due to trailing comma which is now removed, I wasn't able to recreate the error locally

aslesarenko (approved):

Issue #681, opened 2020-08-27
Preserve value order in ContextExtension on serialization


greenhat:
### Problem
ContextExtension values order is not preserved in JSON and binary serializaion. Since ContextExtension serialized values used in transaction id generation it should lead to unstable tx id meaning that the same tx would have different ids after JSON serialization/deserialization.

### Solution
Backward compatibility requirements:
- have to be encoded as a dictionary in JSON (no order preservation);
- have to parse unordered values from bytes;

Assuming the order is lost on JSON roundtrip just switching to order preserving structure to store values is not enough.
Since only the order in binary serialization is critical (included in tx id), I suggest to sort values by keys on binary serialization.

aslesarenko:
I assume deserializer should also enforce ordering requirement, otherwise roundtrip property is not preserved. 
This means the new deserializer should throw exception when order is violated.
This cannot be done in v3.x, so I suggest to change to v4.0 milestone.

greenhat:
@aslesarenko I don't see why binary roundtrip property is important. Ordering is important only in binary serialization (for tx id). In other cases, it used as a key-value map, so the order is not important.

aslesarenko:
@greenhat Well, then I lost the point. 
Could you please clarify the description: 
1) which part of the code this issue relates to
2) what exactly is the problem (give an example please)

aslesarenko:
@greenhat I think roundtrip property is important for consistency of the codebase (since all serializers have this property)
Also soundness of the equality is important.
Unsound equality potentially leads to very subtle bugs.

I'm definitely willing to fix this in v4.x.

greenhat:
@aslesarenko Yes, you're right. I agree that binary roundtrip property is important here for consistency and soundness. I meant that it does not fix the issue - getting values in the different order after a JSON roundtrip.
I assume we cannot fix the order in JSON for compatibility reasons.

kushti:
@aslesarenko @greenhat  I think there's no need to enforce ordering on deserialization, so to change the protocol. Enforcing ordering should be done in JSON and on binary serialization likely. but it is up to a client.

For deserialization, we need to document what happens when two value with the same id come (which one is going to be suppressed), to simplify clients development.

greenhat:
@kushti Enforcing ordering in JSON would mean changing its format from map to array. Which seems like a breaking change. Can we afford it? 
I don't mind leaving binary deserialization as is (no enforcing of the order).

[...]

PR #941, opened 2023-11-23
Fix transaction serialization

aslesarenko:
This PR fixes the tx serialization regression bug.


core/shared/src/main/scala/sigma/util/CollectionUtil.scala
```diff
@@ -41,6 +41,18 @@ object CollectionUtil {
     result
   }
 
+  /** Concatenate all the arrays in the sequence. */
+  def concatArrays[T: ClassTag](seq: Traversable[Array[T]]): Array[T] = {
```
kushti, line 45: Adding here method for one test just is overkill, for tests ++ is okay

interpreter/shared/src/main/scala/sigmastate/interpreter/InterpreterContext.scala
```diff
@@ -48,13 +45,17 @@ object ContextExtension {
...
-        .map(_ => (r.getByte(), r.getValue().asInstanceOf[EvaluatedValue[_ <: SType]]))
-      ContextExtension(mutable.LinkedHashMap(ext:_*))
-    }
 
-  }
+      val values = (0 until extSize)
+          .map(_ => (r.getByte(), r.getValue().asInstanceOf[EvaluatedValue[_ <: SType]]))
 
+      // TODO v6.0: enforce ordering e.g. using the code below (see https://github.com/ScorexFoundation/sigmastate-interpreter/issues/681)
+      // if (VersionContext.current.activatedVersion > JitActivationVersion) {
```
kushti, line 53: Such Versioning does not work here, as old clients will be broken after activation. Remove misleading TODO

sdk/shared/src/main/scala/org/ergoplatform/sdk/JsonCodecs.scala
```diff
@@ -245,7 +245,13 @@ trait JsonCodecs {
 
   implicit val contextExtensionDecoder: Decoder[ContextExtension] = Decoder.instance({ cursor =>
     for {
-      values <- cursor.as[mutable.LinkedHashMap[Byte, EvaluatedValue[SType]]]
+      values <- {
+        // TODO v6.0: enforce ordering e.g. using the code below (see https://github.com/ScorexFoundation/sigmastate-interpreter/issues/681)
+        // if (VersionContext.current.activatedVersion > JitActivationVersion) {
+        //   ContextExtension(mutable.LinkedHashMap(values: _*))
```
kushti, line 251: the same TODO issue basically

kushti (changes requested):

core/shared/src/main/scala/sigma/util/CollectionUtil.scala
aslesarenko: removed

interpreter/shared/src/main/scala/sigmastate/interpreter/InterpreterContext.scala
aslesarenko: removed

sdk/shared/src/main/scala/org/ergoplatform/sdk/JsonCodecs.scala
aslesarenko: removed

aslesarenko:
@kushti what should I do with failing serialization roundtrip tests?

kushti:
> @kushti what should I do with failing serialization roundtrip tests?

I think an issue, to start with

aslesarenko:
> I think an issue, to start with

There is an issue https://github.com/ScorexFoundation/sigmastate-interpreter/issues/681
Should I refer to it from the ignored tests?

core/shared/src/main/scala/sigma/util/CollectionUtil.scala
kushti: push the code

kushti:
> > I think an issue, to start with
> 
> There is an issue #681 Should I refer to it from the ignored tests?

I guess vice-versa , so you should refer to the tests from the issue, which is about way different level of visibility. 

The issue is far from ideal , as talks there are mostly about JSON serialization

aslesarenko:
> talks there are mostly about JSON serialization

Actually, the problem is not just with binary, but also with Json, and issue highlights which complexity. 
Will add references.

kushti (approved):

PR #946, opened 2023-12-30
Removing unused code #2

kushti:
In this PR:
* complexity field removed from ErgoTree and serializers
* unused methods removed from Helpers
* some unused imports removed


sdk/shared/src/main/scala/org/ergoplatform/sdk/ReducingInterpreter.scala
```diff
@@ -31,7 +31,7 @@ class ReducingInterpreter(params: BlockchainParameters) extends ErgoLikeInterpre
     * @return data object containing enough data to sign a transaction without Context.
     */
   def reduce(env: ScriptEnv, ergoTree: ErgoTree, context: CTX): ReducedInputData = {
-    val initCost = ergoTree.complexity + context.initCost
```
kushti, line 34: not sure why ergoTree.complexity was there, while in other reduction methods it was removed in 5.0

aslesarenko: This is by mistake, it seems.

aslesarenko (approved):

PR #971, opened 2024-05-02
[sigma-js] Serialization and deserialization of Box in registers

aslesarenko:
In this PR:
- [moved Value to sigma-data](https://github.com/ScorexFoundation/sigmastate-interpreter/commit/60660f20656370e0c191266604211c4866c34db7)
- [tests for Box Value toHex/fromHex](https://github.com/ScorexFoundation/sigmastate-interpreter/pull/971/commits/1177d09640f4db75708edc70cb8bff805bd6de4c)


data/js/src/main/scala/sigma/data/js/Isos.scala
```diff
@@ -0,0 +1,152 @@
...
+import sigma.ast.syntax.GroupElementConstant
+import sigma.ast.{Constant, GroupElementConstant, SType}
+import sigma.data.Iso.isoStringToArray
+import sigma.data.{CGroupElement, Digest32Coll, Iso}
+import sigma.js.GroupElement
+import sigma.serialization.{ErgoTreeSerializer, ValueSerializer}
+
+import scala.scalajs.js
+
+/** Definitions of isomorphisms for sigma-data module. */
```
kushti, line 23: This ScalaDoc can be avoided by just renaming Isos to Isomorphisms.

What are isomorphisms? What are they used for ?

kushti (approved):

data/js/src/main/scala/sigma/data/js/Isos.scala
aslesarenko: > What are isomorphisms? What are they used for ?

Very convenient abstraction to define conversions between different representations of the same data. See `Iso` trait, it is explained there. 
Also this https://en.wikipedia.org/wiki/Isomorphism

Improved the related ScalaDocs.

PR #972, opened 2024-05-02
[6.0.0] Support Header (de)serialization

### LINKED ISSUE #969: Add support for Header (de)serialization
Currently, header serialization is not supported

`CoreDataSerializer` can be modified to support that, see todo there in the `i958-2` branch

kushti:
In this PR, versioned (for Sigma >=6.0) Header type (de)serialization is implemented. 

For that, ErgoHeader from Ergo codebase, its serializer and related classes and their respective serializers are copied from the original repository. Thus there is a need to import those classes from Sigma in future versions of Ergo node codebase, see https://github.com/ergoplatform/ergo/issues/2163

close #969


interpreter/shared/src/test/scala/sigma/serialization/DataSerializerSpecification.scala
```diff
@@ -23,55 +23,74 @@ import scala.reflect.ClassTag
...
+
+      val randomPrefix = arrayGen[Byte].sample.get
+      val r2 = SigmaSerializer.startReader(randomPrefix ++ bytes, randomPrefix.length)
+      val res2 = DataSerializer.deserialize(tpe, r2)
+      res2 shouldBe obj
+    }
+
+    withVersion match {
+      case Some(ver) =>
+        VersionContext.withVersions(ver, 1) {
```
aslesarenko, line 55: Should be `0` since script versions are zero based.
Also  better pass by named argument, as the meaning is not clear without IDE hints.

interpreter/shared/src/test/scala/sigma/serialization/SerializationSpecification.scala
```diff
@@ -26,10 +27,20 @@ trait SerializationSpecification extends AnyPropSpec
...
-    predefinedBytesTestNotFomZeroElement(bytes, v)
+  protected def roundTripTest[V <: Value[_ <: SType]](v: V, withVersion: Option[Byte] = None): Assertion = {
+    def test() = {
+      val bytes = ValueSerializer.serialize(v)
+      predefinedBytesTest(v, bytes)
+      predefinedBytesTestNotFomZeroElement(bytes, v)
+    }
+    withVersion match {
+      case Some(ver) =>
+        VersionContext.withVersions(ver, 1) {
```
aslesarenko, line 38: Same as above, should be 0.

interpreter/shared/src/test/scala/sigma/serialization/generators/ObjectGenerators.scala
```diff
@@ -707,8 +707,10 @@ trait ObjectGenerators extends TypeGenerators
     powNonce <- nonceBytesGen
     powDistance <- arbBigInt.arbitrary
     votes <- minerVotesGen
-  } yield CHeader(id, version, parentId, adProofsRoot, stateRoot, transactionRoot, timestamp, nBits,
-    height, extensionRoot, minerPk.toGroupElement, powOnetimePk.toGroupElement, powNonce, powDistance, votes)
+    unparsedBytes <- collOfRange(0, 32, arbByte.arbitrary)
+  } yield CHeader(version, parentId, adProofsRoot, stateRoot.digest, transactionRoot, timestamp, nBits,
+    height, extensionRoot, minerPk.toGroupElement, powOnetimePk.toGroupElement, powNonce, powDistance, votes,
+    if(version > HeaderVersion.Interpreter60Version){ unparsedBytes } else {Colls.emptyColl[Byte]})
```
aslesarenko, line 713: This condition is likely related to my previous comment about unparsedBytes parsing.

sdk/shared/src/main/scala/org/ergoplatform/sdk/JsonCodecs.scala
```diff
@@ -146,8 +146,10 @@ trait JsonCodecs {
       powNonce <- cursor.downField("powNonce").as[Coll[Byte]]
       powDistance <- cursor.downField("powDistance").as[sigma.BigInt]
       votes <- cursor.downField("votes").as[Coll[Byte]]
-    } yield new CHeader(id, version, parentId, adProofsRoot, stateRoot, transactionsRoot, timestamp, nBits,
-      height, extensionRoot, SigmaDsl.decodePoint(minerPk), SigmaDsl.decodePoint(powOnetimePk), powNonce, powDistance, votes)
+      unparsedBytes <- cursor.downField("unparsedBytes").as[Option[Coll[Byte]]]
```
aslesarenko, line 149: If it is encoded as Coll, then it should be decoded as Call. Maybe it should be saved as Option as well.

sdk/shared/src/test/scala/org/ergoplatform/sdk/DataJsonEncoderSpecification.scala
```diff
@@ -22,29 +22,48 @@ import scala.reflect.ClassTag
...
+  def roundtrip[T <: SType](obj: T#WrappedType, tpe: T, withVersion: Option[Byte] = None) = {
+    def test() = {
+      val json = DataJsonEncoder.encode(obj, tpe)
+      val res = DataJsonEncoder.decode(json)
+      res shouldBe obj
+    }
+
+    withVersion match {
+      case Some(ver) =>
+        VersionContext.withVersions(ver, 1) {
```
aslesarenko, line 34: script version start from 0.

sdk/shared/src/test/scala/org/ergoplatform/sdk/DataJsonEncoderSpecification.scala
```diff
@@ -187,25 +214,44 @@ class DataJsonEncoderSpecification extends SerializationSpecification {
...
+
+        // not supported case
+        an[SerializerException] should be thrownBy {
+          DataJsonEncoder.encode(SigmaDsl.Colls.fromItems(TupleColl(x, x, x)).asWrappedType, SCollection(STuple(tpe, tpe, tpe)))
+        }
       }
     }
+
+    if (tpe == SHeader)  {
+      VersionContext.withVersions(VersionContext.V6SoftForkVersion, 1) {
```
aslesarenko, line 236: start from 0

aslesarenko (changes requested):

core/shared/src/main/scala/sigma/SigmaDsl.scala
```diff
@@ -459,6 +459,18 @@ trait Header {
...
   def votes: Coll[Byte] //3 bytes
+
+  /** Bytes which are coming from future versions of the protocol, so
+    * their meaning is not known to current version of Sigma, but they
+    * are stored to get the same id as future version users.
+    */
+  def unparsedBytes: Coll[Byte]
+
+  /**
+    * @return bytes without proof of work, needed for working to get the proof on
```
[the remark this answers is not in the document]
kushti: done

data/shared/src/main/scala/org/ergoplatform/ErgoHeader.scala
```diff
@@ -0,0 +1,159 @@
...
+  *
+  * @param pk - miner public key. Should be used to collect block rewards
+  * @param w  - one-time public key. Prevents revealing of miners secret
+  * @param n  - nonce (8 bytes)
+  * @param d  - distance between pseudo-random number, corresponding to nonce `n` and a secret,
+  *           corresponding to `pk`. The lower `d` is, the harder it was to find this solution.
+  */
+case class AutolykosSolution(pk: EcPointType,
+                             w: EcPointType,
+                             n: Array[Byte],
```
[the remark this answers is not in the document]
kushti: case class reworked into class

sdk/shared/src/main/scala/org/ergoplatform/sdk/JsonCodecs.scala
kushti: this field made optional to be compatible with previous versions of clients. Thus we provide them with the new field but do not require it.

interpreter/shared/src/test/scala/sigma/serialization/generators/ObjectGenerators.scala
kushti: Yes, unparsed bytes shouldn't be generated for <= 6

sdk/shared/src/test/scala/org/ergoplatform/sdk/DataJsonEncoderSpecification.scala
kushti: done

interpreter/shared/src/test/scala/sigma/serialization/SerializationSpecification.scala
kushti: done

data/shared/src/main/scala/org/ergoplatform/HeaderWithoutPow.scala
```diff
@@ -0,0 +1,139 @@
...
+    val height = r.getUInt().toIntExact
+    val votes = r.getBytes(3)
+
+    // For block version >= 2, a new byte encodes length of possible new fields.
+    // If this byte > 0, we read new fields but do nothing, as semantics of the fields is not known.
+    val unparsedBytes = if (version > HeaderVersion.InitialVersion) {
+      val newFieldsSize = r.getUByte()
+      if (newFieldsSize > 0 && version > HeaderVersion.Interpreter60Version) {
+        // new bytes could be added only for block version >= 5
+        r.getBytes(newFieldsSize)
```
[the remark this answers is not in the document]
kushti: ??? 
When nodes with V7 activated will start to produce V7 headers, v6 (actually >= 5.0.22) nodes will read and store unparsed bytes. 

see https://github.com/ergoplatform/ergo/pull/2151

data/shared/src/main/scala/org/ergoplatform/HeaderWithoutPow.scala
```diff
@@ -0,0 +1,139 @@
...
+}
+
+
+object DifficultySerializer extends SigmaSerializer[Long, Long] {
+
+  /** Parse 4 bytes from the byte array (starting at the offset) as unsigned 32-bit integer in big endian format. */
+  def readUint32BE(bytes: Array[Byte]): Long = ((bytes(0) & 0xffL) << 24) | ((bytes(1) & 0xffL) << 16) | ((bytes(2) & 0xffL) << 8) | (bytes(3) & 0xffL)
+
+  def uint32ToByteArrayBE(value: Long): Array[Byte] = {
+    Array(0xFF & (value >> 24), 0xFF & (value >> 16), 0xFF & (value >> 8), 0xFF & value).map(_.toByte)
```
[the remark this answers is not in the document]
kushti: ok, made a comment in https://github.com/ergoplatform/ergo/issues/2163 , does not make sense to diverge in copy-pasted code now

sdk/shared/src/test/scala/org/ergoplatform/sdk/DataJsonEncoderSpecification.scala
kushti: done

kushti:
@ergomorphic comments addressed, please make another pass!

kushti:
@ergomorphic comments addressed, please make another pass!

data/shared/src/main/scala/org/ergoplatform/HeaderWithoutPow.scala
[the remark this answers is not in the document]
aslesarenko: Still don't understand why the parsing was changed. 
Currenty the byte is reserved and is expected to be zero, but I don't see where it is enforced.
I can put 2 and then put 2 extra bytes and it will be valid v5.0 header and will be parsed.
So, the only missing part is storing what was parsed in the HeaderWithoutPow instance and update serialization to save extra bytes.
The name `unparsedBytes` is somewhat misleading, because they are literally parsed, but obscure/opaque.
So, `extraBytes` would be better name, and then in future versions they can receive some meaning without stopping being "extra".

interpreter/shared/src/test/scala/sigma/serialization/generators/ObjectGenerators.scala
aslesarenko: Yeah, see my second comment there.

sdk/shared/src/main/scala/org/ergoplatform/sdk/JsonCodecs.scala
aslesarenko: Then the roundtrip property will only be partial. Shouldn't the "we provide them" part be decoupled from encoding? In other words, if we want the field to become required, this can be enforced in the object.

aslesarenko (changes requested):
Please see additional comments.

sdk/shared/src/main/scala/org/ergoplatform/sdk/JsonCodecs.scala
kushti: why should it be required now, when at least until v7 bytes will not be used? Why to break clients now?

kushti:
@ergomorphic comments addressed, please make another pass!

data/shared/src/main/scala/org/ergoplatform/HeaderWithoutPow.scala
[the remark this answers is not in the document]
aslesarenko: How POW is related to serialization? Header can be serializable, but not valid for protocol, so validation is a layer on top of serialization.

sdk/shared/src/main/scala/org/ergoplatform/sdk/JsonCodecs.scala
aslesarenko: > why should it be required now

You said "we provide them with the new field but do not require it" maybe I misunderstood it. Anyway, I think each encoder should have proper roundtrip property and here it is violated.

data/shared/src/main/scala/org/ergoplatform/HeaderWithoutPow.scala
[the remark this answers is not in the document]
kushti: PoW is generated over HeaderWithoutPoW bytes.

sdk/shared/src/main/scala/org/ergoplatform/sdk/JsonCodecs.scala
kushti: It is violated when (and only) unparsed bytes are not provided, yes, which is perfectly showing what is done in order to support existing clients. Considering that breaking changes may happen in years or not happen at all, there is no reason to break the clients.

data/shared/src/main/scala/org/ergoplatform/HeaderWithoutPow.scala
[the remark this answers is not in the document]
aslesarenko: If version == 6, and newFieldsSize > 0, then 
- v5.0 node will parse the header, read all the bytes, but will not save them in the header instance.
- v6.0 node will not read the bytes leaving reader in an invalid state which will lead to failure later
Thus, we allow newFieldsSize > 0 in v5.0 and forbid it in v6.0, right?

sdk/shared/src/main/scala/org/ergoplatform/sdk/JsonCodecs.scala
aslesarenko: By "clients" you mean any external code that directly use this headerDecoder? Then ScalaDoc would be helpful.

data/shared/src/main/scala/org/ergoplatform/HeaderWithoutPow.scala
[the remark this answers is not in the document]
kushti: V 5.0 will parse the header, but PoW check will be failed after. So consensus-wise both 5.0 and 6.0 will reject the header, only with different errors.

kushti:
@ergomorphic comments addressed, please make another pass!

data/shared/src/main/scala/org/ergoplatform/ErgoHeader.scala
```diff
@@ -19,10 +19,10 @@ import sigma.serialization.{GroupElementSerializer, SigmaByteReader, SigmaByteWr
...
   *           corresponding to `pk`. The lower `d` is, the harder it was to find this solution.
   */
-case class AutolykosSolution(pk: EcPointType,
-                             w: EcPointType,
-                             n: Array[Byte],
-                             d: BigInt) {
+class AutolykosSolution(val pk: EcPointType,
+                        val w: EcPointType,
+                        val n: Array[Byte],
+                        val d: BigInt) {
```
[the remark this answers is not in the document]
kushti: It really hadn't with Array[Byte] argument , added equals and hashCode, not sure if can be used anywhere though, as ErgoHeader has custom equality and unlikely solutions are compared anywhere

kushti:
@ergomorphic comments addressed, please make another pass!

kushti:

interpreter/shared/src/test/scala/sigma/serialization/generators/ObjectGenerators.scala
kushti: addressed then

aslesarenko (approved):

PR #975, opened 2024-05-06
[6.0.0] Header.bytes method implementation

kushti:


kushti:
Closed as header.bytes can be replaced by deserializeTo[] from #979

Issue #600, opened 2019-09-19
executeFromVar error on costing


greenhat:
`executeFromVar[SigmaProp](21)`
with context:
```
      val script: SigmaPropValue = GT(Height, IntConstant(1)).toSigmaProp
      val scriptBytes = ValueSerializer.serialize(script)
      ContextExtension(Map(21.toByte -> ByteArrayConstant(scriptBytes)))
```
throw on `SigmaTestingCommons.compile`:
```
Variable executeFromVar not found in environment Map(maxSizeColl -> s19, coll10 -> s28, alice4 -> s39, alice -> s43, coll100 -> s52, coll5 -> s61, alice2 -> s65, coll1000 -> s74, alice3 -> s78)
```

aslesarenko:
@greenhat How to reproduce?

greenhat:
@aslesarenko I made a PR with a failing test - https://github.com/ScorexFoundation/sigmastate-interpreter/pull/612

jozanek:
@kushti this was fixed in #976?

Issue #443, opened 2019-03-18
Finish executeFromX implementation


scalahub:
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/5467a667a6f7fa2e3e25a40f39990f862c2dde57/src/test/scala/sigmastate/utxo/examples/ExecuteFromExamplesSpecification.scala#L36

This line gives error:
     `Variable executeFromVar not found in environment Map(pkA -> s74)`
Full error here:
https://gist.github.com/scalahub/c80c2950f04a21365d0f91c8a4e74733

kushti:
both executeFomVar / executeFromSelfReg done now

PR #976, opened 2024-05-06
Finish executeFromVar implementation 

kushti:
This PR contains complete implementation (with tests) of executeFromVar support in the ErgoScript compiler.  

Related to #443, #600


sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
```diff
@@ -157,6 +159,50 @@ class BasicOpsSpecification extends CompilerTestingCommons
...
+    val script = DeserializeContext(21.toByte, SSigmaProp)
+    val scriptBytes = ValueSerializer.serialize(script)
+    val customExt = Seq(21.toByte -> ByteArrayConstant(scriptBytes))
+    an [Exception] should be thrownBy test("executeFromVar", env, customExt,
+      "executeFromVar[SigmaProp](21)",
+      null,
+      true
+    )
+  }
+
```
aslesarenko, line 205: Please also add test to LanguageSpecificationV6. 
DeserializeContext was not covered there before because it was not supported by compiler, and scripts are required there.
Now it can be added.

aslesarenko (changes requested):

sc/shared/src/main/scala/sigma/compiler/ir/Base.scala
```diff
@@ -203,10 +204,11 @@ abstract class Base { thisIR: IRContext =>
...
 
-  /** Logical AND between two pattern matches of the save value `x`.
-    * Can be used to construct patterns like `case P1 && P2 => ...` */
-  object && {
-    def unapply[T](x: T): Option[(T,T)] = Some((x, x))
+  /**
+    * Def done in order to carry on DeserializeContext through stages of compilation intact
+    */
+  case class DeserializeContextDef[V <: SType](d: DeserializeContext[V], e: Elem[V]) extends Def[V] {
+    override def resultType: Elem[V] = e
```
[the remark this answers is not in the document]
kushti: What does that mean ? What do you propose?

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
kushti: ??? There is nothing 6.0 specific  or versioned in general specific in this PR.

kushti:
@aslesarenko comments addressed, please make another pass

sc/shared/src/main/scala/sigma/compiler/ir/Base.scala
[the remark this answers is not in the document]
aslesarenko: It should be `e: Elem[V#WrappedType]`.

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
aslesarenko: DeserializeContext in not covered in LanguageSpecificationV6, above I explained why. Now it can be covered, but also probably was not used in practice due to lack of ErgoScript support. Now it will be available, so better be covered in the specification.

kushti: Sorry, I do not understand, what do you propose to test ? I mean, which `DeserializeContext` properties, and why in LanguageSpecificationV6 specifically. All other tests in this specification are testing versioned behavior of features added in 6.0. And ScalaDoc for `LanguageSpecificationBase` is also proposing to test versioned behavior in `LanguageSpecification` tests. Have you changed ScalaDoc in some PR maybe ?

aslesarenko: There is no contradiction, every feature can be tested for many versions, but at least for one. Importantly, all components of the interpreter is tested, parsing, compiler stages, tree serialization, evaluation on different inputs, positive and negative cases, expected values are checked. That is why there should be tests for each language feature (operation) here. Versioning is not a central part of it, but important capability.

kushti: Then I guess `DeserializeContext`  should be in LanguageSpecificationV3 along with all features existing since the mainnet launch. But then you need to open an issue at least about doing what you are proposing for all the nodes existing since the launch, to avoid chaotic implementation of 2-3 nodes before it will be forgotten forever. And the question about properties is still  not answered. A typical check in LSV6 is like 

```
      lazy val bitAnd = newFeature(
      { (x: (Int, Int)) => x._1 & x._2 },
      "{ (x: (Int, Int)) => x._1 & x._2 }",
      sinceVersion = VersionContext.V6SoftForkVersion)
      forAll { x: Int =>
        Seq(toAbs).foreach(_.checkEquality(x))
      }
      forAll { x: (Int, Int) =>
        Seq(compareTo, bitOr, bitAnd).foreach(_.checkEquality(x))
      }
```

So a check is testing that for V6 a new features has some behavior. What do you want to check about `DeserializeContext` in LSV* ?

aslesarenko: Yes, it is an existing feature so can be added to LSV5 as existingFeature(...).
I believe described above what I "want to check" by creating a new `property(){...}` for  DeserializeContext.
Most of the other cases like this are either not known or already recorded as issues. Don't think this PR depend on this.

kushti: Trying to implement it , I think, there is inherent issue in `SigmaDslTesting` still, talking about DeserializeContext node, that is, the DSL does not have a function corresponding to executeFromVar and the node, e.g. I dont see how anything but trivial substitutions with evaluated data (via `getVar` can be done), for example, the following test in BasicOpsSpecification:

```
    val bytes = Slice(ByteArrayConstant(Colls.fromArray(Array.fill(5)(1.toByte))), IntConstant(1), IntConstant(3))
    val valueBytes = ValueSerializer.serialize(bytes)
    val customExt = Seq(21.toByte -> ByteArrayConstant(valueBytes))
    test("executeFromVar", env, customExt,
      "{val ba = executeFromVar[Coll[Byte]](21); ba.size == 2 }",
     ... 
```

does not have an alternative due to that

sc/shared/src/main/scala/sigma/compiler/ir/Base.scala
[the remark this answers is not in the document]
kushti: Isn't the type right for e.g. Coll[Byte]: CollElem<Coll[Byte]> ?

kushti:
@aslesarenko please check new notes

aslesarenko (approved):

PR #977, opened 2024-05-09
Correct environment for Template Compiler

### LINKED ISSUE #966: SigmaTemplateCompiler (EIP-5) does not take non-parameter constants into account
Take the following example into account:
`
@contract isItTime(theTime: Long) = {
  sigmaProp(
    theTime > CONTEXT.preHeader.timestamp &&
    1000L < CONTEXT.preHeader.timestamp
  )
}
`

Following issues occur:

1. TyperException: Cannot assign type for variable 'theTime' because it is not found in env Map
~~2. The resulting ContractTemplate (assuming issue 1 is fixed) only contains the parameter constant (naturally with index 0) but not the other constants present in the contract (such as in this case LongConstant(1000)).~~

aslesarenko:
Closes #966 
In this PR:
- fixed parsing of ContractParam
- default parameter values used to create environment for Compiler
- default env is merged with explicitly provided
- more tests for SigmaTemplateCompiler

### DIFF
--- a/sc/shared/src/test/scala/sigmastate/lang/SigmaTemplateCompilerTest.scala
+++ b/sc/shared/src/test/scala/sigmastate/lang/SigmaTemplateCompilerTest.scala
@@ -2,13 +2,14 @@ package sigmastate.lang
 
 import org.ergoplatform.ErgoAddressEncoder
 import org.ergoplatform.sdk.Parameter
-import org.scalatest.matchers.should.Matchers
 import org.scalatest.propspec.AnyPropSpec
 import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
-import sigma.ast.{IntConstant, SInt, SLong, SString, StringConstant}
+import sigma.ast.{SBoolean, SInt, SLong, SString}
+import sigma.exceptions.TyperException
 import sigmastate.eval.CompiletimeIRContext
+import sigmastate.interpreter.Interpreter.ScriptEnv
 
-class SigmaTemplateCompilerTest extends AnyPropSpec with ScalaCheckPropertyChecks with Matchers {
+class SigmaTemplateCompilerTest extends AnyPropSpec with ScalaCheckPropertyChecks with LangTests {
   property("compiles full contract template") {
     val source =
       """/** This is my contracts description.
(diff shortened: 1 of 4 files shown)


Luivatra:
By adding the parameters in the template compile step wont this end up with an ergotree with no ConstantPlaceholders for the parameters? I don't think the SigmaTemplateCompiler can differentiate between normal constants and parameter values.

Luivatra:
By adding the parameters in the template compile step wont this end up with an ergotree with no ConstantPlaceholders for the parameters? I don't think the SigmaTemplateCompiler can differentiate between normal constants and parameter values.

sc/shared/src/main/scala/sigmastate/lang/SigmaTyper.scala
```diff
@@ -19,6 +18,7 @@ import scala.collection.mutable.ArrayBuffer
   */
 class SigmaTyper(val builder: SigmaBuilder,
```
kushti, line 19: ScalaDoc (class args description) is missed

sc/shared/src/test/scala/sigmastate/lang/SigmaTemplateCompilerTest.scala
```diff
@@ -73,16 +72,138 @@ class SigmaTemplateCompilerTest extends AnyPropSpec with ScalaCheckPropertyCheck
...
+    )
   }
+
+  property("fails when the parameter is not provided") {
+    // NOTE: parameter `condition` is not provided */
+    val source      =
+      """/**/
+       |@contract def contractName(low: Int = 0, high: Int) = sigmaProp(low < HEIGHT && HEIGHT < high) && condition
+       |""".stripMargin
+    val env: ScriptEnv = Map.empty // no value for "high"
```
kushti, line 200: env is not used

kushti (approved):

sc/shared/src/main/scala/sigmastate/lang/SigmaTyper.scala
aslesarenko: done

sc/shared/src/test/scala/sigmastate/lang/SigmaTemplateCompilerTest.scala
aslesarenko: fixed

PR #979, opened 2024-05-13
[6.0] Global.deserializeTo[] method

kushti:
This PR is introducing  new Global.deserializeRaw method which is allowing to deserialize a typed object from underlying bytes. It is compatible (has roundtrip with) Global.serialize from #989 .

example:

```
{
     val src = getVar[Coll[Short]](21).get
     val ba = serialize(src)
     val restored = deserializeTo[Coll[Short]](ba)
     src == restored
}
```

### DIFF
--- a/data/shared/src/main/scala/sigma/ast/methods.scala
+++ b/data/shared/src/main/scala/sigma/ast/methods.scala
@@ -1754,6 +1754,7 @@ case object SHeaderMethods extends MonoTypeMethods {
   lazy val powDistanceMethod      = propertyCall("powDistance", SBigInt, 14, FixedCost(JitCost(10)))
   lazy val votesMethod            = propertyCall("votes", SByteArray, 15, FixedCost(JitCost(10)))
 
+  // methods added in 6.0 below
   // cost of checkPoW is 700 as about 2*32 hashes required, and 1 hash (id) over short data costs 10
   lazy val checkPowMethod = SMethod(
     this, "checkPow", SFunc(Array(SHeader), SBoolean), 16, FixedCost(JitCost(700)))
@@ -1824,6 +1825,15 @@ case object SGlobalMethods extends MonoTypeMethods {
     .withInfo(Xor, "Byte-wise XOR of two collections of bytes",
       ArgInfo("left", "left operand"), ArgInfo("right", "right operand"))
 
+  private val deserializeCostKind = PerItemCost(baseCost = JitCost(30), perChunkCost = JitCost(20), chunkSize = 32)
+
+  lazy val deserializeToMethod = SMethod(
+    this, "deserializeTo", SFunc(Array(SGlobal, SByteArray), tT, Array(paramT)), 4, deserializeCostKind, Seq(tT))
+    .withIRInfo(MethodCallIrBuilder,
+      javaMethodOf[SigmaDslBuilder, Coll[Byte], RType[_]]("deserializeTo"))
+    .withInfo(MethodCall, "Deserialize provided bytes into an object of requested type",
+      ArgInfo("first", "Bytes to deserialize"))
+
   /** Implements evaluation of Global.xor method call ErgoTree node.
     * Called via reflection based on naming convention.
     * @see SMethod.evalMethod, Xor.eval, Xor.xorWithCosting
@@ -1858,6 +1868,15 @@ case object SGlobalMethods extends MonoTypeMethods {
     this, "decodeNbits", SFunc(Array(SGlobal, SLong), SBigInt), 7, DecodeNBitsCost)
     .withIRInfo(MethodCallIrBuilder)
     .withInfo(MethodCall, "Decode nbits-encoded big integer number", ArgInfo("nbits", "NBits-encoded argument"))
+
+  def deserializeTo_eval(mc: MethodCall, G: SigmaDslBuilder, bytes: Coll[Byte])
+                        (implicit E: ErgoTreeEvaluator): Any = {
+    val tpe = mc.tpe
+    val cT = stypeToRType(tpe)
+    E.addSeqCost(deserializeCostKind, bytes.length, deserializeToMethod.opDesc) { () =>
+      G.deserializeTo(bytes)(cT)
+    }
+  }
 
   lazy val serializeMethod = SMethod(this, "serialize",
     SFunc(Array(SGlobal, tT), SByteArray, Array(paramT)), 3, DynamicCost)
@@ -1894,6 +1913,7 @@ case object SGlobalMethods extends MonoTypeMethods {
         groupGeneratorMethod,
         xorMethod,
         serializeMethod,
+        deserializeToMethod,
         encodeNBitsMethod,
         decodeNBitsMethod,
         fromBigEndianBytesMethod
--- a/data/shared/src/main/scala/sigma/data/CSigmaDslBuilder.scala
+++ b/data/shared/src/main/scala/sigma/data/CSigmaDslBuilder.scala
@@ -1,15 +1,18 @@
 package sigma.data
 
 import debox.cfor
-import org.ergoplatform.ErgoBox
+import org.ergoplatform.{ErgoBox, ErgoHeader}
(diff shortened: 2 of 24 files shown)


data/shared/src/main/scala/sigma/serialization/MethodCallSerializer.scala
```diff
@@ -23,6 +25,9 @@ case class MethodCallSerializer(cons: (Value[SType], SMethod, IndexedSeq[Value[S
     w.putValue(mc.obj, objInfo)
     assert(mc.args.nonEmpty)
     w.putValues(mc.args, argsInfo, argsItemInfo)
+    if(mc.method.name == SGlobalMethods.deserializeRawMethod.name){
+      w.putType(mc.tpe, typeInfo)
+    }
```
aslesarenko, line 30: Better leave MethodCall unchanged and use PolyMethodCall as described in https://github.com/ScorexFoundation/sigmastate-interpreter/issues/980
This will also give both simplicity of code and efficiency of execution, because specialization will happen at compile time and most of the method calls are simple method calls.

interpreter/shared/src/test/scala/sigma/serialization/DataSerializerSpecification.scala
```diff
@@ -132,6 +132,20 @@ class DataSerializerSpecification extends SerializationSpecification {
...
           t.getMessage.contains(s"BigInt value doesn't not fit into ${SBigInt.MaxSizeInBytes} bytes")
     })
+  }
+
+  property("header roundtrip") {
+    VersionContext.withVersions(VersionContext.V6SoftForkVersion, 1) {
+      forAll { x: Header => roundtrip[SHeader.type](x, SHeader) }
+    }
 
+    an[SerializerException] should be thrownBy (
```
aslesarenko, line 142: For negative cases there is `assertExceptionThrown` which allows testing not only exception type but also the message. See other usages of it as an example.

sc/shared/src/main/scala/sigmastate/eval/GraphBuilding.scala
```diff
@@ -19,6 +20,25 @@ import sigma.serialization.OpCodes
 
 import scala.collection.mutable.ArrayBuffer
 
+
+case class DeserializeRawBytes[V <: SType](bytes: Value[SByteArray], tpe: V) extends NotReadyValue[V] {
```
aslesarenko, line 24: It's not necessary to introduce new operation with opcode for each method, MethodCall is enough and can be translated to MethodCall from GraphBuilding.

sc/shared/src/main/scala/sigmastate/eval/GraphBuilding.scala
```diff
@@ -536,6 +556,9 @@ trait GraphBuilding extends SigmaLibrary { IR: IRContext =>
         val e = stypeToElem(optTpe.elemType)
         ctx.getVar(id)(e)
 
+      case DeserializeRawBytes(bytes, tpe) =>
+        sigmaDslBuilder.deserializeRaw(asRep[Coll[Byte]](eval(bytes)))(stypeToElem(tpe))
```
aslesarenko, line 560: This is not necessary.

sc/shared/src/main/scala/sigmastate/lang/SigmaTyper.scala
```diff
@@ -133,6 +133,18 @@ class SigmaTyper(val builder: SigmaBuilder,
...
+            case Some(method) =>
+              val srcCtx = sel.sourceContext
+              if(method.name == SGlobalMethods.deserializeRawMethod.name) {
+                DeserializeRawBytes(args(0).asInstanceOf[Value[SByteArray]], rangeTpe)
+              } else {
+                processGlobalMethod(srcCtx, method, newArgs)
+              }
+            case _ =>
+              error(s"Cannot find Global method: $n", bound.sourceContext)
+          }
```
aslesarenko, line 147: This shouldn't be necessary.

data/shared/src/main/scala/org/ergoplatform/HeaderWithoutPow.scala
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    HeaderWithoutPow(version, parentId, ADProofsRoot, stateRoot, transactionsRoot, timestamp,
+      nBits, height, extensionHash, votes, unparsedBytes)
+  }
+
+}
+
+
+object DifficultySerializer extends SigmaSerializer[Long, Long] {
```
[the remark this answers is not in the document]
kushti: ??? 
No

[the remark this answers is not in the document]
aslesarenko: ok, just saying

data/shared/src/main/scala/sigma/data/CSigmaDslBuilder.scala
```diff
@@ -200,6 +204,29 @@ class CSigmaDslBuilder extends SigmaDslBuilder { dsl =>
...
+          case Right(prop) =>
+            if(prop.isInstanceOf[EvaluatedValue[_]]) {
+              prop.asInstanceOf[EvaluatedValue[_]].value
+            } else {
+              throw new InvalidType(s"Cannot deserialize($bytes): prop is not evaluated: $prop}")
+            }
+        } */
+      case _ =>
+        val reader = new SigmaByteReader(new VLQByteBufferReader(ByteBuffer.wrap(bytes.toArray)), new ConstantStore(), false)
+        DataSerializer.deserialize(tpe, reader)
```
aslesarenko, line 226: One problem with having special cases for Int, Long, BigInt is that their format will be different whey they are top level vs they are in a tuple or Coll, i.e. `serialize((intValue, longValue)) != serialize(intValue) ++ serialize(longValue)` and also `serialize(intValue) != serialize(Coll(intValue)).drop(1)`.
In other words, when the value inside a structure, it is suddenly has a different format.

I think it is better to just use DataSerializer, keep the implementation simple and generic, and forget about potential differences with other existing methods which return bytes of something.

kushti: It seems you have not understood the proposal of mine expressed over Discord:

> 
> Investigating using DataSerializer, with it round-trip property with existing serializers (longtoByteArray, .propBytes etc) would be broken. Which would lead to following possible consequences:
> 
>     deserialization of those those types (Long, BigInt, SigmaProp) and similar (Short, Int) is corresponding to existing serializers, for other options DataSerializer is used
> 
> 
>     new "serialize: SAny -> SByteArray" option is introduced , with DataSerializer under the hood, but then there are two coexisting options, old one (longToByteArray/byteArrayToLong), and new one serialize/deserialize, for round-tripping, with the exception for .propBytes which can be deserialized via skipping first two bytes though, like " { val bytes = p1.propBytes; val ba = bytes.slice(2, bytes.size); val prop = Global.deserialize[SigmaProp](ba)}
> 
> 
> in case of serialize() (which is the better option I support), there is no need for header.bytes. avlTree.bytes, numeric.bytes (and the latter option would be confusing, as producing different bytes from *ToByteArray). Thus you can take serialize() instead of Numeric.toBytes

aslesarenko (approved):
LGTM, see also 1 comment

data/shared/src/main/scala/sigma/ast/methods.scala
```diff
@@ -1824,6 +1825,16 @@ case object SGlobalMethods extends MonoTypeMethods {
     .withInfo(Xor, "Byte-wise XOR of two collections of bytes",
       ArgInfo("left", "left operand"), ArgInfo("right", "right operand"))
 
+  private val deserializeCostKind = PerItemCost(
+    baseCost = JitCost(20), perChunkCost = JitCost(7), chunkSize = 128)
```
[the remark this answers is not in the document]
kushti: Makes sense, raised cost to 
baseCost = JitCost(30), perChunkCost = JitCost(20), chunkSize = 32

PR #986, opened 2024-05-14
Simplify IRContext

aslesarenko:
In this PR, IRContext cake is simplified by removing unnecessary level of abstractions, which will simplify adding new operations to ErgoScript:
- removed Library, SigmaLibrary, Scalan, SpecialPredef traits 
- package structure in `sc` module reorganized and simplified


kushti (approved):
What `ReflectionGenerator.scala` is doing? It is in the testing codebase but not a test, and ScalaDoc is missed. As it is touched, ScalaDoc is needed.

PR #990, opened 2024-05-17
[6.0] SOption[T] (de)serialization support

### LINKED ISSUE #659: Serialize Option[T] in DataSerializer

kushti:
In this PR, versioned implementation of SOption[T] (de)serialization is done. 

Close #659 

Versioning can be changed, other parts are stable I suppose.


interpreter/shared/src/test/scala/sigma/serialization/DataSerializerSpecification.scala
```diff
@@ -92,6 +92,31 @@ class DataSerializerSpecification extends SerializationSpecification {
...
+
+    VersionContext.withVersions(VersionContext.V6SoftForkVersion, 1) {
+      forAll { in: T#WrappedType =>
+        roundtrip[SType](Some(in).asWrappedType, SOption(tpe))
+        roundtrip[SOption[T]](None, SOption(tpe))
+        roundtrip[SOption[T]](Some(in), SOption(tpe))
+        roundtrip[SOption[SCollection[T]]](Some(Colls.fromItems(in)(tT)), SOption(SCollectionType(tpe)))
+        roundtrip[SCollection[SOption[T]]](Colls.fromItems(Option(in), None.asInstanceOf[Option[T#WrappedType]])(OptionType(tT)), SCollectionType(SOption(tpe)))
+        roundtrip[SOption[SOption[T]]](None, SOption(SOption(tpe)))
+        roundtrip[SOption[SOption[T]]](Some(Some(in)), SOption(SOption(tpe)))
```
aslesarenko, line 115: please add `Some(None)` case.

aslesarenko (approved):
LGTM, please address the comment and it can be merged.

interpreter/shared/src/test/scala/sigma/serialization/DataSerializerSpecification.scala
kushti: done

PR #995, opened 2024-05-22
[6.0] Fix SubstConstants: serialize ErgoTree size

### LINKED ISSUE #994: Fix SubstContants: serialize ErgoTree size
This issue is to fix bug with SubstConstants operation noticed by @Luivatra.

"the ergoscript function substConstants does not include the size/contentlength uint in the output even if the input bytes have it defined"

aslesarenko:
Closes #994


kushti (changes requested):
This PR contains serious changes to another 6.0 method, serialize ( introducing CheckMinimalErgoTreeVersion rule). Those changes should be moved to corresponding PR.

sc/shared/src/test/scala/sigmastate/lang/SigmaBinderTest.scala
```diff
@@ -31,9 +31,6 @@ class SigmaBinderTest extends AnyPropSpec with ScalaCheckPropertyChecks with Mat
     res
   }
 
-  /** Checks that parsing and binding results in the expected value.
-    * @return the inferred type of the expression
-    */
```
aslesarenko, line 36: Why the scaladoc is removed?

aslesarenko (commented):
LGTM, but please address comments and consider it approved (github doesn't allow approval of own PR)

sc/shared/src/test/scala/sigmastate/lang/SigmaBinderTest.scala
kushti: restored

kushti (approved):

PR #1007, opened 2024-06-10
[6.0] Fix for downcasting BigInt to Long fails

### LINKED ISSUE #877: Downcasting BigInt to Long fails
When downcasting a BigInt to Long it fails with a runtime exception, no matter the value.

Example ergoscript that would trigger the error:
```
{
  sigmaProp(1L.toBigInt.toLong < CONTEXT.preHeader.timestamp)
}
```
Resulting error:
java.lang.RuntimeException: Cannot downcast value CBigInt(1) to the type SLong$

kushti:
Close #877

It has test from the issue as well similar ones for all the Numeric types.

Also, Downcast from BigInt to BigInt and Upcast from BigInt to BigInt are fixed now

### DIFF
--- a/sc/shared/src/test/scala/sigma/LanguageSpecificationV6.scala
+++ b/sc/shared/src/test/scala/sigma/LanguageSpecificationV6.scala
@@ -1,8 +1,8 @@
 package sigma
 
-import sigma.ast.{Apply, Downcast, FixedCost, FixedCostItem, FuncValue, GetVar, Global, JitCost, MethodCall, NamedDesc, OptionGet, SBigInt, SByte, SGlobalMethods, SInt, SLong, SShort, STypeVar, ValUse}
-import sigma.data.{CBigInt, ExactNumeric, RType}
-import sigma.eval.{SigmaDsl, TracedCost}
+import sigma.ast.{Apply, Downcast, FixedCost, FixedCostItem, FuncValue, GetVar, JitCost, OptionGet, SBigInt, SByte, SInt, SLong, SShort, ValUse}
+import sigma.data.{CBigInt, ExactNumeric}
+import sigma.eval.SigmaDsl
 import sigma.util.Extensions.{BooleanOps, ByteOps, IntOps, LongOps}
 import sigmastate.exceptions.MethodNotFound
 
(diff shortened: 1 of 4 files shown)


sc/shared/src/test/scala/sigmastate/TestingInterpreterSpecification.scala
```diff
@@ -215,6 +216,46 @@ class TestingInterpreterSpecification extends CompilerTestingCommons
...
+      test()
+    } else {
+      an[Exception] shouldBe thrownBy(test())
+    }
+  }
+
+  property("upcasting to bigint") {
+    testEval("{ sigmaProp(1L.toBigInt < bigInt(\"2\")) }")
+  }
+
```
aslesarenko, line 258: Every v6.0 change should be tested in LanguageSpecificationV6, this is the only way to ensure:
1) existing behavior is preserved for old versions (including required failures)
2) the new behavior is enabled starting from the given version
2) the implementation is complete, i.e. all the compiler and interpreter components work as expected
In this specific case, changeFeature can be used.

Separate unit tests cannot ensure test coverage.

aslesarenko (changes requested):

sc/shared/src/test/scala/sigma/LanguageSpecificationV6.scala
```diff
@@ -172,21 +172,24 @@ class LanguageSpecificationV6 extends LanguageSpecificationBase { suite =>
...
-    )
+      // The `Downcast(bigInt, SBigInt)` node is never produced by ErgoScript compiler, but is still valid ErgoTree.
+      // Fixed in 6.0
+      assertExceptionThrown(
+        SBigInt.downcast(CBigInt(new BigInteger("0", 16)).asInstanceOf[AnyVal]),
+        _.getMessage.contains("Cannot downcast value")
+      )
+    } else {
+      SBigInt.upcast(CBigInt(new BigInteger("0", 16)).asInstanceOf[AnyVal]) shouldBe CBigInt(new BigInteger("0"))
+      SBigInt.downcast(CBigInt(new BigInteger("0", 16)).asInstanceOf[AnyVal]) shouldBe CBigInt(new BigInteger("0"))
```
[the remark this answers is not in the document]
kushti: done

sc/shared/src/test/scala/sigmastate/TestingInterpreterSpecification.scala
kushti: this test is non-versioned, so nothing to bring to LSV6 from it

For versioned tests, alternatives are added to LSV6.

kushti:
@aslesarenko comments addressed , please make another pass

sc/shared/src/test/scala/sigma/LanguageSpecificationV6.scala
```diff
@@ -172,21 +172,26 @@ class LanguageSpecificationV6 extends LanguageSpecificationBase { suite =>
...
-    assertExceptionThrown(
-      SBigInt.upcast(CBigInt(new BigInteger("0", 16)).asInstanceOf[AnyVal]),
-      _.getMessage.contains("Cannot upcast value")
-    )
+    if (activatedVersionInTests < VersionContext.V6SoftForkVersion) {
+      // The `Upcast(bigInt, SBigInt)` node is never produced by ErgoScript compiler, but is still valid ErgoTree.
+      // Fixed in 6.0
+      assertExceptionThrown(
+        SBigInt.upcast(CBigInt(new BigInteger("0", 16)).asInstanceOf[AnyVal]),
+        _.getMessage.contains("Cannot upcast value")
```
aslesarenko, line 180: good enough, but FYI there is also method `exceptionLike[ConcreteExectionType]("Cannot upcast value")` which tests both exception type and the contents of the message.

aslesarenko (approved):
LGTM, but please see the additional comment.

sc/shared/src/test/scala/sigma/LanguageSpecificationV6.scala
kushti: The check is corresponding to previously written code, so such refactoring is a subject of a dedicated issue & PR I suppose

PR #1008, opened 2024-06-11
[6.0] Lazy default for Option.getOrElse and Coll.getOrElse

kushti:
In this PR , lazy default evaluation for Option.getOrElse and Coll.getOrElse is implemented


sc/shared/src/test/scala/sigma/LanguageSpecificationV5.scala
```diff
@@ -7957,77 +7957,81 @@ class LanguageSpecificationV5 extends LanguageSpecificationBase { suite =>
...
-            BlockValue(
-              Vector(
-                ValDef(
-                  3,
-                  List(),
-                  SelectField.typed[Value[STuple]](
-                    ValUse(1, SPair(SCollectionType(SInt), SPair(SInt, SInt))),
-                    2.toByte
+
+    if(!VersionContext.current.isV6SoftForkActivated) {
```
aslesarenko, line 7961: Protocol change like this can be tested using changeFeature (see other tests) in combination with newVersionedResults (also see code). In this case both old and new behaviour will be tested.

In principle this test case can be moved to LSV6.

sc/shared/src/test/scala/sigma/LanguageSpecificationV5.scala
```diff
@@ -8591,13 +8595,15 @@ class LanguageSpecificationV5 extends LanguageSpecificationBase { suite =>
...
         FuncValue(Vector((1, SOption(SLong))), OptionIsDefined(ValUse(1, SOption(SLong))))))
 
-    verifyCases(
-      Seq(
-        (None -> Expected(Success(1L), 1766, costDetails3, 1766)),
-        (Some(10L) -> Expected(Success(10L), 1766, costDetails3, 1766))),
-      existingFeature({ (x: Option[Long]) => x.getOrElse(1L) },
-        "{ (x: Option[Long]) => x.getOrElse(1L) }",
-        FuncValue(Vector((1, SOption(SLong))), OptionGetOrElse(ValUse(1, SOption(SLong)), LongConstant(1L)))))
+    if (!VersionContext.current.isV6SoftForkActivated) {
```
aslesarenko, line 8598: Same as above, this `if` is not necessary when changeFeature + newVersionedResults is used.

sc/shared/src/test/scala/sigma/LanguageSpecificationV6.scala
```diff
@@ -168,7 +169,62 @@ class LanguageSpecificationV6 extends LanguageSpecificationBase { suite =>
...
+        OptionGetOrElse(
+          ValUse(1, SOption(SLong)),
+          ArithOp(LongConstant(1L), LongConstant(0L), OpCode @@ (-99.toByte))
+        )
+      )
+    )
 
+    if (VersionContext.current.isV6SoftForkActivated) {
+      forAll { x: Option[Long] =>
+        Seq(getOrElse).map(_.checkEquality(x))
```
aslesarenko, line 189: Check equality is not enough, as it doesn't involve all interpreter components.

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
```diff
@@ -157,6 +158,43 @@ class BasicOpsSpecification extends CompilerTestingCommons
...
+        |  getVar[Int](1).getOrElse(getVar[Int](44).get) > 0
+        |}
+        |""".stripMargin,
+      null
+    )
+
+    if(VersionContext.current.isV6SoftForkActivated) {
+      optTest()
+    } else {
+      an[Exception] shouldBe thrownBy(optTest())
```
aslesarenko, line 176: It is better to use assertExceptionThrown.

aslesarenko (changes requested):

sc/shared/src/test/scala/sigma/LanguageSpecificationV5.scala
kushti: changedFeature is not enough here, you need for versioned verifyCases, as in 6.0 costing trace would be different in some cases due to difference in default evaluation (at the same time, it wouldnt provide higher cost, so the change is secure costing-wise). 

Seems like *Feature machinery is overly complex, but still not enough to describe protocol changes, thus I would recommend to decide what it is actually testing (or should test), and then follow simpler approaches

sc/shared/src/test/scala/sigma/LanguageSpecificationV5.scala
kushti: Same as above

sc/shared/src/test/scala/sigma/LanguageSpecificationV6.scala
kushti: Please specify what is needed

kushti:
@aslesarenko comments addressed, please make another pass

sc/shared/src/test/scala/sigma/LanguageSpecificationV5.scala
aslesarenko: Please see other examples where changedFeature is used, there are test cases where cost trace depend on the version and you can pass different traces to the constructor of Expected.
The perceived complexity is due to learning curve.

sc/shared/src/test/scala/sigma/LanguageSpecificationV6.scala
aslesarenko: verifyCases will exercise all the components.

sc/shared/src/test/scala/sigma/LanguageSpecificationV5.scala
kushti: changedFeature is tied to V4/V5 switch, see checkExpected / checkVerify functions in SigmaDslTesting. So it cant be used in V6 testing without modifications it seems. So it is definitely not about learning curve, changedFeature is just not ready for V6

kushti: @aslesarenko see https://github.com/ScorexFoundation/sigmastate-interpreter/issues/1024 for example

kushti:
@aslesarenko after all I've found some V6 related modifications in #995 , tweaked them further and made verifyCases tests. Please make another pass

sc/shared/src/test/scala/sigma/LanguageSpecificationV5.scala
kushti: done

aslesarenko (approved):

PR #1013, opened 2024-06-17
[6.0] Global.fromBigEndianBytes implementation

### LINKED ISSUE #993: Global.fromBigEndianBytes method
A new Global.fromBigEndianBytes[TNum] method should be implemented which can parse big-endian two's complement binary representation of TNum type.

Examples:
```scala
val shortBytes = Coll(0.toByte, 16.toBytes) 
val s = fromBigEndianBytes[Short](shortBytes) // using global pre-defined function
// assert (s == 16.toShort)

val intBytes = Coll(0.toByte, 0.toByte) ++ shortBytes
val i = Global.fromBigEndianBytes[Int](intBytes) // using method of Global
// assert (i == 16)
```
It should be mutually inverse to Numeric.toBigEndianBytes, so that, continuing the above example:
```scala
val isEqual = i.toBigEndianBytes == intBytes
// assert (isEqual)
```

kushti:
A new Global.fromBigEndianBytes[TNum] method is implemented in this PR, which can parse big-endian two's complement binary representation of TNum type.

example:

```
{
  val l = 1088800L
  val ba = l.toBytes
  Global.fromBigEndianBytes[Long](ba) == l
}
```

close #993


data/shared/src/main/scala/sigma/data/CSigmaDslBuilder.scala
```diff
@@ -201,6 +204,40 @@ class CSigmaDslBuilder extends SigmaDslBuilder { dsl =>
...
+      case sigma.LongType => if (bytes.length != 8) {
+        throw new IllegalArgumentException("To deserialize SLong with fromBigEndianBytes, exactly eight bytes should be provided")
+      } else {
+        Longs.fromByteArray(bytes.toArray).asInstanceOf[T]
+      }
+      case sigma.BigIntRType =>
+        if (bytes.length > SBigInt.MaxSizeInBytes) {
+          throw SerializerException(s"BigInt value doesn't not fit into ${SBigInt.MaxSizeInBytes} bytes in fromBigEndianBytes")
+        }
+        CBigInt(new BigInteger(bytes.toArray)).asInstanceOf[T]
```
aslesarenko, line 235: This should also satisfy the bit size for BigInt values, to catch the potential problem early.
See multiplication operation.

sc/shared/src/test/scala/sigmastate/ErgoTreeSpecification.scala
```diff
@@ -516,7 +516,7 @@ class ErgoTreeSpecification extends SigmaDslTesting with ContractsTestkit with C
       (SGlobal.typeId, Seq(
         MInfo(1, groupGeneratorMethod), MInfo(2, xorMethod)
       ) ++ (if (isV6Activated) {
-        Seq(MInfo(3, serializeMethod)) // methods added in v6.0
+        Seq(MInfo(3, serializeMethod), MInfo(5, fromBigEndianBytesMethod)) // methods added in v6.0
```
aslesarenko, line 519: worth making a note about id = 4.

aslesarenko (approved):
LGTM, but please see comments.

data/shared/src/main/scala/sigma/data/CSigmaDslBuilder.scala
kushti: done , 

```
        CBigInt(new BigInteger(bytes.toArray).to256BitValueExact).asInstanceOf[T]
```

now

sc/shared/src/test/scala/sigmastate/ErgoTreeSpecification.scala
kushti: done

PR #1014, opened 2024-06-19
[6.0] GetVar(inputIndex, varId) for reading context variable from another input

### LINKED ISSUE #983: GetVar(inputIndex, varId) variant for reading context variable from another input
In addition to current GetVar(varId) to get input context variable, it is worth to have GetVar(inputIndex, varId) variant for reading context variable from another input. This could be useful for sharing proofs and state ids across inputs. 

I think new GetVar variant should be implemented as a method call (only), avoiding introducing new opcodes (so new `Context.getInputVar[T](inputIndex: Short, varId: Byte)` method.

kushti:
close #983


kushti:
moved to #1016

PR #1015, opened 2024-06-28
[6.0] Box.getReg implementation

### LINKED ISSUE #416: Implementation of Box.getReg
Related to #481 
Implement Box.getReg. Currently ignored tests related to getReg:
"Box.getReg"
"Box.getReg equivalence"

Also search TODOs referring to this issue

kushti:
In this PR, Box.getReg to get register contents by index is implemented

Example: 

```
{
   val x = SELF
   x.getReg[Long](0).get == SELF.value &&
   x.getReg[Coll[(Coll[Byte], Long)]](2).get == SELF.tokens &&
   x.getReg[Int](9).isEmpty
}
```

close #416


sc/shared/src/main/scala/sigma/compiler/ir/GraphBuilding.scala
```diff
@@ -1017,6 +1018,10 @@ trait GraphBuilding extends Base with DefRewriting { IR: IRContext =>
           case (box: Ref[Box]@unchecked, SBoxMethods) => method.name match {
             case SBoxMethods.tokensMethod.name =>
               box.tokens
+            case SBoxMethods.getRegMethodV6.name if VersionContext.current.isV6SoftForkActivated =>
+              val c1 = asRep[Int](argsV(0))
+              val c3 = stypeToElem(typeSubst.apply(tT))
+              box.getReg(c1)(c3)
```
aslesarenko, line 1024: Why such names c1 and c3?

sc/shared/src/main/scala/sigma/compiler/phases/SigmaTyper.scala
```diff
@@ -136,7 +136,7 @@ class SigmaTyper(val builder: SigmaBuilder,
     case Apply(ApplyTypes(sel @ Select(obj, n, _), Seq(rangeTpe)), args) =>
       val newObj = assignType(env, obj)
       val newArgs = args.map(assignType(env, _))
-      obj.tpe match {
+      newObj.tpe match {
```
aslesarenko, line 139: Good catch!

sc/shared/src/test/scala/sigmastate/ErgoTreeSpecification.scala
```diff
@@ -367,7 +367,7 @@ class ErgoTreeSpecification extends SigmaDslTesting with ContractsTestkit {
         MInfo(4, BytesWithoutRefMethod),
         MInfo(5, IdMethod),
         MInfo(6, creationInfoMethod),
-        MInfo(7, getRegMethod),
+        MInfo(7, getRegMethodV5),
```
aslesarenko, line 370: It is better to version this test as well.

sc/shared/src/test/scala/sigma/LanguageSpecificationV6.scala
```diff
@@ -265,9 +269,13 @@ class LanguageSpecificationV6 extends LanguageSpecificationBase { suite =>
...
-    // TODO v6.0: related to https://github.com/ScorexFoundation/sigmastate-interpreter/issues/416
-    val getReg = newFeature((x: Box) => x.getReg[Int](1).get,
-      "{ (x: Box) => x.getReg[Int](1).get }",
+    // related to https://github.com/ScorexFoundation/sigmastate-interpreter/issues/416
+    def getReg = newFeature((x: Box) => x.getReg[Long](0).get,
+      "{ (x: Box) => x.getReg[Long](0).get }",
+      FuncValue(
+        Array((1, SBox)),
+        OptionGet(ExtractRegisterAs(ValUse(1, SBox), ErgoBox.R0, SOption(SLong)))
+      ),
```
aslesarenko, line 278: The [idea behind getReg](https://github.com/ScorexFoundation/sigmastate-interpreter/issues/416#issuecomment-2063781813) was to have "computable" index, not just constant.
For example the index can come from Context variable and thus the concrete register will not be fixed in the ErgoTree. This will allow code like:
```val x = box.getReg[Int](getVar[Int](varId).get).get```

So I suggest to add additional test with variable register passed as function argument and have a series of cases in verifyCases.
```"{ (x: (Box, Int)) => x._1.getReg[Long](x._2).get }"```

File: sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala [hunk 5]
```diff
@@ -765,4 +767,25 @@ class BasicOpsSpecification extends CompilerTestingCommons
...
+      )
+    }
+
+    if (VersionContext.current.isV6SoftForkActivated) {
+      getRegTest()
+    } else {
+      an[Exception] should be thrownBy getRegTest()
+    }
+
+  }
```
aslesarenko, line 789: Having test coverage in LSV6 this test doesn't bring much value.
The negative case also better test in LSV6 (there are examples in LSV5 how to do this)

aslesarenko (changes requested):

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
kushti: I disagree, with the help of a test in BasicOpsSpecification compiler bug in https://github.com/ScorexFoundation/sigmastate-interpreter/pull/984 was found and fixed , and LSV6 did not help with that

aslesarenko: I'm not against unit test, but BasicOpsSpecification is not "unit"-test-like suite, it tests both the compiler and the interpreter, so basically serving the purpose of LSVx. 
I'm also ok with leaving this test here, but it looks like duplication to me.

sc/shared/src/test/scala/sigmastate/ErgoTreeSpecification.scala
kushti: done

sc/shared/src/main/scala/sigma/compiler/ir/GraphBuilding.scala
kushti: changed to c1 & c2

sc/shared/src/test/scala/sigma/LanguageSpecificationV6.scala
kushti: `{ (x: (Box, Int)) => x._1.getReg[Long](x._2).get }` is way different from `box.getReg[Int](getVar[Int](varId).get).get` , 
so I've made a test in `BasicOpsSpecification` which is checking exactly context variable usage:

```
{
   val x = SELF.getReg[Long](getVar[Int](1).get).get
   x == SELF.value
}
```
and so the new test can also be used for reference for people asked for computable index.

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
kushti: Yes, BOS is checking serialization / interpreter / compiler, but approach is different (more complex programs, in particular), thus it is possible to catch bugs with it where LSV is missing them (see example above) , and also works for providing useful examples of new features usage (see https://github.com/ergoplatform/ergo/releases/tag/v6.0.0-alpha0 , which is suggesting to find examples in this suite).

kushti:

aslesarenko (approved):

PR #1019, opened 2024-07-08
Revise all error messages adopting usability pattern 

Emmanueleclipse:
**### Changes Made:**

- Switched to clearer explanations: Updated error messages to explicitly state what went wrong and why.
- Used empathy and politeness: Maintained a polite and empathetic tone.
- Added actionable guidance: Included specific instructions on how users can resolve issues or contact support.
- Reduced technical jargon and blame: Reduced technical language and avoided blaming users or third parties.
Note: Currently using “#” in place for a link to contact customer support

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,7 +22,9 @@ lazy val commonSettings = Seq(
         Seq("-Ywarn-unused:_,imports", "-Ywarn-unused:imports", "-release", "8")
       case Some((2, 11)) =>
         Seq()
-      case _ => sys.error("Unsupported scala version")
+      case _ => sys.error("The Scala version you are using is not supported. " +
+        "Please use one of the supported versions: Scala 2.11, 2.12, or 2.13. " +
+        "If the issue keeps happening, contact <a href=\"#\">Customer care</a>.")
     }
   },
   javacOptions ++= javacReleaseOption,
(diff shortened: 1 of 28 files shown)


build.sbt
```diff
@@ -22,7 +22,9 @@ lazy val commonSettings = Seq(
         Seq("-Ywarn-unused:_,imports", "-Ywarn-unused:imports", "-release", "8")
       case Some((2, 11)) =>
         Seq()
-      case _ => sys.error("Unsupported scala version")
+      case _ => sys.error("The Scala version you are using is not supported. " +
+        "Please use one of the supported versions: Scala 2.11, 2.12, or 2.13. " +
+        "If the issue keeps happening, contact <a href=\"#\">Customer care</a>.")
```
aslesarenko, line 27: 1) for the "Help them fix it" part, an instruction of which SBT variable should be checked for the current version.
2) Also there is no "Customer care", only Discord channels. (this also applies to all the other messages)

core/js/src/main/scala/sigma/crypto/Platform.scala
```diff
@@ -246,7 +246,9 @@ object Platform {
     case c: Coll[_] => tpe match {
       case STuple(items) => c.tItem == sigma.AnyType && c.length == items.length
       case tpeColl: SCollection[_] => true
-      case _ => sys.error(s"Collection value $c has unexpected type $tpe")
+      case _ => sys.error(s"The value in the collection $c is not the correct type: $tpe. " +
+        "Please make sure the value is of the correct type. " +
+        "If the issue keeps happening, contact <a href=\"#\">Customer care</a>.")
```
aslesarenko, line 251: The meaning of the message has been changed, which makes it less clear (despite many more words).
Please see how the method is used to come up with a clarifying message.

aslesarenko (changes requested):

satsen:
this should be closed

PR #1020, opened 2024-07-16
[6.0] Serialize SFunc in TypeSerializer

### LINKED ISSUE #847: Serialize SFunc in TypeSerializer
When function type is part of lambda argument like in the following ErgoScript code
```scala
def apply(in: (Int => Int, Int)) = in._1(in._2)
```
then TypeSerializer throws an exception, because case for SFunc is not implemented. 

This test case should work instead of throwing an exception https://github.com/ScorexFoundation/sigmastate-interpreter/blob/b754e143cf38ed86d95698ede744a470dfa053d6/sigmastate/src/test/scala/special/sigma/SigmaDslSpecification.scala#L10040

kushti:
This PR is adding higher-order lambdas support in serialization, thus making them useful.

Example:
```
{
   val c = Coll(Coll(1))
   def fn(xs: Coll[Int]) = {
     val inc = { (x: Int) => x + 1 }
     def apply(in: (Int => Int, Int)) = in._1(in._2)
     val ys = xs.map { (x: Int) => apply((inc, x)) }
     ys.size == xs.size && ys != xs
   }
   c.exists(fn)
 }
```
Close #847

### DIFF
--- a/core/shared/src/main/scala/sigma/serialization/TypeSerializer.scala
+++ b/core/shared/src/main/scala/sigma/serialization/TypeSerializer.scala
@@ -1,6 +1,7 @@
 package sigma.serialization
 
 import debox.cfor
+import sigma.VersionContext
 import sigma.ast.SCollectionType.{CollectionTypeCode, NestedCollectionTypeCode}
 import sigma.ast._
 import sigma.serialization.{CoreByteReader, CoreByteWriter, InvalidTypePrefix}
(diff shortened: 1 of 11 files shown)


core/shared/src/main/scala/sigma/serialization/TypeSerializer.scala
```diff
@@ -189,7 +201,23 @@ class TypeSerializer {
...
+          val tpeParamsLength = r.getUByte()
+          val tpeParams = (1 to tpeParamsLength).map { _ =>
+            val ident = deserialize(r)
+            require(ident.isInstanceOf[STypeVar])
+            STypeParam(ident.asInstanceOf[STypeVar])
+          }
+          SFunc(tDom, tRange, tpeParams)
+        // todo: serialize tParams
         case _ =>
+          // todo: 6.0: replace 1008 check with identical but other opcode
```
aslesarenko, line 220: please clarify "identical but other".

aslesarenko (approved):
LGTM, but also see the comment

core/shared/src/main/scala/sigma/serialization/TypeSerializer.scala
kushti: done, // todo: 6.0: replace 1008 check with identical behavior but other opcode, to activate
          //  ReplacedRule(1008 -> new opcode) during 6.0 activation

PR #1036, opened 2024-11-27
Candidate for 5.0.15 release

### LINKED ISSUE #1030: Uses library outside Maven central
The scrypto artifact used in this library is from a snapshot repository

kushti:
* scrypto version updated to 3.0.0 (close #1030 )
* #976

### DIFF
--- a/sc/shared/src/main/scala/sigma/compiler/ir/Base.scala
+++ b/sc/shared/src/main/scala/sigma/compiler/ir/Base.scala
@@ -2,8 +2,9 @@ package sigma.compiler.ir
 
 import debox.{cfor, Buffer => DBuffer}
 import sigma.compiler.ir.core.MutableLazy
-import sigma.data.OverloadHack.Overloaded1
+import sigma.ast.{DeserializeContext, SType}
 import sigma.data.{AVHashMap, Nullable, RType}
+import sigma.data.OverloadHack.Overloaded1
 import sigma.reflection.RConstructor
 import sigma.util.StringUtil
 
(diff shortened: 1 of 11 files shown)


sc/shared/src/main/scala/sigma/compiler/ir/Base.scala
```diff
@@ -203,10 +204,11 @@ abstract class Base { thisIR: IRContext =>
...
 
-  /** Logical AND between two pattern matches of the save value `x`.
-    * Can be used to construct patterns like `case P1 && P2 => ...` */
-  object && {
-    def unapply[T](x: T): Option[(T,T)] = Some((x, x))
+  /**
+    * Def done in order to carry on DeserializeContext through stages of compilation intact
+    */
+  case class DeserializeContextDef[V <: SType](d: DeserializeContext[V], e: Elem[V]) extends Def[V] {
+    override def resultType: Elem[V] = e
```
aslesarenko, line 211: There is a problem with typing here. In a type like `Elem[V]`, the V should be a type of data value (Int, Long, Boolean, Coll, GroupElement). However, the `V <: SType` part suggests `V` to be SInt, SLong, SBoolean, etc.
The compiler is not complaining likely because of `.asInstanceOf[]` being used.
It should be `..., e: Elem[V#WrappedType]) extends Def[V#WrappedType] ...`

aslesarenko (changes requested):
Looks good, except one place, see the comment.

sc/shared/src/main/scala/sigma/compiler/ir/Base.scala
kushti: Fixed

kushti:

SethDusek (approved):

PR #1038, opened 2024-12-02
[6.0] Fix semantics of AvlTree.insert & new AvlTree.insertOrUpdate method 

### LINKED ISSUE #908: [v6.0] Fix semantics of AvlTree.insert
Current (v5.0) semantics is not consistent with AvlTree.update.
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/3da19f01798bfa17620dcb5ad8ed327f9217f755/interpreter/shared/src/main/scala/sigmastate/types.scala#L2328

kushti:
close #908 

In this PR, semantics of AvlTree.insert fixed, and new AvlTree.insertOrUpdate method added


data/shared/src/main/scala/sigma/eval/AvlTreeVerifier.scala
```diff
@@ -50,6 +50,18 @@ trait AvlTreeVerifier {
...
 
+  /**
+    * Returns Failure if the proof does not verify.
+    * Otherwise, successfully modifies tree and so returns Success.
+    * After one failure, all subsequent operations with this verifier will fail and digest
+    * is None.
+    *
+    * @param key   key to look up
+    * @param value value to check it was inserted or updated
+    * @return Success(Some(value)), Success(None), or Failure
```
aslesarenko, line 61: Make sense to describe each case of returned result.

interpreter/shared/src/main/scala/sigmastate/eval/Extensions.scala
```diff
@@ -120,6 +120,24 @@ object Extensions {
       }
     }
 
+    def insertOrUpdate(
+                entries: Coll[(Coll[Byte], Coll[Byte])],
+                proof: Coll[Byte]): Option[AvlTree] = {
+      if (!tree.isInsertAllowed || !tree.isUpdateAllowed) {
```
aslesarenko, line 126: Both flags are required, however for each given tree it will either be insert or update (but not both).
Is it possible to make it more flexible and require the flag only if the corresponding operation is happening for a given tree.

interpreter/shared/src/main/scala/sigmastate/interpreter/CErgoTreeEvaluator.scala
```diff
@@ -146,7 +146,7 @@ class CErgoTreeEvaluator(
           val insertRes = bv.performInsert(key.toArray, value.toArray)
           // TODO v6.0: throwing exception is not consistent with update semantics
           //  however it preserves v4.0 semantics (see https://github.com/ScorexFoundation/sigmastate-interpreter/issues/908)
```
aslesarenko, line 148: todo can be removed.

interpreter/shared/src/main/scala/sigmastate/interpreter/CErgoTreeEvaluator.scala
```diff
@@ -192,6 +192,37 @@ class CErgoTreeEvaluator(
...
+    } else {
+      val bv     = createVerifier(tree, proof)
+      // when the tree is empty we still need to add the insert cost
+      val nItems = Math.max(bv.treeHeight, 1)
+
+      // here we use forall as looping with fast break on first failed tree operation
+      operations.forall { case (key, value) =>
+        var res = true
+        // the cost of tree update is O(bv.treeHeight)
+        addSeqCost(UpdateAvlTree_Info, nItems) { () =>
```
aslesarenko, line 211:
```suggestion
      // Here (and in the previous methods) the cost is not properly approximated. 
      // When the tree is small (or empty), but there are many `operations`, the treeHeight will grow on every iteration. 
      // So should the cost on every iteration.
      addSeqCost(UpdateAvlTree_Info, nItems) { () =>
```

aslesarenko (approved):
LGTM, but see comments.

interpreter/shared/src/main/scala/sigmastate/eval/Extensions.scala
kushti: " for each given tree it will either be insert or update (but not both)" - no, it could be insert for one key and update for another

interpreter/shared/src/main/scala/sigmastate/interpreter/CErgoTreeEvaluator.scala
kushti: done

data/shared/src/main/scala/sigma/eval/AvlTreeVerifier.scala
kushti: done

PR #1043, opened 2024-12-11
[6.0] ErgoTree version based versioning, polished activation details and revised checkSoftForkCondition

### LINKED ISSUE #904: [v6.0] Revise checkSoftForkCondition
See the corresponding comment 
https://github.com/ScorexFoundation/sigmastate-interpreter/blob/49d1255978d3d0edad2ddf7bfe31e481937b94c0/interpreter/shared/src/main/scala/sigmastate/interpreter/Interpreter.scala#L285

### LINKED ISSUE #1042: Lazyness in getOrElse could cause disagreement between 6.0 and < 6.0 nodes
Something like `Some(5).getOrElse(emptyColl(0))` would cause network split. V6 nodes would accept this script because the invalid operation (indexing empty coll) is never evaluated whereas v5 nodes would reject. Unless ErgoTree version is also >= 3 in which case old nodes would not evaluate script

Discovered by @SethDusek

kushti:
In this section activation of 6.0 features is made dependent on ErgoTree version, as this is what versioning mechanism set in previous release is expecting

close #1042
also close #904


data/js/src/main/scala/sigma/Platform.scala
```diff
@@ -14,33 +14,33 @@ object Platform {
   private[sigma] def liftToConstant(obj: Any, builder: SigmaBuilder): Nullable[Constant[SType]] = {
```
sethdusek, line 14: Needs case for Option[T]? Based on my understanding you can't do substituteconstants on an Option constant if it's not supported here

data/js/src/main/scala/sigma/Platform.scala
```diff
@@ -14,33 +14,33 @@ object Platform {
...
       case n: sigma.BigInt => Nullable(mkConstant[SBigInt.type](n, SBigInt))
       case n: sigma.UnsignedBigInt => Nullable(mkConstant[SUnsignedBigInt.type](n, SUnsignedBigInt))
       case ge: GroupElement => Nullable(mkConstant[SGroupElement.type](ge, SGroupElement))
       case b: Boolean => Nullable(if (b) TrueLeaf else FalseLeaf)
-      case v: String if !VersionContext.current.isV6SoftForkActivated => Nullable(mkConstant[SString.type](v, SString))
-      case h: Header if VersionContext.current.isV6SoftForkActivated  => Nullable(mkConstant[SHeader.type](h, SHeader))
-      case h: PreHeader if VersionContext.current.isV6SoftForkActivated  => Nullable(mkConstant[SPreHeader.type](h, SPreHeader))
+      case v: String if !VersionContext.current.isV3OrLaterErgoTreeVersion => Nullable(mkConstant[SString.type](v, SString))
+      case h: Header if VersionContext.current.isV3OrLaterErgoTreeVersion  => Nullable(mkConstant[SHeader.type](h, SHeader))
+      case h: PreHeader if VersionContext.current.isV3OrLaterErgoTreeVersion  => Nullable(mkConstant[SPreHeader.type](h, SPreHeader))
```
sethdusek, line 36: Is there a use for this yet?

data/js/src/main/scala/sigma/Platform.scala
kushti: Seems to be the case indeed, will add

kushti: Seems better to postpone to 7.0 to minimize codebase changes at this point, made an issue https://github.com/ergoplatform/sigmastate-interpreter/issues/1065

data/js/src/main/scala/sigma/Platform.scala
kushti: Preheader? added a test in https://github.com/ergoplatform/sigmastate-interpreter/pull/930/commits/d9b4e49577712f4863961c4a7cf954565ee87bdc 

I guess could be used already in some apps.
Not sure about substConstants

sethdusek: I don't think substConstants can handle PreHeader since it can't be serialized in DataSerializer

SethDusek (approved):

PR #1045, opened 2025-01-06
Fix for Global.decodeNBits can produce bigint out of 256 bits scope

### LINKED ISSUE #1044: Global.decodeNBits method can produce bigint out of 256 bits scope

kushti:
In this PR, result of Global.decodeNBits is checked to be fit in 256 bits, to avoid spam attacks
close #1044


sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
```diff
@@ -3248,4 +3248,23 @@ class BasicOpsSpecification extends CompilerTestingCommons
...
+    }
+
+    if (VersionContext.current.isV6SoftForkActivated) {
+      // on JVM, InvocationTargetException wrapping (ArithmeticException: BigInteger out of 256 bit range) is thrown
+      an[Exception] should be thrownBy someTest()
+    } else {
+      an[sigma.validation.ValidationException] should be thrownBy someTest()
+    }
+  }
+
```
aslesarenko, line 3269: Would be good to have test cases for exact boundaries for the max value N: 
1) for value N it passes 
2) for N+1 it fails

aslesarenko (approved):
LGTM, but please see comment

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
kushti: done something similar

PR #1047, opened 2025-01-08
Prohibition for using 6.0 types in register and context extension vars

kushti:


data/shared/src/main/scala/org/ergoplatform/validation/ValidationRules.scala
```diff
@@ -155,6 +155,39 @@ object ValidationRules {
...
+          throwValidationException(
+            new SerializerException(s"V6 type used in register or context var extension: $tpe"),
+            Array[Any](tpe))
+        }
+      }
+      v match {
+        case c: Constant[_] => v6TypeCheck(c.tpe)
+        case c: EvaluatedCollection[_, _] => v6TypeCheck(c.elementType)
+        case GroupGenerator =>
+      }
```
aslesarenko, line 177: What is the motivation to add and check this rule?

data/shared/src/main/scala/org/ergoplatform/validation/ValidationRules.scala
```diff
@@ -155,6 +155,39 @@ object ValidationRules {
...
+
+      def v6TypeCheck(tpe: SType) = {
+        if (tpe.isOption || tpe.typeCode == SHeader.typeCode || tpe.typeCode == SUnsignedBigInt.typeCode) {
+          throwValidationException(
+            new SerializerException(s"V6 type used in register or context var extension: $tpe"),
+            Array[Any](tpe))
+        }
+      }
+      v match {
+        case c: Constant[_] => v6TypeCheck(c.tpe)
```
SethDusek, line 174: It seems it is possible to serialize/parse Tuple with v6 type

```scala
  property("impossible to use v6 types in box registers") {
    val trueProp = ErgoTreePredef.TrueProp(ErgoTree.defaultHeaderWithVersion(3))

    val b = new ErgoBoxCandidate(1L, trueProp, 1,
                  additionalRegisters = Map(R4 -> Tuple(UnsignedBigIntConstant(new BigInteger("1")), IntConstant(1))))
    VersionContext.withVersions(3, 3) {
      val bs = ErgoBoxCandidate.serializer.toBytes(b)
      a[sigma.validation.ValidationException] should be thrownBy ErgoBoxCandidate.serializer.fromBytes(bs)
    }
}
```

doesn't throw exception

kushti: @SethDusek fixed, the reason was c.elementType being SAny for Tuple. Made test for UBI as second tuple element, and collection as well

data/shared/src/main/scala/org/ergoplatform/validation/ValidationRules.scala
```diff
@@ -162,6 +162,44 @@ object ValidationRules {
...
```
sethdusek: This seems to be insufficient in its current form, for example a Constant with type STuple(.., UnsignedBigInt | Option | Header) doesn't get caught by this. Example of type (Byte, Short, Byte, UnsignedBigInt) (Constant was built using sigma-rust so passed as base16 bytes here).

```scala
    val reader = new SigmaByteReader(new VLQByteBufferReader(ByteBuffer.wrap(decodeBytes("5402030209050a050105").toArray)), new ConstantStore(), false)

    val v = VersionContext.withVersions(3, 3) {ConstantSerializer(DeserializationSigmaBuilder).parse(reader).asInstanceOf[Constant[STuple]] }
    val b2 = new ErgoBoxCandidate(1L, trueProp, 1,
      additionalRegisters = Map(R4 -> v))
    VersionContext.withVersions(3, 3) {
      val bs2 = ErgoBoxCandidate.serializer.toBytes(b2)
      a[sigma.validation.ValidationException] should be thrownBy ErgoBoxCandidate.serializer.fromBytes(bs2)
    }
```

I think the checks should be recursive

data/shared/src/main/scala/org/ergoplatform/validation/ValidationRules.scala
```diff
@@ -162,6 +162,44 @@ object ValidationRules {
...
+        case t: Tuple => if(t.items.length != 2) {
```
sethdusek: It's possible to parse tuples with length > 2 although they don't seem to be very useful since they can't be evaluated, but they could be stuffed in register/contextextension

SethDusek (changes requested):

data/shared/src/main/scala/org/ergoplatform/validation/ValidationRules.scala
kushti: Indeed, recursive checks done

data/shared/src/main/scala/org/ergoplatform/validation/ValidationRules.scala
kushti: addressed, no length checked now

data/shared/src/main/scala/org/ergoplatform/validation/ValidationRules.scala
```diff
@@ -162,6 +164,47 @@ object ValidationRules {
...
+
+      def v6TypeCheck(tpe: SType) = {
+        if (tpe.isOption || tpe.typeCode == SHeader.typeCode || tpe.typeCode == SUnsignedBigInt.typeCode) {
+          throwValidationException(
+            SerializerException(s"V6 type used in register or context var extension: $tpe"),
+            Array[Any](tpe))
+        }
+      }
+
+      def step(s: SType): Unit = {
```
sethdusek, line 182: Doesn't seem to be sufficient for Coll type that's nested in another type like Tuple or Coll

```scala
    val b7 = new ErgoBoxCandidate(1L, trueProp, 1,
      additionalRegisters = Map(R4 -> ConcreteCollection(Seq(ConcreteCollection(Seq(UnsignedBigIntConstant(new BigInteger("1"))), SUnsignedBigInt)), (SCollection(SUnsignedBigInt)))))
    VersionContext.withVersions(3, 3) {
      val bs = ErgoBoxCandidate.serializer.toBytes(b7)
      a[sigma.validation.ValidationException] should be thrownBy ErgoBoxCandidate.serializer.fromBytes(bs)
    }
```

kushti: oh right, fixed in https://github.com/ergoplatform/sigmastate-interpreter/pull/1047/commits/845c3e02023e2c02a0d10f4b05c2b4a478690caa

SethDusek (approved):

PR #1048, opened 2025-01-09
Fix for Box.getReg v6 - different ids used now

kushti:


data/shared/src/main/scala/sigma/ast/methods.scala
```diff
@@ -1427,7 +1427,7 @@ case object SBoxMethods extends MonoTypeMethods {
         ArgInfo("regId", "zero-based identifier of the register."))
 
   lazy val getRegMethodV6 = SMethod(this, "getReg",
-    SFunc(Array(SBox, SInt), SOption(tT), Array(paramT)), 7, ExtractRegisterAs.costKind, Seq(tT))
+    SFunc(Array(SBox, SInt), SOption(tT), Array(paramT)), 19, ExtractRegisterAs.costKind, Seq(tT))
```
aslesarenko, line 1430: Why 19 and not 9?

kushti: > Why 19 and not 9?

It is occupied by R0, see ErgoTreeSpecification for method ids checks :

```
{ import SBoxMethods._
      (SBox.typeId,  Seq(
        MInfo(1, ValueMethod),
        MInfo(2, PropositionBytesMethod),
        MInfo(3, BytesMethod),
        MInfo(4, BytesWithoutRefMethod),
        MInfo(5, IdMethod),
        MInfo(6, creationInfoMethod),
        MInfo(8, tokensMethod)
      ) ++ (if (isV6Activated) {
        Seq(MInfo(19, getRegMethodV6))
      } else {
        Seq(MInfo(7, getRegMethodV5))
      }) ++ registers(idOfs = 8)
        .zipWithIndex
        .map { case (m,i) => MInfo((8 + i + 1).toByte, m) }, true)
    }
```

SethDusek (approved):

PR #1049, opened 2025-01-10
UnsignedBigInt result type for Global.powHit

kushti:
In this PR, Global.powHit result type set to UnsignedBigInt

### DIFF
--- a/data/shared/src/main/scala/sigma/ast/methods.scala
+++ b/data/shared/src/main/scala/sigma/ast/methods.scala
@@ -16,8 +16,7 @@ import sigma.data.ExactIntegral.{ByteIsExactIntegral, IntIsExactIntegral, LongIs
 import sigma.data.NumericOps.BigIntIsExactIntegral
 import sigma.data.OverloadHack.Overloaded1
 import sigma.data.UnsignedBigIntNumericOps.UnsignedBigIntIsExactIntegral
-import sigma.data.{DataValueComparer, KeyValueColl, Nullable, RType, SigmaConstants}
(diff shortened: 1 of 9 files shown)


data/shared/src/main/scala/sigma/ast/methods.scala
```diff
@@ -1957,10 +1956,10 @@ case object SGlobalMethods extends MonoTypeMethods {
     )
 
   def powHit_eval(mc: MethodCall, G: SigmaDslBuilder, k: Int, msg: Coll[Byte], nonce: Coll[Byte], h: Coll[Byte], N: Int)
-                 (implicit E: ErgoTreeEvaluator): BigInt = {
+                 (implicit E: ErgoTreeEvaluator): UnsignedBigInt = {
     val cost = PowHitCostKind.cost(k, msg, nonce, h)
     E.addCost(FixedCost(cost), powHitMethod.opDesc)
-    CBigInt(Autolykos2PowValidation.hitForVersion2ForMessageWithChecks(k, msg.toArray, nonce.toArray, h.toArray, N).bigInteger)
+    CUnsignedBigInt(Autolykos2PowValidation.hitForVersion2ForMessageWithChecks(k, msg.toArray, nonce.toArray, h.toArray, N).bigInteger)
```
aslesarenko, line 1962: Shouldn't the resulting big int be checked to into 256 bits?

kushti: > Shouldn't the resulting big int be checked to into 256 bits?

It is checked during CUnsignedBigInt construction (as well as non-negativity)

PR #1050, opened 2025-01-13
Getvar variant for 6.0 removed

kushti:
In this PR, Context.getVar got removed, as having method with the same id but different behavior is not supported by clients < 6.0.

Another option was to introduce a new method with different id, but I guess new method Context.getVarFromInput() would be enough (in conjuction with selfIndex when read from self is required)


sc/shared/src/main/scala/sigma/compiler/ir/GraphBuilding.scala
```diff
@@ -1082,11 +1082,7 @@ trait GraphBuilding extends Base with DefRewriting { IR: IRContext =>
               ctx.LastBlockUtxoRootHash
             case SContextMethods.minerPubKeyMethod.name =>
               ctx.minerPubKey
-            case SContextMethods.getVarV6Method.name =>
-              val c2 = asRep[Byte](argsV(0))
-              val c3 = stypeToElem(typeSubst.apply(tT))
-              ctx.getVar(c2)(c3)
-            case SContextMethods.getVarFromInputMethod.name =>
+            case SContextMethods.getVarFromInputMethod.name if VersionContext.current.isV6SoftForkActivated  =>
```
sethdusek: Needs rebase to new versioning scheme?

SethDusek (changes requested):

sc/shared/src/main/scala/sigma/compiler/ir/GraphBuilding.scala
kushti: Updated

SethDusek (approved):

PR #1055, opened 2025-03-04
I1039 Finishes executeFromSelfReg

### LINKED ISSUE #1039: Finish executeFromSelfReg implementation
Like executeFromVar implementation in https://github.com/ergoplatform/sigmastate-interpreter/pull/976 , do executeFromSelfReg ErgoScript function for ErgoTree node already done

novamon:
Implements executeFromSelfReg like executeFromVar. Closes #1039.

### DIFF
--- a/data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
+++ b/data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
@@ -4,11 +4,16 @@ import org.ergoplatform.ErgoAddressEncoder.NetworkPrefix
 import org.ergoplatform.{ErgoAddressEncoder, P2PKAddress}
 import org.ergoplatform.ErgoBox.RegisterId
 import scorex.util.encode.{Base16, Base58, Base64}
+import sigma.data._
+import sigma.{Colls}
 import sigma.ast.SCollection.{SByteArray, SIntArray}
 import sigma.ast.SOption.SIntOption
 import sigma.ast.SigmaPropConstant
 import sigma.ast.syntax._
 import sigma.data.Nullable
+import sigma.data.RType.asType
+import sigma.Evaluation.stypeToRType
+import scala.reflect.ClassTag
 import sigma.exceptions.InvalidArguments
 import sigma.serialization.CoreByteWriter.ArgInfo
 import sigma.serialization.ValueSerializer
@@ -384,7 +389,7 @@ object SigmaPredef {
         Seq(ArgInfo("id", "identifier of the context variable")))
     )
 
-    val ExecuteFromSelfRegFunc = PredefinedFunc("executeFromSelfReg",
+    val ExecuteFromSelfRegWithDefaultFunc = PredefinedFunc("executeFromSelfRegWithDefault",
       Lambda(
         Seq(paramT),
         Array("id" -> SInt, "default" -> tT),
@@ -408,6 +413,50 @@ object SigmaPredef {
          | Type parameter \lst{T} result type of the deserialized script.
          | Throws an exception if the actual script type doesn't conform to \lst{T}.
          | Returns a result of the script execution in the current context
+        """.stripMargin,
+        Seq(ArgInfo("id", "identifier of the register"),
+          ArgInfo("default", "optional default value, if register is not available")))
+    )
+
+    val ExecuteFromSelfRegFunc = PredefinedFunc("executeFromSelfReg",
+      Lambda(
+        Seq(paramT),
+        Array("id" -> SInt),
+        tT, None
+      ),
+      PredefFuncInfo(
+        { case (Ident(_, SFunc(_, rtpe, _)), Seq(id: Constant[SNumericType]@unchecked)) =>
+          val idx: Int = SInt.downcast((id.value.asInstanceOf[AnyVal]))
+          if (idx < 0 || idx >= org.ergoplatform.ErgoBox.allRegisters.length) {
+            throw new InvalidArguments(s"Invalid register specified $id")
+          }
+
+          val d : Option[Value[rtpe.type]] = rtpe match {
+            case SInt       => Some(IntConstant(0).asValue)
+            case SBoolean   =>
+              //.withSrcCtx(currentSrcCtx.value).asInstanceOf[Constant[T]]
+                Some(BooleanConstant(false).withSrcCtx(currentSrcCtx.value))
+            case SByte      => Some(ByteConstant(0).asValue)
+            case SLong      => Some(LongConstant(0L).asValue)
+            case SBigInt    => Some(BigIntConstant(0).asValue)
+            case SSigmaProp => Some(SigmaPropConstant(CSigmaProp(false)).asValue)
+            case c: SCollection[a] =>
+              implicit val elemRType: RType[a#WrappedType] = asType(stypeToRType(c.elemType))
+              implicit val elemCT: ClassTag[a#WrappedType] = elemRType.classTag
+              Some(CollectionConstant(Colls.fromArray(elemRType.emptyArray), c.elemType).asValue)
+            case _ => None
+          }
+
+          val r: RegisterId = org.ergoplatform.ErgoBox.registerByIndex(idx)
+          mkDeserializeRegister[rtpe.type](r, rtpe, d)
+        }),
+      OperationInfo(DeserializeRegister,
+        """Extracts SELF register as \lst{Coll[Byte]}, deserializes it to script
+          | and then executes this script in the current context.
+          | The original \lst{Coll[Byte]} of the script is available as \lst{SELF.getReg[Coll[Byte]](id)}.
+          | Type parameter \lst{T} result type of the deserialized script.
+          | Throws an exception if the actual script type doesn't conform to \lst{T}.
+          | Returns a result of the script execution in the current context
         """.stripMargin,
         Seq(ArgInfo("id", "identifier of the register"),
           ArgInfo("default", "optional default value, if register is not available")))
@@ -440,6 +489,7 @@ object SigmaPredef {
       AvlTreeFunc,
       SubstConstantsFunc,
       ExecuteFromVarFunc,
+      ExecuteFromSelfRegWithDefaultFunc,
       ExecuteFromSelfRegFunc
     ).map(f => f.name -> f).toMap
 
--- a/data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
+++ b/data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
@@ -391,10 +391,14 @@ object SigmaPredef {
         tT, None
       ),
       PredefFuncInfo(
-        { 
-        case (Ident(_, SFunc(_, rtpe, _)), Seq(id: Constant[SNumericType]@unchecked, default)) =>
-          val r: RegisterId = org.ergoplatform.ErgoBox.registerByIndex(SByte.downcast(id.value.asInstanceOf[AnyVal]))
-          mkDeserializeRegister[rtpe.type](r, rtpe, Some(default.asValue[rtpe.type]))
+        { case (Ident(_, SFunc(_, rtpe, _)), Seq(id: Constant[SNumericType]@unchecked, default)) =>
+          val idx: Int = SInt.downcast((id.value.asInstanceOf[AnyVal]))
+          if (idx < 0 || idx >= org.ergoplatform.ErgoBox.allRegisters.length) {
+            default.v.asOption[rtpe.type].get
+          } else {
+            val r: RegisterId = org.ergoplatform.ErgoBox.registerByIndex(idx)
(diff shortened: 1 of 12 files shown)


sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
```diff
@@ -203,6 +205,112 @@ class BasicOpsSpecification extends CompilerTestingCommons
...
+
+  // executeFromSelfReg failure tests - Most fail the same way as these
+  // Coll[Bytes] doesn't and attempts to get the default value but fails
+
+  property("executeFromSelfReg - IReduct") {
+    val bytes = Plus(IntConstant(2), IntConstant(3))
+    val scriptBytes = ValueSerializer.serialize(bytes)
+
+    test("executeFromSelfReg", env, ext,
+      "{ executeFromSelfReg[Int](4, getVar[Int](1)) == 5 }",
```
kushti: should be == 1 , then test if passing (with == 5 it is reducing to false as it should be)

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
```diff
@@ -203,6 +205,112 @@ class BasicOpsSpecification extends CompilerTestingCommons
...
+      additionalRegistersOpt = Some(Map())
+    )
+  }
+
+  property("executeFromSelfReg - IThrows") {
+    val bytes = Plus(IntConstant(2), IntConstant(3))
+    val scriptBytes = ValueSerializer.serialize(bytes)
+
+    test("executeFromSelfReg", env, ext,
+      "{ executeFromSelfReg[Int](4, getVar[Int](1)) == 5 }",
```
kushti: why is this throwing exception at all ?

kushti (changes requested):
Please update LangSpec.md as well

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
novamon: These tests are designed to fail. That's not the issue. The issue is the way it's failing. i.e.
`
property("executeFromSelfReg - IThrows")(ActivatedVersion = 0; ErgoTree version = 0)
[info] BasicOpsSpecification:
[info] - executeFromSelfReg - IThrows *** FAILED ***
[info]   java.lang.RuntimeException: Should be overriden in class sigma.ast.DeserializeRegister: DeserializeRegister(R4,SInt$,Some(GetVar(1,Option[SInt$])))
[info]   at scala.sys.package$.error(package.scala:27)
`

It fails because additionalRegistersOpt is left out on purpose to create a reason to return the default value. If the register is "unavailable", I took to mean not useable or just invalid at all.

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
novamon: It reduces to false no matter what it's equal to. The test below is what tipped me off to a possible issue in DeserializeRegister, and questioning if the value is correctly being returned.

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
kushti: In case of register being missed, default value should be returned, not exception, see DeserializeRegister definition (ErgoLikeInterpreter : 17)

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
kushti: You can start with "executeFromSelfReg[Int](4, 1) == 1" , it should be true . I guess there is some issue on compiler level, try to catch ErgoTree (syntax tree) maybe , find an issue there, and use the tree in the test instead of `null`

sc/shared/src/test/scala/sigmastate/utxo/BasicOpsSpecification.scala
novamon: That was exactly my point 👍 Either way it forced me to find an oddity I didn't expect on script exec. Which explained the reduced to false behaviour, etc.. It's all cleaned up, it should function correctly.

data/shared/src/main/scala/sigma/ast/transformers.scala
```diff
@@ -548,11 +550,23 @@ object DeserializeContext extends ValueCompanion {
 }
 
 /** Extract register of SELF box as Coll[Byte], deserialize it into Value and inline into executing script.
-  * NOTE: it only applicable to SELF box
+  * NOTE: it's only applicable to SELF box
   */
 case class DeserializeRegister[V <: SType](reg: RegisterId, tpe: V, default: Option[Value[V]] = None) extends Deserialize[V] {
   override def companion = DeserializeRegister
   override val opType = SFunc(Array(SBox, SByte, SOption(tpe)), tpe)
+  protected final override def eval(env: DataEnv)(implicit E: ErgoTreeEvaluator): Any = {
```
kushti: What is the purpose of changes here? This is about changing rules of AST node interpretation, so breaking consensus. Should be avoided then.

kushti (changes requested):

data/shared/src/main/scala/sigma/ast/transformers.scala
novamon: The only place I could see adding this was here because of the execution hierarchy and relation to eval itself. The only test that fails without it is ForceDefault. DeserializeRegister without this override will miss an exception from eval which is a generic RuntimeException:
``[info] - executeFromSelfReg - ForceDefault *** FAILED ***
[info]   java.lang.RuntimeException: Should be overriden in class sigma.ast.DeserializeRegister``

I can catch the exception but this seemed more appropriate per the spec of DeserializeRegister, in the spec docs, specifically the pdf. It doesn't seem DeserializeRegister is implementing the tag value and honoring the default value constraint per the spec in the pdf. This allows it to honor only the default value, otherwise it will rely on eval to do the work and eval won't care about an empty register, etc. i.e. the codebase vs the pdf generated spec isn't inline. It still doesn't fix the difference to the spec with the tag value but that isn't needed with the check. Eval is still involved and operates where it's needed here.

novamon: The alternative here builds the tree for the check but again, the default value doesn't get honored in DeserializeRegister because of the forced condition and exceptions. Nearly every function along the chain throws an exception if error. Resulting in 
``Cannot getReg[Int](4): invalid type of value TestValue(SigmaProp(ProveDlog(ECPoint(f59ad3,c24b9a,...)))) at id=4
sigma.exceptions.InvalidType: Cannot getReg[Int](4): invalid type of value TestValue(SigmaProp(ProveDlog(ECPoint(f59ad3,c24b9a,...)))) at id=4
	at sigma.data.CBox.getReg(CBox.scala:41)
``
Which bypasses DeserializeRegister and any other handling. Remember, according to DeserializeRegisters info in Operations.scala @ line 192
```
  object DeserializeRegisterInfo extends InfoObject {
    private val func = predefinedOps.funcs("executeFromSelfReg")
```
executeFromSelfReg is just a facade for it and I think this issue wasn't tested before with DeserializeRegister. I can't find a test for default value being returned anywhere.

https://github.com/novamon/sigmastate-interpreter/tree/i1039-alt

novamon: Reverted the function - Should be fine now

docs/LangSpec.md
```diff
@@ -1017,6 +1017,39 @@ def PK(input: String): SigmaProp
...
+ * 
+ * @param id register id holding the serialized script to execute
+ * @param default value that's returned if the register is unavailable
+ * @tparam T expected type of the register and return type.
+ * @return result of the executed script or default value
+ * @throws InterpreterException when a script reduces to false
+ * @throws InvalidType exception when the result type of the execution value is 
+ *                     different from T.
+ */
+def executeFromSelfReg[T](id: Int, default: Option[T]): T
```
jozanek: Shouldn't default value by of type T? What happens if default is None, what is returned in such case?

sc/shared/src/main/scala/sigma/compiler/ir/Base.scala
```diff
@@ -211,6 +211,13 @@ abstract class Base { thisIR: IRContext =>
     override def resultType: Elem[V#WrappedType] = e
   }
 
+  /**
+   * Def done in order to carry on DeserializeRegister through stages of compilation intact
+   */
+  case class DeserializeRegisterDef[V <: SType](d: DeserializeRegister[V], t: Elem[V#WrappedType], e: Elem[V#WrappedType]) extends Def[V#WrappedType] {
+    override def resultType: Elem[V#WrappedType] = t
+  }
```
jozanek, line 219: `e` is never used, why to have it? Also current usages suggest t == e, this also hints to drop `e`.

novamon: dropped

docs/LangSpec.md
novamon: ErgoScript 5.x, so far as I know, Some & None aren't available. Those were merged in 6.X branch. After the original work was started. This was implemented per spec, which was default: Option[T]

You end up with output like this from this script execution "{ executeFromSelfReg[Int](4, None) == 2 }":
```
property("executeFromSelfReg - None")(ActivatedVersion = 0; ErgoTree version = 0)
[info] BasicOpsSpecification:
[info] - executeFromSelfReg - None *** FAILED ***
[info]   sigma.exceptions.TyperException: line 1: { executeFromSelfReg[Int](4, None) == 2 }
[info]                                      ^
[info] Cannot assign type for variable 'None' because it is not found in env HashMap(bit_& -> [T](T,T) => T,
```

data/shared/src/main/scala/sigma/ast/transformers.scala
```diff
@@ -4,13 +4,15 @@ import org.ergoplatform.ErgoBox.RegisterId
...
 import sigma.ast.syntax.SValue
+import sigma.data.RType.asType
 import sigma.data.{CSigmaProp, RType}
 import sigma.eval.ErgoTreeEvaluator
 import sigma.eval.ErgoTreeEvaluator.DataEnv
 import sigma.serialization.CoreByteWriter.ArgInfo
 import sigma.serialization.OpCodes
 import sigma.serialization.ValueCodes.OpCode
 import sigma.{Box, Coll, Evaluation}
+import sigma.exceptions.InvalidType
```
jozanek: Are those 2 new imports needed at all?

docs/LangSpec.md
```diff
@@ -1017,6 +1017,39 @@ def PK(input: String): SigmaProp
...
+ * 
+ * @param id register id holding the serialized script to execute
+ * @param default value returned if the register is unavailable
+ * @tparam T expected type of the register and return type.
+ * @return result of the executed script or default value
+ * @throws InterpreterException when a script reduces to false
+ * @throws InvalidType exception when the result type of the execution value is 
+ *                     different from T.
+ */
+def executeFromSelfReg[T](id: Int, default: Option[T]): T
```
jozanek: `default: T`

novamon: noted, incoming.

data/shared/src/main/scala/sigma/ast/transformers.scala
novamon: No, and noted.. in next push.

data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
```diff
@@ -383,25 +389,78 @@ object SigmaPredef {
...
+        tT, None
+      ),
+      PredefFuncInfo(
+        { case (Ident(_, SFunc(_, rtpe, _)), Seq(id: Constant[SNumericType]@unchecked)) =>
+          val idx: Int = SInt.downcast((id.value.asInstanceOf[AnyVal]))
+          if (idx < 0 || idx >= org.ergoplatform.ErgoBox.allRegisters.length) {
+            throw new InvalidArguments(s"Invalid register specified $id")
+          }
+
+          val d : Option[Value[rtpe.type]] = rtpe match {
```
kushti: Better to avoid undocumented defaults hardcoded in the compiler. so set val d = None just, (and then ForceDefault test for executeFromSelfReg can be removed just)

kushti (approved):
approved, but please address notes!

data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
novamon: It wasn't because of ForceDefault. It was to allow the script reduction to take place. If you send None value through DeserializeRegister as a default value, you'll end up with a java.lang.RuntimeException, which blocks scripts reducing to false. Removing the default values and those two tests.

PR #1069, opened 2025-04-21
Context improvements and possibility to reject transactions using "soft" preheader fields

kushti:
In this PR, we add new flag to execution context to allow for prohibition for soft fields usage in a script, where a soft field is a preheader field which value is defined by a miner, not the protocol (miner pubkey, timestamp, votes). 

Needed for input-blocks validation in Ergo.


build.sbt
```diff
@@ -65,11 +65,11 @@ lazy val commonSettings = Seq(
...
 lazy val crossScalaSettings = Seq(
   crossScalaVersions := Seq(scala213, scala212, scala211),
-  scalaVersion := scala213
+  scalaVersion := scala212
 )
 lazy val crossScalaSettingsJS = Seq(
-  crossScalaVersions := Seq(scala213),
-  scalaVersion := scala213
+  crossScalaVersions := Seq(scala212),
+  scalaVersion := scala212
```
jozanek, line 72: What was intention behind this donwgrade?

core/shared/src/main/scala/sigma/SigmaDsl.scala
```diff
@@ -966,7 +972,7 @@ trait SigmaDslBuilder {
   def BigInt(n: BigInteger): BigInt
 
   /** Create DSL unsigned big integer from existing `java.math.BigInteger`*/
-  def UnsignedBigInt(n: BigInteger): UnsignedBigInt
+  def unsignedBigInt(n: BigInteger): UnsignedBigInt
```
jozanek, line 975: breaking change

core/shared/src/main/scala/sigma/SigmaDsl.scala
```diff
@@ -796,6 +796,12 @@ trait Context {
   /** The version of ErgoTree currently executed by interpreter.
     * See [[ErgoLikeContext]] class for details. */
   def currentErgoTreeVersion: Byte
+
+  /**
+    * @return whether pre-header fields which may vary from pre-header to pre-header (i.e. not derived
+    *         deterministically from previous header but set by miner). There are
+    */
```
jozanek, line 803: unfinished doc?

interpreter/shared/src/main/scala/sigmastate/eval/CContext.scala
```diff
@@ -45,7 +46,13 @@ case class CContext(
...
   @inline override def LastBlockUtxoRootHash = lastBlockUtxoRootHash
 
-  @inline override def minerPubKey = _minerPubKey
+  @inline override def minerPubKey: Coll[Byte] = {
+    if(softFieldsAllowed) {
+      _minerPubKey
+    } else {
+      throw new SoftFieldAccessException("minerPubKey")
+    }
+  }
```
jozanek, line 55: Two notions of "soft field" now exist with no single source of truth (other being data/ast/values.scala PreHeader methods, via softMethodIds). Defensible, but the split is undocumented and easy to let drift (e.g. if a new soft field is added, you must remember both sites).

sc/shared/src/test/scala/sigmastate/TestingInterpreterSpecification.scala
```diff
@@ -496,6 +507,51 @@ class TestingInterpreterSpecification extends CompilerTestingCommons
...
+    val msg = Array.fill(32)(Random.nextInt(100).toByte)
+    val proof = NoProof
+    val env = testingContext(614401, softFieldsAllowed = true)
+    verifier.verify(prop, env, proof, msg).map(_._1).getOrElse(false) shouldBe true
+
+    assertExceptionThrown(
+      verifier.verify(prop, testingContext(614401, softFieldsAllowed = false), proof, msg).get,
+      rootCause(_).isInstanceOf[SoftFieldAccessException]
+    )
+  }
```
jozanek, line 553: These five properties all prove the flag rejects soft-field access - but none prove it doesn't over-reject. For input-block validation the more dangerous direction is a flag that silently invalidates ordinary scripts when switched off.

Could you add a negative-control property asserting that with softFieldsAllowed = false:

- the non-soft preHeader fields still verify (version, parentId, nBits, height) — these hit the same MethodCall.eval guard, so this is what proves the guard discriminates by methodId rather than blanket-rejecting any SPreHeader access; and
- ordinary context access (HEIGHT, INPUTS, SELF) is unaffected.

testEval(code, softFieldsAllowed = false) already asserts the script verifies to true, so each case is a one-liner. A bug where the guard read "receiver is SPreHeader → reject" instead of "SPreHeader and methodId ∈ softMethodIds → reject" would pass every current test and be caught only by this one.

data/shared/src/main/scala/sigma/ast/values.scala
```diff
@@ -1345,6 +1345,14 @@ case class MethodCall(
           argsBuf(len + i) = extra(i)
         }
         var res: Any = null
+
+        // Check if accessing soft fields of PreHeader when soft fields are not allowed
+        if (method.objType.typeId == SPreHeader.typeId &&
+              !E.context.softFieldsAllowed &&
+              SPreHeaderMethods.softMethodIds.contains(method.methodId)) {
+          throw new SoftFieldAccessException(method.name)
+        }
```
jozanek, line 1383: This is a consensus-critical path. Please add an in-source note that the softFieldsAllowed = true default preserves all released-version behavior, the flag is only false in input-block validation, and the node must never pass false on the normal full-block path. The HF-safety of this whole PR rests on that invariant, and right now it's nowhere in the code.

Issue #1028, opened 2024-10-10
Missing Automatic-Module-Name


satsen:

shubham5080:
@kushti i  fixed this issue , Please review PR .

PR #1100, opened 2025-12-13
Add predefined function to strip ErgoTree header byte

### LINKED ISSUE #1075: Make an ErgoScript function to compare two instances ErgoTree without headers
example how tree hash is compared with expected , with no header byte considered 

      val receiptTreeHash = blake2b256(receiptOutputErgoTree.slice(1, receiptOutputErgoTree.size))

Make a high-level ErgoScript function to skip header byte

shubham5080:
PR Title

Add predefined function to strip ErgoTree header byte

PR Description

This PR introduces a new predefined ErgoScript function for working with serialized ErgoTrees without the header byte, addressing the use case described in issue #1075 & #1028.

Summary of changes

Added a high-level predefined function that returns serialized ErgoTree bytes with the header byte removed.

This allows safe hashing and comparison of ErgoTrees without requiring manual slicing in user scripts.

The implementation is cross-platform (JVM + Scala.js) and does not affect existing cost models or serialization formats.

Test updates and rationale

Two test adjustments were required during integration:

SigmaCompilerTest

The compiler/type-checker lowers the predefined function into an equivalent Slice over the serialized ErgoTree bytes.

The test expectation was updated to reflect this lowered form, matching the compiler’s actual output rather than asserting a non-lowered AST node.

TestingInterpreterSpecification

The original test attempted to serialize and deserialize a constant value produced by the new function, which is not a valid standalone serialized value and caused a deserialization error.

This check was replaced with an equivalent runtime comparison of the sliced bytes, keeping the test focused on the intended semantics (header removal) rather than serialization internals.

Both test suites now pass:

SigmaCompilerTest

TestingInterpreterSpecification

Design notes

The function is intended for hashing and comparison of ErgoTrees and does not expose invalid serialized structures as deserializable values.

Lowering to existing primitives (Slice) is intentional and keeps the compiler and interpreter behavior consistent and stable.

No existing opcodes were modified or renumbered.

Verification

Focused tests can be run with:

sbt "testOnly SigmaCompilerTest"
sbt "testOnly TestingInterpreterSpecification"

All tests pass successfully.

Follow-ups (optional)

If preferred in the future, the function could be preserved as a first-class compiled node rather than lowering to Slice. This would require additional graph-builder and serializer support and is intentionally out of scope for this PR.

This PR adds an Automatic-Module-Name entry to the JAR manifest,
providing a stable Java module name for Java 9+ users.

Chosen module name:
org.ergoplatform.sigmastate.interpreter

The attribute is added via shared JVM build settings, so it
applies to all generated JVM artifacts. This change is fully
backward-compatible and does not affect runtime behavior.

Fixes: #1075 & #1028.


shubham5080:
please check does this PR working or not.

PR #1115, opened 2026-04-14
Candidate for  6.0.4 release 

kushti:
* #1110 
* #1118 
* #1126 
* #1130 
* #1131
* #1136 & #1138
* #1152
* JavaScript dependencies update
* more tests in BasicOpsSpecification, to get full coverage of LangSpec.md with simple examples

Contributors : @a-shannon , @jozanek , @kushti


a-shannon (changes requested):
[P1] Bare `None` bypasses V6 method gating

`SigmaTyper` now handles `Ident("None")` by calling `SGlobalMethods.noneMethod` directly. That bypasses the normal version-filtered method lookup, so bare `None` typechecks even under `V6SoftForkVersion - 1`.

I confirmed with a scratch test:

```scala
{ val X: Option[Int] = None; X }
```

This should fail pre-V6, like `Global.none[Int]()` does, but it currently typechecks. Please route this through the same version-aware lookup path as `Global.none`, or explicitly reject bare `None` before V6. A negative compatibility test should be added beside the new `None` inference tests.

Verification I ran:

- `sbt "scJVM/testOnly sigmastate.lang.SigmaTyperTest -- -z None"` passed the existing `None` tests.
- A scratch negative test for pre-V6 bare `None` failed, confirming the regression.
- `sbt "scJVM/testOnly sigmastate.utxo.BasicOpsSpecification"` passed 211 tests, 1 ignored.

a-shannon (approved):
The previous bare `None` concern looks addressed on the current head.

What I checked:
- bare `None` no longer bypasses version gating: pre-V6 it falls through to the unresolved-identifier path, while V6 only accepts it when an `Option[T]` context is available
- `None` inference is covered for explicit `val` type ascription, sibling `if` branch inference, untyped bare `None`, and pre-V6 `Global.none[Int]()` rejection
- `executeFromSelfRegWithDefault` now returns the default expression for out-of-range register ids, while `executeFromSelfReg` still rejects invalid ids; tests cover valid `SigmaProp` / `Coll[Byte]` / `Int`, missing-register default, invalid `99`, boundary `10`, and negative id for the non-default form
- Bulletproof range-proof tests pass for the focused Scala/JVM cases, including negative tampered/out-of-range checks

Checks I ran:
- `sbt "scJVM/testOnly sigmastate.lang.SigmaTyperTest -- -z None" "scJVM/testOnly sigmastate.utxo.BasicOpsSpecification -- -z executeFromSelfRegWithDefault" "scJVM/testOnly sigmastate.utxo.BasicOpsSpecification -- -z Bulletproof"` passed
- `npm audit --omit=dev` reports 0 vulnerabilities
- `git diff --check` passed

GitHub CI status from this pass:
- Scala jobs pass on 2.11.12, 2.12.21, and 2.13.18
- JS build/tests pass, including 10 Jest suites / 41 tests
- the only red check is the final `npm publish --tag snapshot` step, failing after tests pass with npm 404; that looks like npm publish token/package permission configuration rather than a test failure

No code blockers from my pass.

PR #1120, opened 2026-05-10
use secp256k1-jni for EC scalar multiplication, BC as fallback

### LINKED ISSUE #970: Consider using secp256k1-jni from Bitcoin-s
Reportedly:

```
[Libsecp256k1](https://github.com/bitcoin-core/secp256k1) is used to preform cryptographic operations on the secp256k1 curve. This is the curve that bitcoin uses. There is a signficant speedup when using this library compared to java crypto libraries like bouncy castle.
```

(from https://bitcoin-s.org/docs/secp256k1/secp256k1 )

So it is worth to try to use secp256k1-jni and so use BouncyCastle as a failback option only (for platforms where a native lib wont load)

jozanek:
Adds bitcoin-s-secp256k1jni as a JVM-only dependency and routes exponentiatePoint through NativeSecp256k1.pubKeyTweakMul when the native library is available, falling back to BouncyCastle otherwise.

Closes #970

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -204,7 +204,8 @@ lazy val core   = crossProject(JVMPlatform, JSPlatform)
   .jvmSettings(
     crossScalaSettings,
     libraryDependencies ++= Seq(
-      bouncycastleBcprov
+      bouncycastleBcprov,
+      "org.bitcoin-s" % "bitcoin-s-secp256k1jni" % "1.9.10"
     )
   )
   .jsSettings(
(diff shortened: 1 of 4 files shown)


kushti:
is it possible to add a property test that EC ops always produce identical results (and also unit tests for edge cases)

jozanek:
Added `SecP256K1NativeSpec`: property tests comparing native vs BC on random points/scalars, plus unit tests for edge cases (0, ±1, order±1, multiples of the order, 2^255, 2^256±k, infinity, concurrency).

It caught a bug in the original commit: scalars >= 2^256 had their leading byte dropped, so native returned p·(n mod 2^256). Not reachable from ErgoTree's 256-bit `BigInt`, but fixed anyway by reducing the scalar mod order before the native call.

PR #636, opened 2020-01-26
Test for proveDHTuple(g, g^x, g^y, g^xy) should work for x and not y

kushti:
A new test added in regards with DH proving. Namely, the new test is checking that proveDHTuple(g, g^x, g^y, g^xy) should work for x and not y.

PR #1127, opened 2026-05-12
reject improper DHT secrets at construction and signing time

### LINKED ISSUE #638: Signing for DHT could work for improper secret
Currently for DHT secret prover can generate a proof for secret not really satisfying a statement to be proven. See new failing "DH tuple - signing w. proper and improper secrets" test in https://github.com/ScorexFoundation/sigmastate-interpreter/pull/636/ 

Happily, verification fails, so there is no problem with the Ergo protocol.

jozanek:
Closes #638 

Note: Originally I wanted to make the case-class primary constructor private. On 2.12/2.13 this worked fine, but there's bug in 2.11. Maybe it would be good time to reconsider dropping of old unsupported 2.11.x?

### DIFF
--- a/interpreter/js/src/main/scala/sigma/interpreter/js/ProverSecret.scala
+++ b/interpreter/js/src/main/scala/sigma/interpreter/js/ProverSecret.scala
@@ -42,7 +42,7 @@ object ProverSecret extends js.Object {
   def dht(w: js.BigInt, dhtProp: SigmaProp): ProverSecret = {
     dhtProp.sigmaBoolean match {
       case dht: ProveDHTuple =>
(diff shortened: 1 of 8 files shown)


kushti (approved):
approved but please import the test from https://github.com/ergoplatform/sigmastate-interpreter/pull/636 as well

jozanek:
Imported the test from #636 as "DH tuple - signing w. proper and improper secrets". Two adaptations: the original `prove(...).get shouldBe false` compared a prover result to a Boolean and was an intentionally failing repro, so it is now `Failure(InterpreterException)` from the signing-time check; and the improper secret is built via the primary constructor since `create` rejects it.

PR #1138, opened 2026-05-16
Fix pre-V6 gate for bare `None` type inference

### LINKED ISSUE #398: Correct type inference for Option.None
Example expression `if(b.R5[Int].isDefined) {None} else {(pk, value)}`.
SigmaTyper should be fixed, in particular method 
`def assignType(env: Map[String, SType],
                 bound: SValue,
                 expected: Option[SType] = None)`
Hint, `expected` should be passed and used.

jozanek:
Closes #398

### DIFF
--- a/sc/shared/src/main/scala/sigma/compiler/phases/SigmaTyper.scala
+++ b/sc/shared/src/main/scala/sigma/compiler/phases/SigmaTyper.scala
@@ -1,5 +1,6 @@
 package sigma.compiler.phases
 
+import sigma.VersionContext
 import sigma.ast.SCollection.{SBooleanArray, SByteArray}
 import sigma.ast.SigmaPredef._
 import sigma.ast._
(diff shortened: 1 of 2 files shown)


sc/shared/src/main/scala/sigma/compiler/phases/SigmaTyper.scala
```diff
@@ -94,35 +98,36 @@ class SigmaTyper(val builder: SigmaBuilder,
       val newItems = items.map(assignType(env, _))
       assignConcreteCollection(c, newItems)
 
-    case i @ Ident("None", _) =>
```
kushti, line 97: can this case be preserved but versioned checks made inside ? complicating Ident processing for all the identifiers is not that good option

kushti (changes requested):

sc/shared/src/main/scala/sigma/compiler/phases/SigmaTyper.scala
jozanek: updated

kushti (approved):

PR #1086, opened 2025-11-03
feat: Add 6 compiler-level utility functions for ErgoScript box verif…

bigpandamx:
# Add 6 Compiler-Level Utility Functions for ErgoScript Box Verification

**Resolves:** #1037  
**Type:** Feature Implementation  
**Compatibility:** Sigma 6.0+

## 📋 Summary

This PR implements 6 utility functions requested by @bdkent for common ErgoScript box verification operations. These functions are now available as **global compiler-level utilities**, eliminating the need for developers to copy-paste utility code across contracts.

### 1. `verifyBoxHasMarkerToken(box: Box, tokenId: Coll[Byte]): Boolean`
Verifies a box contains a specific token ID with amount ≥ 1.

### 2. `verifyBoxHasNoMarkerToken(box: Box, tokenId: Coll[Byte]): Boolean`  
Verifies a box does NOT contain a specific token ID.

### 3. `verifyUsedAdditionalRegisters(box: Box, expectedCount: Int): Boolean`
Validates that a box uses only the specified number of additional registers (R4-R9).

### 4. `verifySameForBasicRequiredRegisters(inBox: Box, outBox: Box): Boolean`
Compares `value` and `propositionBytes` between two boxes.

### 5. `verifySameForRequiredRegisters(inBox: Box, outBox: Box): Boolean`
Compares `value`, `propositionBytes`, and `tokens` between two boxes.

### 6. `verifySpentToken(inBox: Box, outBox: Box, tokenId: Coll[Byte], amount: Long): Boolean`
Verifies that a specific token amount was properly spent (reduced) between input and output boxes.

### Multi-Layer Integration
- **SigmaPredef.scala**: Added `PredefinedFunc` definitions with proper cost modeling (Method IDs 11-16)
- **methods.scala**: Added `SMethod` definitions with correct type signatures  
- **SigmaDsl.scala**: Added method signatures to `SigmaDslBuilder` and `SigmaContract` traits
- **CSigmaDslBuilder.scala**: Implemented concrete verification logic using modern API patterns
- **IR Layer**: Updated `SigmaDslImpl`, `SigmaDslUnit`, `GraphIRReflection` for compiler support

### API Modernization
- Uses `box.getReg(n).isDefined` instead of deprecated `box.R4[Any].isEmpty` pattern
- Clean case destructuring: `case (id, amount) =>` instead of `t._1`, `t._2` accessors
- Flattened parameter structure for better usability

## 📊 Files Changed

| **File** | **Purpose** | **Changes** |
[...]

PR #1176, opened 2026-06-11
Add compiler-level box verification util functions (closes #1037)

### LINKED ISSUE #1037: compiler level util functions
I find myself reusing these functions in my ergoscripts. It might be useful to add this kind of functionality to the global utils:

```scala
def verifyBoxHasMarkerToken(args: (Box, Coll[Byte])): Boolean = {
  val box = args._1
  val token = args._2
  box.tokens.exists({ (t: (Coll[Byte], Long)) => t._1 == token && t._2 >= 1L })
}

def verifyBoxHasNoMarkerToken(args: (Box, Coll[Byte])): Boolean = {
  val box = args._1
  val token = args._2
  box.tokens.forall({ (t: (Coll[Byte], Long)) => t._1 != token })
}

def verifyUsedAdditionalRegisters(args: (Box, Int)): Boolean = {
  val box = args._1
  val used = args._2
  (used >= 1 || box.R4[Any].isEmpty) &&
  (used >= 2 || box.R5[Any].isEmpty) &&
  (used >= 3 || box.R6[Any].isEmpty) &&
  (used >= 4 || box.R7[Any].isEmpty) &&
  (used >= 5 || box.R8[Any].isEmpty) &&
  (used >= 6 || box.R9[Any].isEmpty)
}

def verifySameForBasicRequiredRegisters(args: (Box, Box)): Boolean = {
  val inBox = args._1
  val outBox = args._2
  inBox.value == outBox.value &&
  inBox.propositionBytes == outBox.propositionBytes
}

def verifySameForRequiredRegisters(args: (Box, Box)): Boolean = {
  val inBox = args._1
  val outBox = args._2
  verifySameForBasicRequiredRegisters(args) &&
  inBox.tokens == outBox.tokens
}

def verifySpentToken(args: ((Box, Box), (Coll[Byte], Long))): Boolean = {
  val inBox = args._1._1
  val outBox = args._1._2
  val token = args._2._1
  val amount = args._2._2
  inBox.tokens.forall({ (it: (Coll[Byte], Long)) =>
    {
      if (it._1 == token) {
        outBox.tokens.exists({ (ot: (Coll[Byte], Long)) => it._1 == ot._1 && it._2 == ot._2 + amount })
      } else {
        outBox.tokens.exists({ (ot: (Coll[Byte], Long)) => it._1 == ot._1 && it._2 == ot._2 })
      }
    }
  })
}
```

Ergologica:
# Add compiler-level utility functions for common box verification patterns

Closes #1037

## Summary

This PR adds the 6 utility functions proposed in #1037 as **predefined global functions of the ErgoScript frontend** (`SigmaPredef.scala`). Each function is expanded at compile time into ErgoTree nodes that are already supported by the interpreter, so **no changes to the consensus-critical ErgoTree level** (serializers, opcodes, evaluator, `SigmaDsl`) are involved — addressing the review feedback that led to closing #1086.

| Function | Expands to |
|---|---|
| `verifySameForBasicRequiredRegisters(inBox: Box, outBox: Box): Boolean` | `inBox.value == outBox.value && inBox.propositionBytes == outBox.propositionBytes` |
| `verifySameForRequiredRegisters(inBox: Box, outBox: Box): Boolean` | the above `&& inBox.tokens == outBox.tokens` |
| `verifyUsedAdditionalRegisters(box: Box, used: Int): Boolean` | `(used >= 1 \|\| box.R4[Any].isEmpty) && ... && (used >= 6 \|\| box.R9[Any].isEmpty)` |
| `verifyBoxHasMarkerToken(box: Box, token: Coll[Byte]): Boolean` | `box.tokens.exists { t => t._1 == token && t._2 >= 1L }` |
| `verifyBoxHasNoMarkerToken(box: Box, token: Coll[Byte]): Boolean` | `box.tokens.forall { t => t._1 != token }` |
| `verifySpentToken(inBox: Box, outBox: Box, token: Coll[Byte], amount: Long): Boolean` | `inBox.tokens.forall { it => if (it._1 == token) outBox.tokens.exists { ot => it._1 == ot._1 && it._2 == ot._2 + amount } else outBox.tokens.exists { ot => it._1 == ot._1 && it._2 == ot._2 } }` |

The semantics follow the reference implementations given in #1037 (uncurried argument lists are used, consistently with the other predefined global functions).

## Implementation notes

- All functions are registered in `globalFuncs` of `PredefinedFuncRegistry` and are expanded by the typer via the existing `PredefinedFuncApply` mechanism (`PredefFuncInfo.irBuilder`), like `allOf`, `xorOf`, etc.
- The produced trees only use existing operations (`EQ`, `BinAnd`/`BinOr` (lazy), `Exists`/`ForAll`, `ExtractAmount`, `ExtractScriptBytes`, `ExtractRegisterAs`, `MethodCall(Box.tokens)`, `SelectField`, `If`, `ArithOp.Plus`), so the resulting `ErgoTree` serializes and evaluates in all script versions.
- `verifyUsedAdditionalRegisters` — one behavioral caveat, documented in the function's `OperationInfo`: the emptiness check is expressed as `R{i}[Any].isEmpty` (as in the issue's reference code). When a register beyond the first `used` ones **is** defined, `Box.getReg[Any]` throws `InvalidType` instead of returning `Some(_)`, so the script evaluation **fails** rather than returning `false`. Used as a guarding predicate this is the desired outcome (the spending attempt is rejected), but the value cannot be negated/composed to recover from the failing branch. As far as I can tell, current ErgoTree has no type-agnostic "register is defined" primitive that could avoid this.
- `verifySpentToken` follows the reference semantics literally: it requires the remainder `ot._2 == it._2 - amount` to be present in the output box, therefore spending a token **completely** (down to zero) yields `false` by design (covered by a test).

## Testing

New `UtilFunctionsSpecification` (in `sc/shared/src/test/scala/sigmastate/utxo/`) with 20 properties run across script versions via `CompilerCrossVersionProps`. Each function is covered with positive and negative cases; every script is compiled, wrapped into an `ErgoTree`, proven and verified through `ContextEnrichingTestProvingInterpreter` / `ErgoLikeTestInterpreter` against a real spending context (`SELF` + `OUTPUTS(0)` with configurable values, tokens and registers). The ErgoTree serialization roundtrip is exercised implicitly by proving/verification.

### DIFF
--- a/data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
+++ b/data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
@@ -2,7 +2,7 @@ package sigma.ast
 
 import org.ergoplatform.ErgoAddressEncoder.NetworkPrefix
 import org.ergoplatform.{ErgoAddressEncoder, P2PKAddress}
-import org.ergoplatform.ErgoBox.RegisterId
+import org.ergoplatform.ErgoBox.{R4, R5, R6, R7, R8, R9, RegisterId}
 import scorex.util.encode.{Base16, Base58, Base64}
 import sigma.data._
 import sigma.{Colls}
@@ -529,11 +529,19 @@ object SigmaPredef {
         Seq(ArgInfo("id", "identifier of the register")))
     )
 
+    /** Element type of the `tokens` collection of a box: (tokenId, amount). */
+    private val STokenElemType: SType = org.ergoplatform.ErgoBox.STokenType
+
+    /** Builds `box.tokens` as a typed collection of (Coll[Byte], Long) tuples. */
+    private def boxTokens(box: Value[SBox.type]): Value[SCollection[STuple]] =
+      mkMethodCall(box, SBoxMethods.tokensMethod, IndexedSeq.empty, Map.empty).asCollection[STuple]
+
     /** Checks that two boxes have equal monetary value and equal guarding script. */
     val VerifySameForBasicRequiredRegistersFunc = PredefinedFunc("verifySameForBasicRequiredRegisters",
       Lambda(Array("inBox" -> SBox, "outBox" -> SBox), SBoolean, None),
       PredefFuncInfo(
-        { case (_, Seq(inBox: Value[SBox.type]@unchecked, outBox: Value[SBox.type]@unchecked)) =>
+        { case (_, Seq(inBox: Value[SBox.type]@unchecked, outBox: Value[SBox.type]@unchecked))
+            if inBox.tpe == SBox && outBox.tpe == SBox =>
           mkBinAnd(
             mkEQ(mkExtractAmount(inBox), mkExtractAmount(outBox)),
             mkEQ(mkExtractScriptBytes(inBox), mkExtractScriptBytes(outBox))
@@ -550,15 +558,14 @@ object SigmaPredef {
     val VerifySameForRequiredRegistersFunc = PredefinedFunc("verifySameForRequiredRegisters",
       Lambda(Array("inBox" -> SBox, "outBox" -> SBox), SBoolean, None),
       PredefFuncInfo(
-        { case (_, Seq(inBox: Value[SBox.type]@unchecked, outBox: Value[SBox.type]@unchecked)) =>
-          val tokensIn = mkMethodCall(inBox, SBoxMethods.tokensMethod, IndexedSeq.empty, Map.empty)
-          val tokensOut = mkMethodCall(outBox, SBoxMethods.tokensMethod, IndexedSeq.empty, Map.empty)
+        { case (_, Seq(inBox: Value[SBox.type]@unchecked, outBox: Value[SBox.type]@unchecked))
+            if inBox.tpe == SBox && outBox.tpe == SBox =>
           mkBinAnd(
             mkBinAnd(
               mkEQ(mkExtractAmount(inBox), mkExtractAmount(outBox)),
               mkEQ(mkExtractScriptBytes(inBox), mkExtractScriptBytes(outBox))
             ),
-            mkEQ(tokensIn, tokensOut)
+            mkEQ(boxTokens(inBox), boxTokens(outBox))
           )
         }),
       OperationInfo(MethodCall,
@@ -573,18 +580,14 @@ object SigmaPredef {
     val VerifyUsedAdditionalRegistersFunc = PredefinedFunc("verifyUsedAdditionalRegisters",
       Lambda(Array("box" -> SBox, "used" -> SInt), SBoolean, None),
       PredefFuncInfo(
-        { case (_, Seq(box: Value[SBox.type]@unchecked, used: Value[SInt.type]@unchecked)) =>
+        { case (_, Seq(box: Value[SBox.type]@unchecked, used: Value[SInt.type]@unchecked))
+            if box.tpe == SBox && used.tpe == SInt =>
           def regIsEmpty(reg: RegisterId): BoolValue =
             mkLogicalNot(mkOptionIsDefined(
               mkExtractRegisterAs(box, reg, SOption(SAny)).asValue[SOption[SAny.type]]))
           // (used >= n || box.R{n+3}[Any].isEmpty) for n = 1..6 (registers R4..R9)
           val clauses: Seq[BoolValue] = Seq[(Int, RegisterId)](
-            1 -> org.ergoplatform.ErgoBox.R4,
-            2 -> org.ergoplatform.ErgoBox.R5,
-            3 -> org.ergoplatform.ErgoBox.R6,
-            4 -> org.ergoplatform.ErgoBox.R7,
-            5 -> org.ergoplatform.ErgoBox.R8,
-            6 -> org.ergoplatform.ErgoBox.R9
+            1 -> R4, 2 -> R5, 3 -> R6, 4 -> R7, 5 -> R8, 6 -> R9
           ).map { case (n, reg) => mkBinOr(mkGE(used, IntConstant(n)), regIsEmpty(reg)) }
           clauses.reduce[BoolValue](mkBinAnd)
         }),
@@ -592,27 +595,21 @@ object SigmaPredef {
         """Returns \lst{true} if \lst{box} uses only the first \lst{used} non-mandatory registers,
           | i.e. all registers with index greater than \lst{used} (counting R4 as 1, ..., R9 as 6)
           | are empty.
-          | NOTE: when a register beyond the first \lst{used} ones is defined, the emptiness check
-          | \lst{R{i}[Any].isEmpty} fails with an exception (instead of returning \lst{false}),
-          | so the script evaluation fails, which is the desired behavior when the function is
-          | used as a guarding predicate of a box.
+          | NOTE: when a register beyond the first \lst{used} ones is defined, checking that the
+          | register is empty (\lst{isEmpty} of the register read as \lst{Option[Any]}) fails with
+          | an exception instead of returning \lst{false}, so the script evaluation fails, which is
+          | the desired behavior when the function is used as a guarding predicate of a box.
         """.stripMargin,
         Seq(ArgInfo("box", "the box to check"),
             ArgInfo("used", "number of additional registers (R4..R9) allowed to be used")))
     )
 
-    /** Element type of the `tokens` collection of a box: (tokenId, amount). */
-    private val STokenElemType: SType = org.ergoplatform.ErgoBox.STokenType
-
-    /** Builds `box.tokens` as a typed collection of (Coll[Byte], Long) tuples. */
-    private def boxTokens(box: Value[SBox.type]): Value[SCollection[STuple]] =
-      mkMethodCall(box, SBoxMethods.tokensMethod, IndexedSeq.empty, Map.empty).asCollection[STuple]
-
     /** Checks that a box contains at least one unit of the given token. */
     val VerifyBoxHasMarkerTokenFunc = PredefinedFunc("verifyBoxHasMarkerToken",
       Lambda(Array("box" -> SBox, "token" -> SByteArray), SBoolean, None),
       PredefFuncInfo(
-        { case (_, Seq(box: Value[SBox.type]@unchecked, token: Value[SByteArray]@unchecked)) =>
+        { case (_, Seq(box: Value[SBox.type]@unchecked, token: Value[SByteArray]@unchecked))
+            if box.tpe == SBox && token.tpe == SByteArray =>
           val t = Ident("$t", STokenElemType).asValue[STuple]
           // t._1 == token && t._2 >= 1L
           val body = mkBinAnd(
@@ -631,7 +628,8 @@ object SigmaPredef {
     val VerifyBoxHasNoMarkerTokenFunc = PredefinedFunc("verifyBoxHasNoMarkerToken",
       Lambda(Array("box" -> SBox, "token" -> SByteArray), SBoolean, None),
       PredefFuncInfo(
-        { case (_, Seq(box: Value[SBox.type]@unchecked, token: Value[SByteArray]@unchecked)) =>
+        { case (_, Seq(box: Value[SBox.type]@unchecked, token: Value[SByteArray]@unchecked))
+            if box.tpe == SBox && token.tpe == SByteArray =>
           val t = Ident("$t", STokenElemType).asValue[STuple]
           // t._1 != token
           val body = mkNEQ(mkSelectField(t, 1.toByte).asValue[SByteArray], token)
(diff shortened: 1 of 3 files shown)


data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
```diff
@@ -529,7 +529,159 @@ object SigmaPredef {
         Seq(ArgInfo("id", "identifier of the register")))
     )
 
+    /** Checks that two boxes have equal monetary value and equal guarding script. */
+    val VerifySameForBasicRequiredRegistersFunc = PredefinedFunc("verifySameForBasicRequiredRegisters",
+      Lambda(Array("inBox" -> SBox, "outBox" -> SBox), SBoolean, None),
+      PredefFuncInfo(
+        { case (_, Seq(inBox: Value[SBox.type]@unchecked, outBox: Value[SBox.type]@unchecked)) =>
```
jozanek: The `@unchecked` patterns erase to arity-only checks, so a user-defined 2-arg `def` with the same name (any arg types) is rewritten by this builder after typing. Add explicit `tpe == SBox` / `SByteArray` guards so non-matching calls fall through to a normal type error. (Applies to all six builders.)

data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
```diff
@@ -529,7 +529,159 @@ object SigmaPredef {
...
+        Seq(ArgInfo("inBox", "the first box"), ArgInfo("outBox", "the second box")))
+    )
+
+    /** Same as [[VerifySameForBasicRequiredRegistersFunc]] but additionally requires equal tokens. */
+    val VerifySameForRequiredRegistersFunc = PredefinedFunc("verifySameForRequiredRegisters",
+      Lambda(Array("inBox" -> SBox, "outBox" -> SBox), SBoolean, None),
+      PredefFuncInfo(
+        { case (_, Seq(inBox: Value[SBox.type]@unchecked, outBox: Value[SBox.type]@unchecked)) =>
+          val tokensIn = mkMethodCall(inBox, SBoxMethods.tokensMethod, IndexedSeq.empty, Map.empty)
+          val tokensOut = mkMethodCall(outBox, SBoxMethods.tokensMethod, IndexedSeq.empty, Map.empty)
```
jozanek: Move the `boxTokens` helper (defined below at L609) above this function and reuse it here instead of the two inline `mkMethodCall`s.

data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
```diff
@@ -529,7 +529,159 @@ object SigmaPredef {
...
+            mkLogicalNot(mkOptionIsDefined(
+              mkExtractRegisterAs(box, reg, SOption(SAny)).asValue[SOption[SAny.type]]))
+          // (used >= n || box.R{n+3}[Any].isEmpty) for n = 1..6 (registers R4..R9)
+          val clauses: Seq[BoolValue] = Seq[(Int, RegisterId)](
+            1 -> org.ergoplatform.ErgoBox.R4,
+            2 -> org.ergoplatform.ErgoBox.R5,
+            3 -> org.ergoplatform.ErgoBox.R6,
+            4 -> org.ergoplatform.ErgoBox.R7,
+            5 -> org.ergoplatform.ErgoBox.R8,
+            6 -> org.ergoplatform.ErgoBox.R9
```
jozanek: Nit: import `ErgoBox.{R4, R5, R6, R7, R8, R9}` instead of fully-qualified names in the clause list.

data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
```diff
@@ -529,7 +529,159 @@ object SigmaPredef {
...
+                       token: Value[SByteArray]@unchecked, amount: Value[SLong.type]@unchecked)) =>
+          val it = Ident("$it", STokenElemType).asValue[STuple]
+          val itId = mkSelectField(it, 1.toByte).asValue[SByteArray]
+          val itAmount = mkSelectField(it, 2.toByte).asValue[SLong.type]
+          // builds: outBox.tokens.exists(ot => it._1 == ot._1 && it._2 == ot._2 [+ amount])
+          def existsInOut(addAmount: Boolean): Value[SBoolean.type] = {
+            val ot = Ident("$ot", STokenElemType).asValue[STuple]
+            val otId = mkSelectField(ot, 1.toByte).asValue[SByteArray]
+            val otAmount = mkSelectField(ot, 2.toByte).asValue[SLong.type]
+            val rhs = if (addAmount) mkPlus(otAmount, amount) else otAmount
```
jozanek, line 659: `Plus` is checked arithmetic: `ot._2 + amount` near `Long.MaxValue` makes evaluation throw instead of returning `false`. Mention this in the OperationInfo doc.

data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
```diff
@@ -529,7 +529,159 @@ object SigmaPredef {
...
+          }
+          // if (it._1 == token) <spent> else <unchanged>
+          val ifBody = mkIf(mkEQ(itId, token), existsInOut(addAmount = true), existsInOut(addAmount = false))
+          val outerLam = mkLambda(Array[(String, SType)]("$it" -> STokenElemType), SBoolean, Some(ifBody))
+          mkForAll(boxTokens(inBox), outerLam)
+        }),
+      OperationInfo(MethodCall,
+        """Returns \lst{true} if every token of \lst{inBox} is preserved in \lst{outBox}, except for
+          | the token with id \lst{token} whose amount in \lst{outBox} must be exactly \lst{amount} less
+          | than in \lst{inBox}.
```
jozanek, line 671: Doc omits two caveats users will hit: (1) if `inBox` holds no `token`, the `forall` is vacuously true — the guard passes with nothing spent; (2) full spend (down to zero) returns `false`. Add both to this OperationInfo.

data/shared/src/main/scala/sigma/ast/SigmaPredef.scala
```diff
@@ -529,7 +529,159 @@ object SigmaPredef {
...
+        """.stripMargin,
+        Seq(ArgInfo("inBox", "the input box"), ArgInfo("outBox", "the output box"),
+            ArgInfo("token", "the token id being spent"), ArgInfo("amount", "the spent amount")))
+    )
+
     val globalFuncs: Map[String, PredefinedFunc] = Seq(
+      VerifySameForBasicRequiredRegistersFunc,
```
jozanek: Nit: append the six new entries after the existing ones instead of prepending.

sc/shared/src/test/scala/sigmastate/utxo/UtilFunctionsSpecification.scala
```diff
@@ -0,0 +1,204 @@
...
+      outTokens = Seq(token1 -> 60L))
+  }
+
+  property("verifySpentToken - another token not preserved") {
+    testEval(s"""sigmaProp(verifySpentToken(SELF, OUTPUTS(0), fromBase16("${tokenHex(1)}"), 40L) == false)""",
+      selfTokens = Seq(token1 -> 100L, token2 -> 7L),
+      outTokens = Seq(token1 -> 60L, token2 -> 6L))
+  }
+
+  property("verifySpentToken - token fully spent is not supported") {
```
jozanek, line 199: Add a pinning test next to this one: `verifySpentToken` when `inBox` holds no `token` at all — currently evaluates to `true` (vacuous forall). That semantics should be asserted, not implicit.

jozanek (changes requested):
**Name reservation / shadowing.** `PredefinedFuncApply` post-processes every typed `Apply(Ident(name), _)` (`SigmaTyper.scala:350`), so these six names silently capture same-named, same-arity user functions on recompile — and per #1037 these exact names already exist in user scripts (the tupled 1-arg forms escape only by arity). Options: (a) skip expansion when the name is bound in the typer env, (b) make redefinition a compile error, (c) accept and list the reserved names in release notes. Which do we want? The mechanism predates this PR (`allOf` etc.), but this PR grows the surface with plausible user-space names.

**LangSpec.** `docs/LangSpec.md` (Predefined functions section) needs entries for the six functions — required by `docs/review/checklist.md` for user-visible ErgoScript changes.

**Spec appendix.** The `.tex` appendices are generated from OperationInfo via `GenPredefTypesApp` — should be re-run; note `\lst{R{i}[Any].isEmpty}` has literal braces that may break LaTeX.

**Naming (maintainer sign-off).** These names are frozen forever once released. `verifySameForBasicRequiredRegisters` checks value+propositionBytes; the `verify` prefix and length are debatable. Fine to keep #1037's names, but it should be a deliberate call.

**Question — version gating.** I assume no gate is intended since all emitted nodes are pre-existing (unlike the tree-version-gated `BareNone` frontend feature). Confirming that's deliberate, so newly compiled v0 trees using these helpers are expected.

Ergologica:
Thanks for the thorough review — everything below is in the three new commits (4368fd3, 27aef5f, 53819f3).

**Name reservation / shadowing.** I checked what the typer actually does today rather than what `PredefinedFuncApply` alone suggests: `SigmaTyper.typecheck` seeds the environment with the predefined function names (`predefinedEnv`), and the `Block` case rejects any `val`/`def` whose name is already bound (`Variable X already defined`). So a user script that declares `def verifySpentToken(...)` — with the same signature, a different arity or as a `val` — fails at typing before any expansion can silently replace it. In other words option (b) is already the status quo for every predefined name, these six included; the only binding the typer does not reject is a lambda parameter, and with the explicit type guards (below) a parameter can now be captured only if it has exactly the predefined signature. I added three pinning tests in `UtilFunctionsSpecification` (same-signature `def`, any-signature `def` for all six names, `val` rebinding) so this stays a deliberate choice, and a short note in `LangSpec.md` saying that predefined names are reserved (your option (c)), which also covers the scripts mentioned in #1037: they will get a clear compile error rather than a behaviour change. If you'd rather have the typer skip the expansion when the name is user-bound (option (a)), I can do that too — it would be a change in `SigmaTyper` affecting all predefined functions, so I preferred not to widen the scope without a maintainer's call.

**LangSpec.** Added the six declarations, with their docs, at the end of the *Predefined functions* section, plus the reserved-names note. I could not find `docs/review/checklist.md` on `master` or `v6.0.7-candidate` — if it lives in a pending PR, point me to it and I'll follow it.

**Spec appendix.** Two findings. First, `SpecGen.predefFuncs` filters out functions whose `opDesc` is `MethodCall`/`PropertyCall`/`Constant` (`noFuncs`), and the six helpers are declared with `OperationInfo(MethodCall, ...)` because they expand to a mix of existing nodes: they never enter `predeffunc_rows.tex` / `predeffunc_sections.tex`. Second, `GenPredefFuncsApp` currently does not run on `master` at all: `SpecGen.collectOpsTable` asserts one function per opcode and `executeFromSelfReg` / `executeFromSelfRegWithDefault` both map to `DeserializeRegister` (opcode 213), independently of this PR. With that assertion relaxed locally the generator produces no entry for the six functions (0 occurrences) and only a pre-existing 3-line drift in `predeffunc_sections.tex`, which I did not commit. I rewrote the `R{i}` wording anyway, in the OperationInfo and in the test comment.

**Naming.** Kept #1037's names, deliberately: they are the names already in use in the scripts that motivated the issue, and the `verify` prefix reads as a predicate. If @kushti prefers shorter names (e.g. `sameValueAndScript`, `hasToken`) the rename is mechanical — happy to do it before merge, since the names are frozen once released.

**Version gating.** Confirmed, no gate is intended. The expansion only emits nodes that exist since ErgoTree v0 (`BinAnd`/`BinOr`, `EQ`/`NEQ`/`GE`, `ExtractAmount`, `ExtractScriptBytes`, `ExtractRegisterAs`, `OptionIsDefined`, `LogicalNot`, `MethodCall(Box.tokens)`, `Exists`/`ForAll`, `If`, `Plus`, `SelectField`), so newly compiled v0 trees using the helpers are expected and valid; there is no ErgoTree-level or consensus change (the whole spec runs under `CompilerCrossVersionProps`, i.e. across tree versions).

**`@unchecked` patterns.** Added explicit `tpe` guards to all six builders (`inBox.tpe == SBox && outBox.tpe == SBox`, `token.tpe == SByteArray`, `used.tpe == SInt`, `amount.tpe == SLong`), so a non-matching application falls through to the normal type error.

**`boxTokens`.** Moved the helper (and `STokenElemType`) above the first function and reused it in `verifySameForRequiredRegisters` instead of the two inline `mkMethodCall`s.

**`ErgoBox.{R4..R9}`.** Imported; the clause list is now `1 -> R4, 2 -> R5, ..., 6 -> R9`.

**Checked `Plus`.** Documented in the OperationInfo of `verifySpentToken`: if `outAmount + amount` overflows `Long` the evaluation throws instead of returning `false`.

**`verifySpentToken` doc caveats.** Both added to the OperationInfo (and to LangSpec): `inBox` without the token makes the `forall` vacuously true, and a full spend (token absent in `outBox`) returns `false`.

**`globalFuncs` order.** The six entries are now appended after `FromBigEndianBytesFunc`.

**Pinning test.** Added `verifySpentToken - inBox without the token is vacuously true` (plus the variant with `inBox` holding no tokens at all), next to the full-spend test.

Locally: `UtilFunctionsSpecification` 25/25, `sigmastate.lang.*` + `sigma.compiler.*` 128/128 (Scala 2.13). On CI the three JVM matrices are green; the JS job of the intermediate commit failed in 21 s at the sbt launcher download from Maven Central (checksum file not served — a rate-limit hiccup unrelated to the change), the run on the final commit is past that step. The PR is still based on `master`; happy to retarget to `v6.0.7-candidate` if you prefer.

jozanek (approved):
All findings addressed — verified against the three new commits, not just the reply.

**Shadowing.** You're right, my claim was overstated: `typecheck` seeds `assignType(predefinedEnv, bound)` and the `Block` case rejects rebinding with `Variable ... already defined` (`SigmaTyper.scala:86-88, 691`) — a loud compile error, never a silent hijack. Option (b) is the status quo; the tpe guards close the lambda-parameter corner, and the pinning tests + LangSpec note make it explicit. No need for option (a).

**checklist.md.** My mistake — it came from a review playbook not part of this repository; ignore it. The LangSpec update stands on its own and is done.

**Spec appendix.** Confirmed: `SpecGen.predefFuncs` filters `noFuncs = Set(Constant, MethodCall, PropertyCall)` (`SpecGen.scala:59-61`), so nothing to regenerate here. Please file the pre-existing `GenPredefFuncsApp` breakage (opcode 213 collision) as a separate issue so it isn't lost.

**Non-blocking:** naming sign-off stays with @kushti — names are frozen once released.

PR #1177, opened 2026-06-11
Add example of Bitcoin transaction parsing in ErgoScript (closes #1114)

### LINKED ISSUE #1114: Example of a Bitcoin transaction parsing in ErgoScript
Consider following contract: 
SELF box has transaction id in register R4 and Bitcoin UTXO script bytes hash in register R5. 

Script allows for the box to be spent if a Bitcoin transaction which bytes are passed in a context extension variable has the requested id and has output with script hash from R5.

Ergologica:
# Example: parsing a Bitcoin transaction in ErgoScript

Closes #1114

## Summary

This PR adds `BitcoinTxParsingExampleSpecification`, an example contract (with a test suite) implementing the scenario described in #1114:

> SELF box has transaction id in register R4 and Bitcoin UTXO script bytes hash in R5. The script allows the box to be spent if a Bitcoin transaction whose bytes are passed in a context extension variable has the requested id and has an output with the script hash from R5.

## The contract

The ErgoScript contract (see `contractScript` in the spec):

- checks the **txid**: `sha256(sha256(txBytes)) == SELF.R4[Coll[Byte]].get`. The txid is kept in internal byte order (not reversed as displayed by explorers). Since the txid commits to all transaction bytes, this check also authenticates everything parsed afterwards.
- **fully parses the legacy Bitcoin transaction format** (`version | nInputs | inputs | nOutputs | outputs | lockTime`): reads varints, skips the inputs (each `prevTxId(32) | vout(4) | scriptSig(varint-prefixed) | sequence(4)`), then scans the outputs (each `value(8) | scriptPubKey(varint-prefixed)`) checking whether `sha256(scriptPubKey) == SELF.R5[Coll[Byte]].get` for some output.
- requires the parser to consume the transaction exactly up to the trailing 4-byte `lockTime`.

Since ErgoScript has no unbounded loops, inputs and outputs are traversed with `fold` over a fixed range, so the contract supports up to 10 inputs and 10 outputs (the counts are checked explicitly, and the bounds are easy to enlarge). Bitcoin var-ints are supported in the 1-byte and `0xfd` (3-byte) encodings, which covers any realistic transaction within these bounds.

For segwit transactions the txid is defined over the witness-stripped serialization, so they should be passed to the contract in exactly that (legacy) serialization; this is documented in the example.

## Tests

10 properties, run across script versions via `CompilerCrossVersionProps`:

- **real transaction**: the first ever Bitcoin payment (block 170, Satoshi → Hal Finney, txid `f4184fc5…9e16`, 1 input, 2 P2PK outputs) — spendable with either output's script hash; negative cases for a wrong expected script hash, a hash of a transaction fragment which is not an output script (the input's scriptSig), a wrong txid and tampered transaction bytes;
- **synthetic transactions** (built by a small legacy-format serializer in the test): multi-input/multi-output, a 300-byte script exercising the `0xfd` varint path (with a 260-byte scriptSig on the input side as well), the maximum supported 10 inputs / 10 outputs, and rejection of a transaction exceeding the input bound.

The example only adds a test-suite file; no production code is touched.


kushti (approved):

kushti:
@Ergologica please reach me to collect the bounty!
mail : kushti@protonmail.ch
tg: @kushti_ru

PR #1160, opened 2026-05-28
Fix select error messages and selector source position

jozanek:
Closes #409

PR #1178, opened 2026-06-24
Candidate for 6.0.6 release

kushti:
* #1160
* #1173 
* #1177 
* #1180 
* #1185


odiseusme (changes requested):
Blocking: this breaks JVM binary compatibility with v6.0.5.

CompilerException(String, Option[SourceContext]) existed in v6.0.5, but this PR replaces it with the 3-arg constructor including cause.

I reproduced it with an already-compiled client:

v6.0.5 -> works
PR #1178 -> java.lang.NoSuchMethodError:
CompilerException.<init>(String, scala.Option)

So the default arg preserves source compatibility, but not the old JVM ABI.

Minimal fix: keep the old constructor(s) as forwarding constructors to the new cause-aware ones. Same applies to the other affected public compiler exceptions.

jozanek (commented):
Verified the two `data/` main-source changes are not consensus-observable: `ArgInfo` is doc-metadata only and never shapes serialized bytes, the six new `InfoObject`s have no code references, and Context method ids 4–10 are identical in both v5 and v6 method buckets, so `Operations` init is version-independent. The `Box_methods.tex` `getReg` → `getRegV5` change is faithful regeneration. The golden-test suite locking generated files to their generators is a nice structural improvement.

Two review-level notes that have no inline anchor:

- **MINOR (PR description):** The body is just four PR links. Since this release merge touches `data/` main sources, one paragraph answering the red-zone questions would make it self-reviewing — e.g.: "The `Operations.scala` additions are doc metadata (`ArgInfo`) never read by serializers, with method ids 4–10 present identically in the v5 and v6 method sets; `CompilerExceptions` changes are compiler-only and unreachable from the interpreter."
- **NIT (PR title):** Typo: "Canidate" → "Candidate".

**Approve with changes** — 0 BLOCKER, 0 MAJOR, 4 MINOR, 3 NIT.

data/shared/src/main/scala/sigma/exceptions/CompilerExceptions.scala
```diff
@@ -7,9 +7,13 @@ import sigma.ast.SourceContext
   *
   * @param message the error message
   * @param source an optional source context with location information
+  * @param cause an optional underlying cause for the exception
   */
-class CompilerException(message: String, val source: Option[SourceContext] = None)
-    extends SigmaException(message, None) {
+class CompilerException(
```
jozanek, line 12: **MINOR:** Adding the `cause` parameter changes the public constructor signatures of `CompilerException` and all four subclasses. Source-compatible thanks to the defaults, but binary-incompatible for downstream consumers (Appkit, ergo-wallet, tooling) that construct or subclass these exceptions. Per the precedent in #883 this is acceptable, but it should be called out in the 6.0.6 release notes rather than discovered by consumers at runtime.

sc/shared/src/main/scala/sigma/compiler/SigmaCompiler.scala
```diff
@@ -155,6 +162,66 @@ object SigmaCompiler {
...
+      "Script compilation failed (stack overflow): script is too complex or recursive",
+      source,
+      Some(cause))
+  }
+
+  /** Runs the given computation and converts a stack overflow into a recoverable
+    * [[CompilerException]] while rethrowing all unrelated throwables (including
+    * [[OutOfMemoryError]]) unchanged.
+    */
+  private def checkStackOverflow[A](source: Option[SourceContext])(body: => A): A = {
```
jozanek, line 217: **MINOR (question):** `checkStackOverflow` also wraps `compileTyped`, so a `StackOverflowError` can be caught mid-`IR.buildGraph` inside a mutable, typically reused `IRContext`. The new "compiler is still usable after a stack overflow" test only proves recovery for the typecheck/binder phase. Is `IRContext` state provably safe to reuse after an aborted `buildGraph`? If a half-built graph can survive and later produce a *wrong* tree instead of an error, that's worse than the fatal error this change removes. Suggest either a test that reuses the same `IR` after a caught SOE in `compileTyped`, or a doc note saying the `IRContext` must be discarded after this exception.

sc/shared/src/test/scala/sigmastate/utxo/examples/BitcoinTxParsingAmountBindingSpecification.scala
```diff
@@ -0,0 +1,493 @@
...
+        Map(1.toByte -> ByteArrayConstant(txBytes.toColl))
+    }
+    val res = prover.prove(emptyEnv, tree, ctx, fakeMessage)
+    if (expectSuccess) {
+      val pr = res.getOrThrow
+      val verifier = new ErgoLikeTestInterpreter
+      verifier.verify(emptyEnv, tree, ctx.withExtension(pr.extension), pr.proof, fakeMessage)
+        .getOrThrow._1 shouldBe true
+    } else {
+      res.isFailure shouldBe true
```
jozanek: **MINOR:** The negative cases assert only `res.isFailure`, so a test named "rejects when amount is below R6 threshold" would also pass if the script threw `IndexOutOfBoundsException` or failed for an unrelated reason — the classic "passes for the wrong reason" pattern. Where the guard under test is a boolean check, consider asserting the failure is reduce-to-false (or use `assertExceptionThrown` with the expected message). Same pattern applies to the `spend` helper in `BitcoinTxParsingExampleSpecification.scala`.

sc/shared/src/main/scala/sigma/compiler/SigmaCompiler.scala
```diff
@@ -155,6 +162,66 @@ object SigmaCompiler {
...
+    } catch {
+      // Convert a stack overflow (JVM StackOverflowError or JS RangeError) into a checked
+      // CompilerException so that a malformed/pathological script results in a normal
+      // compilation error instead of a fatal runtime error.
+      case t: Throwable if isStackOverflow(t) =>
+        throw compilerStackOverflowException(source, t)
+      // Everything else (OutOfMemoryError, InternalError, UnknownError,
+      // InterruptedException, ordinary exceptions, etc.) is unrelated to recursion depth
+      // and must propagate unchanged.
+      case t: Throwable => throw t
```
jozanek: **NIT:** The `case t: Throwable => throw t` clause is redundant — an unmatched throwable propagates anyway. The clause can be dropped; the explanatory comment about OOM/InternalError can stay above the guarded case.

sc/shared/src/main/scala/sigma/compiler/SigmaCompiler.scala
```diff
@@ -155,6 +162,66 @@ object SigmaCompiler {
...
+    * reference JS-only types, the JS case is detected by inspecting the exception class name
+    * and message.
+    *
+    * The overload taking `isJVM` is package-private so tests can assert both platform
+    * predicates without running on the actual platform.
+    */
+  private[sigma] def isStackOverflow(t: Throwable): Boolean =
+    isStackOverflow(t, Environment.current.isJVM)
+
+  private[sigma] def isStackOverflow(t: Throwable, isJVM: Boolean): Boolean = {
```
jozanek, line 185: **NIT:** The JS predicate matches engine-specific message strings ("call stack size exceeded" for V8/JSC, "too much recursion" for SpiderMonkey). An engine with different wording falls through to propagating the raw error — which is the fail-safe direction, so this is fine, but worth a comment stating the fallback is intentional.

sc/shared/src/main/scala/sigma/compiler/SigmaCompiler.scala
kushti: fixed

sc/shared/src/main/scala/sigma/compiler/SigmaCompiler.scala
kushti: ScalaDoc updated with a note

data/shared/src/main/scala/sigma/exceptions/CompilerExceptions.scala
kushti: old signatures added back

sc/shared/src/test/scala/sigmastate/utxo/examples/BitcoinTxParsingAmountBindingSpecification.scala
kushti: fixed

sc/shared/src/main/scala/sigma/compiler/SigmaCompiler.scala
kushti: comment added

PR #1179, opened 2026-06-24
Post 6.0 refactoring pt #1: code cleanup

kushti:


sc/shared/src/test/scala/sigma/SigmaDslTesting.scala
```diff
@@ -816,11 +805,11 @@ class SigmaDslTesting extends AnyPropSpec
...
             case (Success(s), Success(old)) =>
               s shouldBe old
-            case (Failure(e), oldRes @ Success(_)) =>
+            case (Failure(_), oldRes @ Success(_)) =>
               // allow this because of allowNewToSucceed
               oldRes shouldBe expRes
-            case (Success(s), Failure(e)) =>
-              fail(s"Old version fail while scalaFunc succeeds: $s <+++> $e")
+            case (Success(s), Failure(_)) =>
+              fail(s"Old version fail while scalaFunc succeeds: $s")
```
jozanek, line 812: The removed `$e` was the only diagnostic for *why* the old version failed when this assertion trips.
Please keep the binder and restore it: `case (Success(s), Failure(e)) => fail(s"Old version fail while scalaFunc succeeds: $s <+++> $e")`.

sc/shared/src/main/scala/sigma/compiler/phases/SigmaTyper.scala
```diff
@@ -288,7 +288,7 @@ class SigmaTyper(val builder: SigmaBuilder,
           if (args.length != argTypes.length)
             error(s"Invalid argument type of application $app: invalid number of arguments", app.sourceContext)
           val typedArgs = args.zip(argTypes).map {
-            case (arg, expectedType) => assignType(env, arg)
+            case (arg, _) => assignType(env, arg)
```
jozanek, line 291: FYI: the discarded `expectedType` sat next to `assignType(env, arg)`, which has an `expected: Option[SType] = None` param — the binder likely marked never-wired-up intent to propagate the expected type.
Keeping behavior unchanged is right; consider filing a tracking issue so the hint isn't lost with this cleanup.

sc/shared/src/test/scala/sigma/SigmaDslTesting.scala
```diff
@@ -1400,8 +1389,10 @@ class SigmaDslTesting extends AnyPropSpec
   }
 
   /** Default implementation of [[Sampled]]. */
-  case class SampledData[A](samples: Seq[A])(implicit val arbitrary: Arbitrary[A])
-      extends Sampled[A]
+  case class SampledData[A](samples: Seq[A])(arb: Arbitrary[A])
```
jozanek, line 1392: Nit: behavior-equivalent (single construction site, case-class equality uses only the first param list), just slightly more verbose than the `implicit val` form.
Fine as is — no action needed if implicit elimination in test code is the goal.

jozanek (commented):
**Approve, modulo one small required fix** (the dropped exception in the `SigmaDslTesting` fail message — see inline).

**Question:** the `// required for Scala 2.11` comments were removed while the imports were kept — intentional, given 2.11 is still in CI? (Build stays green, so fine, but the "why" for those imports is now undocumented.)

PR #1183, opened 2026-07-10
Add gold insurance contract example (closes #1181)

### LINKED ISSUE #1181: Example of "Physical or digital gold: simple insurance on Ergo"
Implement contract in "[Physical or digital gold: simple insurance on Ergo](https://www.ergoforum.org/t/physical-or-digital-gold-simple-insurance-on-ergo/4715)"

smallboxlabs:
Implements the contract from the forum post
["Physical or digital gold: simple insurance on Ergo"](https://www.ergoforum.org/t/physical-or-digital-gold-simple-insurance-on-ergo/4715)
as a self-contained example spec (closes #1181).

The insurance box locks the vault's digital gold (gold-pegged tokens) with
`R4` = insurance deadline (height) and `R5` = the id of the unique insurance NFT
issued to the client, and allows two spending paths:

- **redemption, before (and at) the deadline** — any transaction presenting the
  insurance NFT in one of its inputs may send the digital gold to any address
  ("either party can present the NFT": the path is keyless, ownership of the NFT's
  own box is the real gate, and the NFT changes hands on successful delivery);
- **reclaim, strictly after the deadline** — the vault takes the collateral back
  with its own key, no NFT required.

The redemption window closes exactly at the deadline, so the paths never overlap:
an expired box cannot be raced by the NFT holder, and before expiry the vault cannot
pull the collateral out from under the client. Outputs are deliberately
unconstrained — the NFT-in-inputs check is the entire gate.

Properties cover both happy paths, the deadline boundary from both sides, and the
negative cases: no NFT, reclaim too early, NFT presented after expiry (window
closed, vault path intact), a different token posing as the NFT, and a stranger
after expiry. Tested green on Scala 2.13 and 2.12 locally
(`scJVM/testOnly sigmastate.utxo.examples.GoldInsuranceExampleSpecification`).

### DIFF
--- a/sc/shared/src/test/scala/sigmastate/utxo/examples/GoldInsuranceExampleSpecification.scala
+++ b/sc/shared/src/test/scala/sigmastate/utxo/examples/GoldInsuranceExampleSpecification.scala
@@ -1,0 +1,148 @@
+package sigmastate.utxo.examples
+
+import org.ergoplatform.ErgoBox
+import sigma.Extensions.ArrayOps
+import sigma.ast.{ByteArrayConstant, IntConstant}
+import sigma.ast.syntax._
+import sigma.data.AvlTreeData
+import sigmastate.CompilerCrossVersionProps
+import sigmastate.eval.Extensions.ArrayByteOps
+import sigmastate.helpers.TestingHelpers._
+import sigmastate.helpers.{CompilerTestingCommons, ContextEnrichingTestProvingInterpreter, ErgoLikeContextTesting, ErgoLikeTestInterpreter}
+
+/** Example of a simple insurance contract for a physical gold purchase
+  * (see https://github.com/ergoplatform/sigmastate-interpreter/issues/1181 and
+  * https://www.ergoforum.org/t/physical-or-digital-gold-simple-insurance-on-ergo/4715).
+  *
+  * A client buys a physical gold brick from a vault, to be delivered within some period,
+  * and pays an insurance premium on top. The vault then:
+  *   - issues a unique insurance NFT to the client, and
+  *   - locks an equivalent amount of digital gold (gold-pegged tokens) in a box guarded
+  *     by the contract below.
+  *
+  * The insurance box can be spent in two ways:
+  *   - redemption, before (and at) the insurance deadline: any transaction which presents
+  *     the insurance NFT in one of its inputs may send the digital gold to any address.
+  *     Normally that is the client claiming compensation for a failed delivery, but since
+  *     the NFT itself is the claim ticket, the same path serves the vault when the client
+  *     surrenders the NFT on successful delivery ("either party can present the NFT");
+  *   - reclaim, strictly after the deadline: the vault takes the digital gold back with
+  *     its own key, no NFT required.
+  *
+  * The redemption window closes exactly at the deadline, so the two spending paths never
+  * overlap: an expired insurance box cannot be raced by the NFT holder, and before the
+  * expiry the vault cannot pull the collateral out from under the client.
+  *
+  * The contract constrains only the inputs of the spending transaction; where the digital
+  * gold goes is deliberately left free.
+  */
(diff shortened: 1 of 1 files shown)


jozanek (commented):
**Questions / unknowns:**
1. The redemption window *includes* the deadline height (`HEIGHT <= deadline`), and reclaim starts strictly after. Is that inclusive boundary what the forum post intends, or just a convention choice here? Either is fine for an example, but worth stating deliberately since people will copy it.
2. `tokenBox` uses `TrueTree` for the NFT holder's box — intentional per its scaladoc, just confirming you considered that the example therefore never demonstrates the NFT owner's signature being what fixes the outputs.
3. FYI: in the redemption tests the prover being `client` is cosmetic — the path reduces to a trivially-true proposition, provable by anyone. The code comment already says the path is keyless, so no action needed.

sc/shared/src/test/scala/sigmastate/utxo/examples/GoldInsuranceExampleSpecification.scala
```diff
@@ -0,0 +1,148 @@
...
+      self = inputs.head,
+      activatedVersionInTests)
+
+    val res = prover.prove(env, tree, ctx, fakeMessage)
+    if (expectSuccess) {
+      val proof = res.get
+      val verifier = new ErgoLikeTestInterpreter
+      verifier.verify(env, tree, ctx, proof, fakeMessage).get._1 shouldBe true
+    } else {
+      res.isFailure shouldBe true
```
jozanek, line 113: **Consider:** this passes for *any* prover failure, not just "proposition reduced to false" — an unrelated evaluation error would go unnoticed. Per docs/review/testing.md, use `assertExceptionThrown`/`exceptionLike` to also assert the failure message.

sc/shared/src/test/scala/sigmastate/utxo/examples/GoldInsuranceExampleSpecification.scala
```diff
@@ -0,0 +1,148 @@
...
+  private val env = Map("vault" -> vaultPubKey)
+
+  private val insuranceScript =
+    """{
+      |  val deadline = SELF.R4[Int].get
+      |  val insuranceNftId = SELF.R5[Coll[Byte]].get
+      |
+      |  // the insurance NFT is presented if any input of the spending transaction
+      |  // carries the token issued for this insurance deal
+      |  val nftPresented = INPUTS.exists({ (b: Box) =>
```
jozanek, line 58: **Consider:** `INPUTS` includes `SELF`, so if a deployer sets `R5` to a token id the insurance box itself carries (e.g. the digital-gold id), redemption becomes trivially satisfiable by anyone. Add one scaladoc line: "the NFT id must not be a token the insurance box itself holds."

sc/shared/src/test/scala/sigmastate/utxo/examples/GoldInsuranceExampleSpecification.scala
```diff
@@ -0,0 +1,148 @@
...
+    } else {
+      res.isFailure shouldBe true
+    }
+  }
+
+  property("insurance NFT unlocks the digital gold before the deadline") {
+    trySpend(client, insuranceDeadline - 1, Some(tokenBox(insuranceNftId)), expectSuccess = true)
+    // the redemption window includes the deadline itself
+    trySpend(client, insuranceDeadline, Some(tokenBox(insuranceNftId)), expectSuccess = true)
+  }
```
jozanek, line 121: **Optional:** the forum's "successful delivery" flow is the *vault* presenting the surrendered NFT before the deadline — currently only tested post-deadline (line 137). Add `trySpend(vault, insuranceDeadline - 1, Some(tokenBox(insuranceNftId)), expectSuccess = true)` to pin it.

jozanek (changes requested):

PR #1185, opened 2026-07-15
Wrap non-OOM VirtualMachineError into a checked CompilerException during typecheck in compiler

kushti:
In this PR, during typecheck, StackOverflowError (and other non-OOM VirtualMachineErrors) is wrapped into a checked CompilerException so that a malformed/pathological script results in a normal compilation error instead of a fatal JVM error.

### DIFF
--- a/data/shared/src/main/scala/sigma/exceptions/CompilerExceptions.scala
+++ b/data/shared/src/main/scala/sigma/exceptions/CompilerExceptions.scala
@@ -7,9 +7,13 @@ import sigma.ast.SourceContext
   *
   * @param message the error message
   * @param source an optional source context with location information
+  * @param cause an optional underlying cause for the exception
   */
-class CompilerException(message: String, val source: Option[SourceContext] = None)
-    extends SigmaException(message, None) {
+class CompilerException(
+    message: String,
+    val source: Option[SourceContext] = None,
+    cause: Option[Throwable] = None
+) extends SigmaException(message, cause) {
 
   override def getMessage: String = source.map { srcCtx =>
     val lineNumberStrPrefix = s"line ${srcCtx.line}: "
@@ -22,28 +26,49 @@ class CompilerException(message: String, val source: Option[SourceContext] = Non
   *
   * @param message the error message
   * @param source an optional source context with location information
+  * @param cause an optional underlying cause for the exception
   */
-class BinderException(message: String, source: Option[SourceContext] = None)
-    extends CompilerException(message, source)
+class BinderException(
+    message: String,
+    source: Option[SourceContext] = None,
+    cause: Option[Throwable] = None
+) extends CompilerException(message, source, cause)
 
 /** Exception thrown during the type checking phase of the compiler.
   *
   * @param message the error message
   * @param source an optional source context with location information
+  * @param cause an optional underlying cause for the exception
   */
-class TyperException(message: String, source: Option[SourceContext] = None)
-    extends CompilerException(message, source)
+class TyperException(
+    message: String,
+    source: Option[SourceContext] = None,
+    cause: Option[Throwable] = None
+) extends CompilerException(message, source, cause)
 
 /** Exception thrown during the building phase of the compiler.
   *
   * @param message the error message
   * @param source an optional source context with location information
+  * @param cause an optional underlying cause for the exception
   */
-class BuilderException(message: String, source: Option[SourceContext] = None)
-  extends CompilerException(message, source)
+class BuilderException(
+    message: String,
+    source: Option[SourceContext] = None,
+    cause: Option[Throwable] = None
+) extends CompilerException(message, source, cause)
 
-class GraphBuildingException(message: String, source: Option[SourceContext])
-    extends CompilerException(message, source)
+/** Exception thrown during graph building.
+  *
+  * @param message the error message
+  * @param source an optional source context with location information
+  * @param cause an optional underlying cause for the exception
+  */
+class GraphBuildingException(
+    message: String,
+    source: Option[SourceContext] = None,
+    cause: Option[Throwable] = None
+) extends CompilerException(message, source, cause)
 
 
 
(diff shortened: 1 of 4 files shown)


odiseusme:
Reviewed with the upcoming SigmaSDK release in mind. For the reported binder-loop failure, the containment point is appropriate, and the regression test correctly pins the cyclic environment binding that overflows in the binder's reduction/rewriting path.

I would ship this after two small predicate changes.

### 1. JVM predicate is broader than its name and Scaladoc claim (must fix)

On the JVM, `isStackOverflow` currently matches:

```scala
t.isInstanceOf[VirtualMachineError] && !t.isInstanceOf[OutOfMemoryError]
```

Besides `StackOverflowError`, this also matches `InternalError` and `UnknownError`. Those represent VM-level failures, not necessarily pathological scripts. Converting them into an ordinary `CompilerException` with the message "script is too complex or recursive" would misclassify a genuine JVM or compiler defect and could defeat callers' fail-fast handling.

Suggested change:

```scala
if (Environment.current.isJVM)
  t.isInstanceOf[StackOverflowError]
```

This also makes the implementation match the method name and Scaladoc: "represents a stack overflow."

### 2. The Scala.js branch contains a false-positive clause that adds no useful coverage (must fix)

The message checks end with:

```scala
message.contains("RangeError")
```

Because `js.JavaScriptException.getMessage` delegates to the wrapped native value's `toString`, this matches any native `RangeError`, including unrelated failures such as:

```text
RangeError: Invalid array length
```

Such an error could indicate a compiler bug, yet it would be relabeled as script complexity or recursion.

The generic `"RangeError"` alternative does not appear necessary for stack-overflow coverage:

* V8 and JavaScriptCore use "Maximum call stack size exceeded," already matched by `"call stack size exceeded"`.
* Firefox uses "too much recursion," already matched separately.

The list can therefore be reduced to:

```scala
message.contains("call stack size exceeded") ||
message.contains("too much recursion")
```

I also confirmed that the shared regression test runs under Scala.js in CI through `scJS/test` on Node 20. Since the test creates unbounded binder recursion, it genuinely exercises the JavaScript stack-overflow detection path.

### 3. The original throwable is discarded (should fix)

`CompilerException` currently extends `SigmaException(message, None)` and does not accept a cause. The catch site therefore cannot preserve the original `StackOverflowError` through the normal constructor.

Discarding it loses the stack trace needed to determine whether the overflow occurred in the binder, typer, builder, or an unrelated helper. That distinction matters when deciding whether an input was pathological or the compiler itself regressed.

Possible fixes:

* Add an optional cause parameter to `CompilerException`, defaulted to `None` to avoid call-site churn.
* Construct the exception, call `initCause(t)`, and then throw it.

### 4. Add negative-direction tests (should fix)

After narrowing the predicates, tests should lock down what must not be converted:

* `InternalError` is not wrapped.
* `UnknownError` is not wrapped.
* On Scala.js, a non-stack native `RangeError` is not wrapped.

One additional recovery test would be valuable: after catching the overflow, reuse the same `SigmaCompiler` instance to compile a normal expression. Returning a recoverable exception implicitly assumes that no persistent compiler or builder state was left unusable by the aborted call.

### 5. Scope note (not a blocker)

Only `typecheck` is guarded:

* The string overload calls `parse` before entering the protected method.
* `compileTyped` performs `buildGraph` and `buildTree` afterward, outside the catch.

The expression grammar is recursively nested, so sufficiently deep source may still overflow during parsing and escape as a fatal error. I have not established the actual reachable threshold or whether surrounding request-size limits prevent it.

If this PR is specifically intended to contain the known binder-loop failure, the present boundary is reasonable. If the release notes will describe it more broadly as making compilation of untrusted source safe, the protection would need to cover the other recursive phases or, preferably, those phases would need explicit depth or complexity limits.

Which scope is intended?

### Follow-up issue

For this exact reproducer, the underlying problem is that the binder repeatedly expands a cyclic environment binding until the physical call stack is exhausted.

Cycle detection during environment-binding resolution (tracking identifiers currently being expanded and rejecting a revisit) would produce a deterministic compiler error before stack exhaustion. The overflow catch would then remain useful as defense in depth for unexpected recursive paths.

That improvement should be tracked separately rather than block this PR.

For completeness, this does not appear consensus-facing. Source compilation is off-chain; ErgoTree serialization and interpretation of already compiled trees are unchanged.

kushti:
@odiseusme comments addressed, please make another pass!

odiseusme:
Second pass on 3cd6db6. The fix commit is substantially stronger than the original: all five review points are addressed in the implementation, and the scope question got answered by expanding the boundary itself, with `checkStackOverflow` now wrapping `parse`, `typecheck`, and `compileTyped` (graph and tree building included).

Verified:

* JVM predicate narrowed to `StackOverflowError`. The negative tests prove `InternalError`/`UnknownError` fail the predicate; unchanged propagation then follows from the wrapper's fallthrough.
* JS message list reduced to the two engine-specific clauses, with a negative test for the non-stack `RangeError: Invalid array length` shape. That test is synthetic (a name-mimicking class exercising the string heuristic from shared code); the genuine end-to-end JS coverage is the cyclic-binder regression test, which overflows for real on Node via `scJS/test`. A small JS-only negative test with a native non-stack `RangeError` would close the remaining gap, but I don't consider it blocking.
* Cause preservation works: the original throwable is passed as `Some(cause)` and the regression test asserts `getCause` is populated.
* The reuse test covers recovery after a binder overflow on the same compiler instance. It doesn't exercise recovery after an overflow in the newly guarded graph-building path; fine as is, just noting the claim's scope.
* No double-wrapping (an inner `CompilerException` fails the predicate and passes through outer wrappers unchanged) and OOM still propagates via the fallthrough.

One remaining request before merge: the constructor change is binary incompatible. With the added defaulted `cause` parameter, the previously published two-argument constructors of `CompilerException`, `BinderException`, `TyperException`, `BuilderException`, and `GraphBuildingException` no longer exist in bytecode, so pre-compiled downstream code that constructs or subclasses them would break on a binary-only dependency bump. Catch sites are unaffected, and this repo doesn't run MiMa, so this may be acceptable; but preserving the old signatures costs two lines per class:

```scala
def this(message: String, source: Option[SourceContext]) = this(message, source, None)
```

Either add the forwarding constructors or make an explicit call that the ABI change is fine for this release; your decision. With that resolved, LGTM for the SigmaSDK release.

Follow-ups remain non-blocking as before: binder cycle detection for the cyclic-env reproducer, and more neutral message wording (a stack overflow during compilation is proven; "script is too complex" is a diagnosis the compiler can't always know).

PR #1190, opened 2026-09-07
Cross-compile all modules to Scala 3 on JVM and Scala.js

### LINKED ISSUE #947: Add cross-compilation to Scala3
The following cross-compilation targets should be supported:

    JVM: Scala 2.11, 2.12, 2.13, 3.3.x
    JS: Scala 2.13, 3.3.x (via latest Scala.js)

The dependencies should be consistent across ScorexFoundation projects. See related PRs (https://github.com/ScorexFoundation/scorex-util/pull/30, https://github.com/ScorexFoundation/debox/issues/5, https://github.com/input-output-hk/scrypto/issues/105) for concrete versions of deps that can be used, example configurations and typical code fixes.

Can start from enabling [migration mode](https://docs.scala-lang.org/scala3/guides/migration/tooling-migration-mode.html#automatic-rewrites).

infinityscroll:
Closes #947.

Extends the core migration in #1175 to all six modules, Scala.js, and the combined JVM artifact. The original four commits from #1175 are retained in the branch history.

Adds Scala 3.3.8 with migration mode and Scala.js 1.21.0, while retaining JVM Scala 2.11/2.12/2.13 and JS Scala 2.13. Dependencies and CI cover the supported version/platform combinations.

The review updates replace the Scala 3 cost and evaluation-mode wrappers with opaque types and remove the Scala 2 supertagged fallback from Scala 3. Scala 2 retains its existing types. Versioned test execution and compiler registration now share their implementation, with thin adapters for the differing ScalaTest APIs. Regression tests cover arithmetic, type separation, fixture behavior, caller positions, and printable cost expressions.

Local validation used JDK 17 and Node 22:

- All six modules and their test sources compile on all six supported targets. Full core/data/interpreter/parsers/SDK suites and focused compiler registration, ErgoTree, and serialization suites passed across that matrix.
- Both production Scala.js builds passed all 41 exported-API tests.
- Scala 3 binary, documentation, sources, test artifacts, and POM packaged successfully. A separate consumer compiled and ran against the packaged JAR without migration flags or supertagged; invalid typed values were rejected.

The full compiler runtime suites and Java 8 environment are covered by the PR CI matrix rather than the local review regression run.


data/shared/src/main/scala-3/sigma/ast/JitCost.scala
```diff
@@ -0,0 +1,38 @@
+package sigma.ast
+
+/** Represents cost estimation computed by JITC interpreter.
+  * The JITC costs use 10x more accurate scale comparing to block cost values.
+  *
+  * @see toBlockCost
+  */
+// TODO make 'value` private[sigma] after code moved from sigmastate package
+case class JitCost private[sigma](val value: Int) extends AnyVal {
```
jozanek: why are you still using extends AnyVal in scala 3?

data/shared/src/main/scala-3/sigma/eval/EvalSettings.scala
```diff
@@ -0,0 +1,91 @@
+package sigma.eval
+
+import sigma.eval.EvalSettings.EvaluationMode
+import supertagged.TaggedType
```
jozanek: Why do you still use supertagged in scala 3?

jozanek (changes requested):
Hey, thanks for PR! Anyway, I went over first 2 files and it looks like you didn't take advantag of scala 3 at all. Can you confirm you properly set your prompt to do that?

interpreter/shared/src/test/scala-3/sigmastate/CrossVersionProps.scala
```diff
@@ -0,0 +1,81 @@
...
+import sigma.VersionContext
+import sigmastate.eval.CProfiler
+
+import scala.collection.mutable
+import scala.util.DynamicVariable
+
+/** Scala 3's inline `property` is final, so registration stays with ScalaTest and
+  * the fixture hook wraps execution instead. This also retains each caller's source position.
+  */
+trait CrossVersionProps extends AnyPropSpecLike with TestsBase {
```
aslesarenko: Does this duplication really necessary? Have you tried to avoid it?

sc/shared/src/test/scala-3/sigmastate/CompilerCrossVersionProps.scala
```diff
@@ -0,0 +1,23 @@
...
+    if (okRunTestsWithoutMCLowering) {
+      val testName2 = s"${testName}_MCLowering"
+      // Retain the Scala 2 registration scope; changing the deferred test body's
+      // mode is a separate behavior change from cross-compilation.
+      _lowerMethodCalls.withValue(false) {
+        registerTest(testName2, testTags*)(testFun)(pos)
+      }
+    }
+  }
+}
```
aslesarenko, line 17: Duplication.

aslesarenko (changes requested):
looks legit, but the duplication looks avoidable

aslesarenko (approved):
LGTM, but since I don't maintain this code you better get another approval.

PR #1193, opened 2026-09-20
Make the graph IR's call node carry its ErgoTree descriptor; delete the staged wrappers

jozanek:
## What

The compiler's graph IR (`sc/shared/src/main/scala/sigma/compiler/ir`) used to represent a DSL
method call as a `MethodCall` node holding a reflective `RMethod`, built by generated "staged
wrapper" traits (`wrappers/sigma/{SigmaDslUnit,CollsUnit,WOptions}.scala` and their `impl/`
adapters). Every DSL method existed four times: the staged trait, the adapter body, an extractor
object used by TreeBuilding, and a `GraphIRReflection` registration so Scala.js could resolve the
method by name. Adding a method meant touching all four, and JVM and JS could disagree.

This PR makes the IR `MethodCall` node carry the descriptor the call was built from:

- `MethodCallee(method: SMethod)` for methods with an `SMethod` (identity `(objType, methodId)`);
- `OpCallee(op: ValueCompanion, opCode)` for ErgoTree operations without one (`SigmaAnd`,
  `CalcBlake2b256`, `ArithOp` with its op code, `ConcreteCollection`, ...).

GraphBuilding builds these nodes straight from the typed AST; rewrite rules match the descriptor
(`CallPattern`); TreeBuilding lowers a node either through a row in the new `Lowering` table
(keyed by callee, one row per dedicated ErgoTree node) or, by default, rebuilds a plain
`sigma.ast.MethodCall` from the descriptor. Nothing is looked up by name any more, and a method
with no dedicated ErgoTree node needs no code in `sc` at all.

Removed: `sigma.compiler.ir.wrappers.*` (all staged traits, adapters and constants),
`GraphIRReflection`, the reflective Elem cache (`cachedElemByClass` and the owner-kind plumbing),
the reflective invoke path of `MethodCall`, the JVM-only `ReflectionGenerator` and its test hooks.
The IR types of DSL values are now the runtime `sigma.*` types (`Ref[sigma.Box]`, `Elem[sigma.Coll[A]]`).

## Behaviour

None observable. The emitted ErgoTree (AST, including specialised `SMethod` types, and bytes) is
byte-identical for every script in `LanguageSpecificationV5`, `LanguageSpecificationV6`,
`ErgoTreeSpecification`, `BasicOpsSpecification` and `ErgoTreeBuildingTest` (538 tests, ~90k
snapshot lines compared per commit against the base). Failures and their messages are unchanged.
The suspected `typeSubst` loss in `MethodCall.mirror` (`MethodCallMirrorSpec`) turned out to be
unreachable from ErgoScript, so there is no behaviour change to report; the new node keeps the
substitution by construction.

Consensus-side modules (`core`, `data`, `interpreter`) are untouched. The graph IR is compile-time
only since v5.0.

Two notes for `sc` API users:

- `CompilerResult.compiledGraph` is now typed `Ctx#Ref[sigma.Context => Any]` instead of
  `Ctx#Ref[Ctx#Context => Any]` (the staged `Context` type no longer exists). The published JS API
  only uses `buildTree` and is unaffected.
- The per-method `isV3OrLaterErgoTreeVersion` guards inside GraphBuilding are gone: the typer's
  version-gated method tables reject a v6 method under an older version before the IR runs, so
  `compile` behaves as before. Only `compileTyped` with an AST typed under a different
  `VersionContext` than the one active at build time could reach those guards. There, a call that
  lowers to a dedicated node without consulting the method table (a `getReg` with a constant
  register id becomes `ExtractRegisterAs`) now builds a tree the older version can evaluate, while
  a call that needs the version-gated descriptor lookup still fails, from TreeBuilding instead of
  the IR arm. A dynamic register id, and ordinary source compilation under the older version, are
  rejected as before.

The `typeSubst` fix in `MethodCall.mirror` (commit 82cb87336) is preventive: `MethodCallMirrorSpec`
passes unchanged on the base commit, so no script that failed before compiles now.

## Verification

- Snapshot gate on every commit (see above); identical to the base at the final head.
- `sbt +compile`, `sbt +Test/compile` (2.12.21 and 2.13.18); full `scJVM/test`, `scJS/test`,
  `scJS/fastOptJS` + the `sigma-js` jest suite.
- New tests: `GenericMethodCallSpec` (methods with no lowering row round-trip on JVM and JS with
  no per-method code), `IRRewriteRulesSpec` (every live rewrite rule still fires),
  `MethodCallMirrorSpec` (type substitution survives lambda inlining).
- Deleted tests are exactly the IR-internal suites that tested past the removed interface:
  `CollsStagingTests`, `SigmaDslStaginTests`, `WOptionTests`, `WrappersTests`, `BaseLiftableTests`,
  `ReflectionGenerator` + `InterpreterReflectionGeneratorTests`, `ZLastSpec`.
- Adapted tests: `LoopTests`, `TypeDescsTests`, `ErgoScriptTestkit`, `TestContexts`, `TestLibrary`
  and the JVM `GraphVizExport` (they used the staged types or the reflection registry directly).
  `SigmaDslTesting`, `CompilerTestsBase` and the two `Platform` helpers gained the ErgoTree snapshot
  hook (`SIGMA_TREE_SNAPSHOT_DIR`, off by default); `ErgoTreeBuildingTest` uses fixed prover
  secrets (it asserts a fixed tree). The four `BasicOpsSpecification` cases whose random inputs
  end up in the script stay random: `Platform.testSeed` fixes their seed only while a snapshot is
  recorded, honours `SIGMA_TEST_SEED` for replay, and otherwise prints the fresh seed. No expected
  value changed.
- Reviewed before opening: a five-axis review (two independent read-only passes plus the author's)
  found no blocking issue; its findings are the last six commits (a third callee for the global
  builtins, flattened Elem descriptors, orphaned members deleted, comments and names cleaned,
  the mirror spec made to exercise `mirror`, four more rewrite-rule cases). A second pass made the
  AST fallback generic for every method container (only three shapes are special-cased) and made
  two rule tests consume an element so they cannot pass through the `length(map)` fold.
- Review findings on the PR (see the threads) are fixed in the last five commits: `DslConst`
  mirrors to itself again; a call node keeps only its method's explicit type arguments, so the
  inferred and written-out forms of an inferable type argument build one node and round-trip; call
  identity compares result types structurally rather than by display name; the v5 numeric methods
  fail in graph building with the old message; and `CallIdentitySpec`, `LoweringRowsSpec` and two
  more rule cases pin all of it.

## Follow-ups (not in this PR)

- What is left of the liftable machinery (`Liftable.lift`, `FuncConst`, `ThunkConst`, the
  pair/function/thunk instances, `Elem.liftable`) is reachable only through `Elem.sourceType`,
  which names the primitive Elems; removing it changes `Elem`'s surface and is left for its own PR.
- `core/.../reflection/Platform.scala` still documents `safeSimpleName` as debugging-only; it feeds
  `Elem.name` (as it did before this PR). One-line comment fix in a consensus module: separate PR.
- Candidate 2 of the architecture review: fold the remaining per-node lowering arms into the same
  descriptor-keyed table.

## Size

`sc` main sources: 11557 -> 6075 lines (5482 fewer). 37 commits, one per
entity or cleanup step, each independently gated.


sc/shared/src/main/scala/sigma/compiler/ir/Base.scala
```diff
@@ -420,6 +383,20 @@ abstract class Base { thisIR: IRContext =>
     })
   }
 
+  /** Node embedding a literal value of a DSL type (Box, Coll, BigInt, ...) into the graph.
+    * Unlike [[Const]] it is never given to the constant store, and it is extracted to a ValDef
+    * when shared (see `TreeBuilding`).
+    * @param constValue the literal value
+    * @param eT         type descriptor of the value's IR type
+    */
+  case class DslConst[T](constValue: T)(implicit val eT: Elem[T]) extends BaseDef[T] {
```
a-shannon, line 392: **[P2] Restore constant-node mirroring on `DslConst`**

[`DslConst`](https://github.com/ergoplatform/sigmastate-interpreter/blob/6b838c8a64eebda36f97308b045051dd8a10a215/sc/shared/src/main/scala/sigma/compiler/ir/Base.scala#L392-L398) no longer inherits `LiftedConst.mirror`, but it does not provide its own override. When map fusion inlines a constant mapper, mirroring falls through to `Def.transform` and throws.

Compile under V6 with `Map("xs" -> sigma.Colls.fromItems(1, 2, 3))` as the environment:

```scala
{
  getVar[Coll[Int]](1).get
    .map({ (x: Int) => xs })
    .map({ (ys: Coll[Int]) => ys.size })(0) > 0
}
```

The head fails with:

```text
Cannot transform definition using transform(DslConst(Coll(1,2,3)))
```

This reaches the live map-fusion rule: the first mapper has no dependency on its argument, its schedule falls back to its constant root, and `mirrorApply` attempts to mirror that root. The old `CollConst` inherited a mirror returning `self`.

The focused correction is to restore the immutable-constant behavior:

```scala
override def mirror(t: Transformer): Ref[T] = self
```

Please cover a constant mapper participating in map fusion. A direct source-level `f(1)` is insufficient: that path can keep an uninlined `Apply` and miss the faulty mirror path.

sc/shared/src/main/scala/sigma/compiler/ir/MethodCalls.scala
```diff
@@ -1,173 +1,100 @@
...
         case other: MethodCall =>
           receiver == other.receiver &&
-          method == other.method &&
+          callee == other.callee &&
           resultType.name == other.resultType.name &&
-          neverInvoke == other.neverInvoke &&
-          isAdapterCall == other.isAdapterCall &&
-          args.length == other.args.length &&
-          args.sameElementsNested(other.args) // this is required in case method have T* arguments
+          typeSubst == other.typeSubst &&
```
a-shannon, line 61: **[P2] Canonicalize redundant type substitutions before CSE**

[`buildMethodCall`](https://github.com/ergoplatform/sigmastate-interpreter/blob/6b838c8a64eebda36f97308b045051dd8a10a215/sc/shared/src/main/scala/sigma/compiler/ir/GraphBuilding.scala#L391-L395) retains the entire source `typeSubst`, and [`MethodCall.equals`/`hashCode`](https://github.com/ergoplatform/sigmastate-interpreter/blob/6b838c8a64eebda36f97308b045051dd8a10a215/sc/shared/src/main/scala/sigma/compiler/ir/MethodCalls.scala#L54-L73) includes it in node identity. Inferred and explicitly supplied *inferable* type parameters therefore distinguish equivalent calls.

For example, under V6:

```scala
{
  val x = getVar[Int](1).get
  sigmaProp(Global.serialize(x) == Global.serialize[Int](x))
}
```

The old wrapper represented both calls identically. The head keeps two calls, one with an empty substitution and one with `T -> SInt`, so the equality does not fold to `BoolToSigmaProp(TrueLeaf)`. It also retains the context-variable dependency that the folded output eliminates.

The base/head run confirmed different `ValueSerializer.serialize(buildTree)` output for this script:

```text
base: d10101
head: d801d601e4e30104d193dc6a03dd017201dc6a03dd017201
```

These are serialized expression bytes, without an ErgoTree header.

The same problem is reproducible with collections:

```scala
{
  val xs = getVar[Coll[Int]](1).get
  sigmaProp(xs.updated(0, 1) == xs.updated[Int](0, 1))
}
```

There is a second consequence of the same normalization issue: compiled ASTs no longer survive serialization with equality preserved for these explicit forms:

```scala
sigmaProp(Global.serialize[Int](getVar[Int](1).get).size > 0)
sigmaProp(getVar[Int](1).filter[Int]({ (x: Int) => x > 0 }).isDefined)
```

`serialize` and `Option.filter` have no serialized `explicitTypeArgs`. The compiler now retains `Map(T -> SInt)` in the AST, but [`MethodCallSerializer`](https://github.com/ergoplatform/sigmastate-interpreter/blob/6b838c8a64eebda36f97308b045051dd8a10a215/data/shared/src/main/scala/sigma/serialization/MethodCallSerializer.scala#L23-L32) writes no substitution for those methods, so deserialization produces `Map.empty`. `checkSerializationRoundTrip` fails on both examples.

Normalizing only in `TreeBuilding` would fix the latter symptom but leave CSE broken. I would canonicalize the IR node's substitution to the descriptor's actual `explicitTypeArgs`, while preserving method specialization and the result type. Required substitutions for methods such as `none[T]`, `some[T]`, and `fromBigEndianBytes[T]` must remain intact. The correction itself has not been implemented or tested in this review.

a-shannon (commented):
Reviewed `6b838c8a64eebda36f97308b045051dd8a10a215` against `20eb76b20e429342e6e9f9e377bb562e00c66997`.

The descriptor-based direction makes sense: it removes duplicated method wiring, and separating receiver operations from global builtins preserves the distinction between `p && q` and `allZK`. I reproduced two regressions that I would address before merging; the inline comments include reproductions and proposed corrections. Five regression probes pass on the base and fail on this head.

## Version-boundary clarification

The description says that a cross-version `compileTyped` call still fails, just later during descriptor lookup. There is a concrete exception: typecheck `SELF.getReg[Int](4).isDefined` under `(activatedVersion = 3, ergoTreeVersion = 3)`, then call `compileTyped` under `(2, 2)`.

The new [constant-register lowering](https://github.com/ergoplatform/sigmastate-interpreter/blob/6b838c8a64eebda36f97308b045051dd8a10a215/sc/shared/src/main/scala/sigma/compiler/ir/Lowering.scala#L103-L111) succeeds as `OptionIsDefined(ExtractRegisterAs(Self, R4, Option[Int]))`; it never needs the version-gated descriptor lookup. A dynamic register ID remains rejected, and ordinary source compilation under the older version still rejects `getReg`.

The lowered constant-register expression has valid legacy semantics, so I would treat this as a documentation and test clarification, separately from the two regressions above.

## Validation

Local execution used Scala 2.13.18, Microsoft OpenJDK 17.0.18, and SBT 1.12.11. The nine properties in `GenericMethodCallSpec`, `IRRewriteRulesSpec`, and `MethodCallMirrorSpec` passed on the head. The same eight additional probes were then executed against both revisions:

| Probe | Base | Head |
| --- | --- | --- |
| Captured collection in map fusion compiles | Pass | `DslConst` mirroring exception |
| Inferred/explicit `serialize` calls fold to true | Pass | Fail |
| Explicit `serialize[Int]` AST round-trip | Pass | Fail |
| Explicit `Option.filter[Int]` AST round-trip | Pass | Fail |
| Inferred/explicit `Coll.updated` calls fold to true | Pass | Fail |
| V6 constant-register AST built under `(2, 2)` | Rejected | Accepted as a legacy node |
| Ordinary older-version source `getReg` | Rejected | Rejected |
| V6 dynamic-register AST built under `(2, 2)` | Rejected | Rejected |

The baseline comparison rebuilt the exact base `sc` production and test sources in a separate output directory, reusing the unchanged build definition and `core`, `data`, `interpreter`, `parsers`, and `sdk` dependencies. Thus the five regression probes pass on the base and fail on the head. The baseline suite's one failure is the deliberately observed constant-register acceptance difference in the table, not an additional regression.

The [three existing CI jobs](https://github.com/ergoplatform/sigmastate-interpreter/actions/runs/35519065694) are green. I did not independently rerun the full snapshot corpus, Scala 2.12, Scala.js, or node-level evaluation. The findings concern compilation, IR optimization, and AST/serialization compatibility.

sc/shared/src/main/scala/sigma/compiler/ir/GraphBuilding.scala
```diff
@@ -428,6 +388,42 @@ trait GraphBuilding extends Base with DefRewriting { IR: IRContext =>
...
+  protected def globalOp(op: ValueCompanion, argsV: Seq[Ref[Any]], resultElem: Elem[_]): Ref[Any] =
+    asRep[Any](mkMethodCall(asRep[Any](sigmaDslBuilder), GlobalOpCallee(op), argsV, Map(), resultElem))
+
+  /** Same as [[globalOp]] with the result type of the AST `node` it lowers. */
+  protected def globalOpFor(node: SValue, op: ValueCompanion, argsV: Seq[Ref[Any]]): Ref[Any] =
+    globalOp(op, argsV, stypeToElem(node.tpe))
+
+  /** `Coll(items)` as a call node of the ConcreteCollection builtin. */
+  protected def fromItems[A](items: Seq[Ref[A]], eA: Elem[A]): Ref[sigma.Coll[A]] =
+    asRep[sigma.Coll[A]](mkMethodCall(asRep[Any](sigmaDslBuilder), GlobalOpCallee(ConcreteCollection), items, Map(), collElement(eA)))
```
a-shannon, line 415: # [P2] Keep structural result types in the identity of empty collection calls

Additional review of `6b838c8a64eebda36f97308b045051dd8a10a215` against `20eb76b20e429342e6e9f9e377bb562e00c66997`.

The new [`fromItems`](https://github.com/ergoplatform/sigmastate-interpreter/blob/6b838c8a64eebda36f97308b045051dd8a10a215/sc/shared/src/main/scala/sigma/compiler/ir/GraphBuilding.scala#L413-L415) stores the collection element type only in the call's `resultType`. For an empty collection, the receiver, callee, arguments and substitution are otherwise identical. [`MethodCall.equals` and `hashCode`](https://github.com/ergoplatform/sigmastate-interpreter/blob/6b838c8a64eebda36f97308b045051dd8a10a215/sc/shared/src/main/scala/sigma/compiler/ir/MethodCalls.scala#L54-L73) compare the result type by its display name, but [`FuncElem.getName`](https://github.com/ergoplatform/sigmastate-interpreter/blob/6b838c8a64eebda36f97308b045051dd8a10a215/sc/shared/src/main/scala/sigma/compiler/ir/TypeDescs.scala#L130-L133) does not preserve function association in that name.

This valid V6 source demonstrates the collision:

```scala
(Coll[(Int => Long) => Byte](), Coll[Int => Long => Byte]())
```

The two element types are structurally different, but both render as `Int => Long => Byte`. The base preserves both tuple component types. On the head, the compiler emits one shared empty collection and both tuple fields have the first collection's type. Reversing their order makes both fields take the other type instead. Thus `buildTree.tpe` differs from the typechecked expression's type.

It can also prevent a Boolean expression from compiling:

```scala
{
  val a = Coll[(Int => Long) => Byte]()
  val b = Coll[Int => Long => Byte]()
  b.getOrElse(0, { (x: Int) => { (y: Long) => 1.toByte } })(1)(2L) == 1.toByte
}
```

This compiles on the base. On the head, the unused `a` primes the CSE entry for `b` with the wrong element type. Lowering reconstructs `ByIndex` from that wrong collection type; the subsequent applications acquire the wrong types, and tree construction throws `ConstraintFailed` with a `NoType` operand at the final equality.

The ambiguous display name existed before this PR, but the old `colBuilder.fromItems` included the structural element descriptor in its arguments. Removing that argument exposes the collision. This is distinct from the already reported redundant-substitution issue: here two different calls are incorrectly merged, and both have an empty substitution.

I tested this two-line correction in `MethodCalls.scala`:

```diff
-          resultType.name == other.resultType.name &&
+          resultType == other.resultType &&
@@
-      h = h * 31 + resultType.name.hashCode
+      h = h * 31 + resultType.hashCode
```

This keeps structural result types in call identity and uses a matching hash. Changing display formatting alone would leave semantic identity dependent on a presentation convention.

Validation used Scala 2.13.18, JDK 17.0.18 and SBT 1.12.11 on JVM. Four identical focused properties pass on the base; on the unmodified head, one passes and three fail. The passing control uses empty `Coll[Int]` and `Coll[Long]`; the failures cover both function-type orders and the Boolean consumer above. The base run used the exact base compiler sources with unchanged dependencies.

With the correction, all four properties pass, as does an additional control checking that equality between two identically typed empty collections still folds to `sigmaProp(true)`. Those five properties plus `GenericMethodCallSpec`, `IRRewriteRulesSpec` and `MethodCallMirrorSpec` give **14 passed, 0 failed**. This validates the focused correction on JVM; Scala.js, Scala 2.12 and node-level behavior were not tested locally.

The wider `scJVM / Test / test` run included all upstream suites and the five new properties: **986 passed, 6 failed, 7 ignored**. The six failures concern CRLF-sensitive printing, source positions, generated documentation and template descriptions. Replaying the five affected upstream suites with the original head implementation reproduced all six failures; the three type-collision properties also failed again. The two local diagnostic suites for the previously reported findings were excluded from the wider run. Thus the full Windows run is not green, but none of its six failures was introduced by this correction.

cafebedouin (commented):
Review carried out by Claude (Anthropic, Claude Fable 5.1) for cafebedouin, with independent passes by Grok (grok-4.6-build) and Gemini (3.8 Flash, High). Head `6b838c8a6` against base `20eb76b20`; JDK 8u502, Scala 2.13.18, JVM only.

**Summary.** The byte-identity claim reproduces, and extends from the five gated suites to the whole `scJVM` module. No blocking issue was found beyond the three @a-shannon has already posted; one minor item is filed inline. @a-shannon noted not re-running the snapshot corpus — that re-run, a measurement of how far the corpus reaches into the new lowering, and a mutation check of the three new specs are what this review adds. Scala 2.12 and Scala.js were not run here (the CI JS job is green on the head).

### 1. The byte-identity gate
- **Reproduces.** Five suites on base + the hook commit (`5633d00e3`) and on the head: 538/538 on both, 90,674 lines each, `diff -r` empty; a second head run is identical.
- **Can fail.** With the `SizeMethod` row emitting a plain `MethodCall` instead of `SizeOf` (same meaning, different bytes), 354 snapshot lines change and 4 tests fail.
- **Holds for the whole module.** `scJVM/test` is 987/987 on both sides and 51 suites write a snapshot on both. 22 are byte-identical; 26 more are equal once random keys, hashes and ids are set aside (each line deserialized and re-serialized with constant extraction; template bytes, constant types, and every constant value other than `GroupElement` / `SigmaProp` / `Coll[Byte]` required to match) — 93,426 of 93,577 paired lines. The other 151 lines (`ErgoLikeInterpreterSpecification`, `RevenueSharingExamplesSpecification`, `ErgoTreeSerializerSpecification`, and one `SigmaCompilerTest` script whose source embeds a random key) differ the same way between two runs of the *same* head: test randomness, not the PR. Seeding the provers' secrets in the example suites, as the hook commit does for `ErgoTreeBuildingTest`, would let the hook cover them directly.

Three points for the description:
- The 90,674 lines are **857 distinct scripts**, recorded across versions and `-wrapper` duplicates.
- `ErgoTreeSpecification` compiles nothing (it passes hand-built expressions to `funcJitFromExpr`; the "script" argument is a label), so it writes no snapshot and is evidence neither way.
- The AST claim is stronger than stated. The snapshot is bytes only, but `checkCompilerResult` asserts `deserialize(serialize(tree)) == tree` before recording, on base too (the check predates the PR). Identical bytes plus a passing round trip on both sides means equal compiled trees — specialised `SMethod` types and `typeSubst` included — for every recorded script, not only the 49 `expectedExpr` cases that carry a specialised method. The gate's limit is its corpus, not its mechanism: under v6, `{ sigmaProp(Global.serialize[Int](getVar[Int](1).get).size > 0) }` gives `d191b1dc6a03dd01e4e301040400` on both sides but round-trips to an equal tree only on base (@a-shannon's second finding); had any gated suite used the explicit `[T]` form, the existing assertion would have failed it.

### 2. How much of the lowering table the corpus reaches
With a review-only probe at the `MethodCall` arm of `TreeBuilding.buildValue` (diff below), the five suites exercise **47 of the 53 `Lowering` rows**, and 82 distinct methods take the generic `plainMethodCall` path. Never reached:
- `OpCallee(ArithOp.operations(MinCode))` and `(MaxCode)` — BigInt `min` / `max` (that arm is `op.tpe == SBigInt` only);
- `MethodCallee(SGlobalMethods.xorMethod)` (the `GlobalOpCallee(Xor)` builtin row does fire);
- the three deliberate error rows: `GlobalOpCallee(SigmaAnd)`, `GlobalOpCallee(SigmaOr)`, `IsProvenMethod`.

To cover those, and the branches inside rows that a row hit does not distinguish, 25 targeted scripts were compiled under v5 and v6 on both sides: **43 of 50 results are identical**, including `min`/`max`, `Global.xor`, constant and non-constant `getReg`, and all four `proveDlog` / `proveDHTuple` shapes. The seven that differ are the two v5 numeric cases (inline) and cross-checks of posted findings (§3). A row hit is not a branch hit: `CreateProveDlog` fires 60 times in the five suites, yet removing its constant fold changes no snapshot line and fails no test, because no gated script calls `proveDlog` on a constant.

The error rows could not be reached from source — `isValid` is rejected by the typer, and `allZK` / `anyZK` over a non-literal collection fail in graph building (`Variable allZK not found in environment`) on both sides — so "keeps the old failure" is untested there rather than wrong (§4 has one exception). `min`, `max` and `Global.xor` cases in a gated suite would close the rest; scripts below.

### 3. Corroboration of the posted findings
- **`DslConst.mirror`** — reproduced under v5 and v6 with @a-shannon's script and with a `GroupElement` environment constant (`OUTPUTS.map({ (b: Box) => g }).map({ (e: GroupElement) => e.getEncoded.size })(0) > 0`). An inline `Coll(1, 2, 3)` literal in the same place compiles byte-identical to base, as does the same constant captured by the *second* mapper: the trigger is a DSL-typed `Constant` captured by the first mapper of a fused `map`.
- **Type substitution in call identity** — `Global.serialize(x) == Global.serialize[Int](x)`: `d10101` on base, `d801d601e4e30104d193dc6a03dd017201dc6a03dd017201` on the head, as reported. For whoever writes the regression test: through `CompilerTestsBase.compile` this does not fail readably — the round-trip assertion's mismatch report itself throws (`InvocationTargetException` ← `debox.KeyNotFoundException: key 12 was not found`), so the comparison has to be made outside that helper.
- **`resultType` compared by name** — reproduced with `(Coll[(Int => Long) => Byte](), Coll[Int => Long => Byte]())`: base keeps both element types; on the head both fields take the first collection's type behind one shared `ValDef`. Within one `IRContext` the collision also carries across compilations: compiled alone, the reversed-order script takes `Coll[Int => Long => Byte]` for both fields; compiled after the first script in the same context it takes that script's type and emits its bytes. The posted `resultType ==` correction should cover this (not tested here). The same by-name comparison exists at `primitives/Functions.scala:260` — unchanged from base, not from this PR, no misbehaviour shown — and that correction does not touch it.

### 4. The three new specs can fail
One mutant at a time on the head, all three specs run, the change reverted (baseline 9/9): **19 of 20 mutants are killed.**
- `IRRewriteRulesSpec`: all five arms of `IRContext.rewriteDef` and twelve of the thirteen arms of `GraphBuilding.rewriteDef` kill it when disabled — eleven at the case named for the rule, and five sigma arms because `isValid` then reaches `TreeBuilding` and the `IsProvenMethod` error row fires with the old message (the one exercise of an error row found). `anyOf(single) => single` is killed only incidentally, by the `anyOf(bools ++ sigmas)` case: there is an `allOf(single)` case on a plain boolean but no `anyOf` twin.
- `MethodCallMirrorSpec`: `MethodCall.mirror` passing `Map.empty` fails both the direct `mirror` property and the map-fusion property. `plainMethodCall` ignoring `mc.typeSubst` fails `GenericMethodCallSpec` and four mirror properties.
- The survivor is `ThunkForce(Def(ThunkDef(root, sch))) if sch.isEmpty => root`: no case covers it, and with it disabled the five gated suites still pass 538/538 with an unchanged snapshot. Dead or unexercised — worth a case if live, deleting if not.

### 5. Checked, nothing further found
- The PR adds or reshapes two graph node classes, `DslConst` and `MethodCall`. The three posted findings are one missing `mirror` and two identity defects in `MethodCall.equals` / `hashCode` (too fine on `typeSubst`, too coarse on the result type's name). No other node class in the diff adds or changes those overrides.
- Rewrite rules, all three `rewriteDef` sites: `GraphBuilding` is one-to-one with base apart from `SDBM.Colls(_) => colBuilder`; `DefRewriting` loses the `tryInvoke` arm (the PR's purpose); `IRContext` loses three `replicate` rules and the `WOptionConst(Some(x)).getOrElse` fold, none reachable from ErgoScript as far as could be traced (no `replicate` `SMethod`; base `GraphBuilding` constructs neither node).
- The `(objType, methodId)` key: no collision between `getRegV5` / `getReg` (7 / 19) or `getVar` / `getVarFromInput` (11 / 12), and the table has 53 entries at runtime, so no two rows collapse.

<details>
<summary>Reproduction: the probe, and the 25 differential scripts</summary>

The byte-identity run needs nothing beyond the PR itself: `SIGMA_TREE_SNAPSHOT_DIR=<fresh dir>` with `scJVM/testOnly *LanguageSpecificationV5 *LanguageSpecificationV6 *ErgoTreeSpecification *BasicOpsSpecification *ErgoTreeBuildingTest` on `5633d00e3` and on the head, then `diff -r`.

Row coverage — a review-only probe in `TreeBuilding.buildValue`; `ReviewProbe.hit` appends `kind what` to the file named by `$SIGMA_ROW_COVERAGE`:

```diff
         loweringFor(callee) match {
-          case Some(row) => row(mc, obj, args)
+          case Some(row) => ReviewProbe.hit("row", callee.getClass.getSimpleName + ":" + callee.toString); row(mc, obj, args)
           case None => callee match {
-            case MethodCallee(m) => plainMethodCall(mc, m, obj, args)
+            case MethodCallee(m) => ReviewProbe.hit("plain", m.objType.getClass.getSimpleName + "." + m.name); plainMethodCall(mc, m, obj, args)
```

Differential — each script compiled with `CompilerTestingCommons.compile` and a `TestingIRContext` under `VersionContext.withVersions(v, v)` for v = 2 and 3; the result is the hex of the serialized expression, or the exception. Environment: `g` = the group generator, `xs` = `Colls.fromItems(1, 2, 3)`.

```scala
"bigint-min-max"      -> "{ val a = getVar[BigInt](1).get; val b = getVar[BigInt](2).get; sigmaProp(min(a, b) == max(a, b)) }"
"global-xor-method"   -> "{ sigmaProp(Global.xor(getVar[Coll[Byte]](1).get, getVar[Coll[Byte]](2).get).size > 0) }"
"xor-builtin"         -> "{ sigmaProp(xor(getVar[Coll[Byte]](1).get, getVar[Coll[Byte]](2).get).size > 0) }"
"allZK-nonliteral"    -> "{ allZK(getVar[Coll[SigmaProp]](1).get) }"
"anyZK-nonliteral"    -> "{ anyZK(getVar[Coll[SigmaProp]](1).get) }"
"isValid"             -> "{ sigmaProp(getVar[SigmaProp](1).get.isValid) }"
"getReg-nonconst"     -> "{ sigmaProp(SELF.getReg[Int](getVar[Int](1).get).isDefined) }"
"getReg-const"        -> "{ sigmaProp(SELF.getReg[Int](4).isDefined) }"
"proveDlog-nonconst"  -> "{ proveDlog(getVar[GroupElement](1).get) }"
"proveDlog-const"     -> "{ proveDlog(g) }"
"dhtuple-const"       -> "{ proveDHTuple(g, g, g, g) }"
"dhtuple-mixed"       -> "{ proveDHTuple(g, g, getVar[GroupElement](1).get, g) }"
"getVar-explicit-ctx" -> "{ sigmaProp(CONTEXT.getVar[Int](1.toByte).isDefined) }"
"getVarFromInput"     -> "{ sigmaProp(CONTEXT.getVarFromInput[Int](0.toShort, 1.toByte).isDefined) }"
"num-toByte"          -> "{ sigmaProp(getVar[Int](1).get.toByte == 1.toByte) }"
"num-toBigInt"        -> "{ sigmaProp(getVar[Long](1).get.toBigInt == getVar[BigInt](2).get) }"
"num-toBytes"         -> "{ sigmaProp(getVar[Long](1).get.toBytes.size > 0) }"
"num-toBits"          -> "{ sigmaProp(getVar[Long](1).get.toBits.size > 0) }"
"bigint-toUnsigned"   -> "{ sigmaProp(getVar[BigInt](1).get.toUnsigned == getVar[UnsignedBigInt](2).get) }"
"global-none"         -> "{ sigmaProp(Global.none[Int]().isDefined) }"
"fromBigEndianBytes"  -> "{ sigmaProp(Global.fromBigEndianBytes[Long](getVar[Coll[Byte]](1).get) > 0L) }"
"deserializeTo"       -> "{ sigmaProp(Global.deserializeTo[Int](getVar[Coll[Byte]](1).get) > 0) }"
"xcheck-mapfusion-dslconst" -> "{ getVar[Coll[Int]](1).get.map({ (x: Int) => xs }).map({ (ys: Coll[Int]) => ys.size })(0) > 0 }"
"xcheck-serialize-explicit" -> "{ val x = getVar[Int](1).get; sigmaProp(Global.serialize(x) == Global.serialize[Int](x)) }"
"xcheck-mapfusion-group"    -> "{ OUTPUTS.map({ (b: Box) => g }).map({ (e: GroupElement) => e.getEncoded.size })(0) > 0 }"
```

The probe and the specs can be pushed as a branch if that is easier than re-deriving them.
</details>

sc/shared/src/main/scala/sigma/compiler/ir/TreeBuilding.scala
```diff
@@ -165,11 +134,38 @@ trait TreeBuilding extends Base { IR: IRContext =>
...
+        case (SCollectionMethods.FlatMapMethod, Seq(f)) =>
+          Map(SCollection.tOV -> f.asFunc.tpe.tRange.asCollection.elemType)
+        case (SCollectionMethods.ZipMethod, Seq(coll)) =>
+          Map(SCollection.tOV -> coll.asCollection[SType].tpe.elemType)
+        case _ => EmptySubst
+      }
+      val specMethod = generic.withConcreteTypes(typeSubst + (SCollection.tIV -> col.tpe.elemType))
+      builder.mkMethodCall(col, specMethod, args.toIndexedSeq, Map())
+    } else {
+      val generic = m.objType.getMethodById(m.methodId)
```
cafebedouin: **[P3] Keep the v5 numeric-method failure a compiler error**

Under v5, `getVar[Long](1).get.toBytes` and `.toBits` fail on both base and head, but differently:

```text
base: GraphBuildingException: … Type Long doesn't have methods: Don't know how to buildNode(MethodCall(…SMethod(SNumericTypeMethods,toBytes,…)))
head: java.lang.Exception: SNumericTypeMethods.getMethods shouldn't ever be called
```

No script that compiled stops compiling, so the severity is low. It is a counterexample to "Failures and their messages are unchanged", and the head failure has no source position.

Head stack: `TreeBuilding.plainMethodCall` (`TreeBuilding.scala:156`) → `SNumericTypeMethods$.getMethodById` → `getMethods` (`methods.scala:481`).

The v5 copies of the numeric methods keep `objType = SNumericTypeMethods` — the object; only the v6 copies rebind `objType = this` (`methods.scala`, `trait SNumericTypeMethods`). `toBytes` and `toBits` are in the v5 set, so they typecheck under v5. The object does not extend the trait, so the `_: SNumericTypeMethods` arm of the `GraphBuilding` fallback does not match them on either side. Base then fell through to `throwError("Type … doesn't have methods")` (base `GraphBuilding.scala:1312`); head falls through to `buildMethodCall`, and this line resolves through `m.objType.getMethodById` on a container whose `getMethods()` throws by design.

This also contradicts the new comment at `MethodCalls.scala:17-18` ("the numeric methods never become call nodes"): under v5 they do.

Either rejecting in the `GraphBuilding` fallback when `method.objType` is the `SNumericTypeMethods` object, or resolving through the receiver's container here (as the base fallback did with `MethodsContainer.getMethod(obj.tpe, name)`), restores the old failure. Under v6 both sides compile and agree (`d191b1db0506e4e301050400`).

sc/shared/src/main/scala/sigma/compiler/ir/Base.scala
jozanek: Fixed in 550bf3a49, with a map-fusion test.

sc/shared/src/main/scala/sigma/compiler/ir/MethodCalls.scala
jozanek: Fixed in e571cfd9b as suggested (explicit type args only); tests in `CallIdentitySpec`.

sc/shared/src/main/scala/sigma/compiler/ir/GraphBuilding.scala
jozanek: Fixed in ab6f4d64a with your two-line change; tests in `CallIdentitySpec`.

sc/shared/src/main/scala/sigma/compiler/ir/TreeBuilding.scala
jozanek: Fixed in 36884de73, old message restored; test in `CallIdentitySpec`.

jozanek:
@a-shannon thanks, all three fixed and gated byte-identical. Description corrected on the version boundary.

jozanek:
@cafebedouin thanks. `anyOf(single)`, `min`/`max`/`Global.xor` cases and the dead `ThunkForce` rule: bb9183ae3. Prover seeding: follow-up.

cafebedouin:
**[Verified] At `bb9183ae3` the fixes hold and the gate is still byte-identical to base.**

Executed (JDK 8, Scala 2.13, JVM): the five gated suites give snapshots byte-identical to base `5633d00e3`; the review's 50-row differential matches base on every row, v5 `toBytes`/`toBits` included; with the new `SNumericTypeMethods` arm removed, only the new `CallIdentitySpec` property fails; full `scJVM/test` passes 993/993. Not run: Scala 2.12, Scala.js.

Recommended: nothing further; prover seeding as a follow-up is fine.

